PR #1, opened 2024-02-08
Init the repository

kongzii:


.github/workflows/python_ci.yaml
```diff
@@ -0,0 +1,52 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ./.github/actions/python_prepare
+      - name: Run pytest
+        run: poetry run pytest
+        env:
+          RUN_PAID_TESTS: "1"
+    if: github.event_name == 'workflow_dispatch'
+    # TODO: Remove once the environment variables are fixed in the new repository.
+    continue-on-error: true
```
evangriffiths, line 44: Looks like it needs removing to pass CI

evangriffiths (approved):

.github/workflows/python_ci.yaml
kongzii: Added smoke test to not forget to enable it later on. I didn't know that 0 tests == error.

PR #15, opened 2024-02-15
Refactoring deployment a bit

kongzii:
## Summary by CodeRabbit

- **Refactor**
	- Enhanced and modularized deployment functionality for both local and Google Cloud Platform (GCP) scenarios.
- **New Features**
	- Added support for deploying with updated parameters such as memory and cron schedule on GCP.
- **Bug Fixes**
	- Fixed parameter types to allow `None` for `labels`, `env_vars`, and `secrets` in GCP deployment functions.
- **Tests**
	- Updated tests to reflect changes in deployment method calls.


examples/cloud_deployment/gcp/deploy.py
```diff
@@ -1,23 +1,13 @@
...
 if __name__ == "__main__":
-    current_dir = os.path.dirname(os.path.realpath(__file__))
-    fname = deploy_to_gcp(
-        requirements_file=None,
-        extra_deps=[
-            "git+https://github.com/gnosis/prediction-market-agent-tooling.git@main"
-        ],
-        function_file=f"{current_dir}/agent.py",
+    agent = DeployableCoinFlipAgent()
+    agent.deploy_gcp(
```
kongzii, line 8: `local` and `gcp` deployment have such different arguments that it would be cumbersome to if-else it. Plus, half of the arguments would always be un-used. 

So I split it into different functions, but comment if there was a different idea about how to do it 🙏

evangriffiths: I had a thought that we could reconcile this by using `functions-framework` that starts a local server from which you can run your function. Not to do as part of this PR, but should mean we can do the same for both types of deployment, and means we could more easily test it

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -34,26 +35,72 @@ def answer_binary_market(self, market: AgentMarket) -> bool:
...
+        entrypoint_template = f"""
+from {path_to_agent_file.replace("/", ".").replace(".py", "")} import *
+import functions_framework
+from prediction_market_agent_tooling.markets.markets import MarketType
+
+@functions_framework.http
+def main(request) -> str:
+    {self.__class__.__name__}().run(market_type={market_type.__class__.__name__}.{market_type.name})
+    return "Success"
+"""
```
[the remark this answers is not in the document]
evangriffiths: Wooo nice. The problem I saw with not making `run` a static function is that the user can now define their agent like:
```
from pydantic import BaseModel
from prediction_market_agent_tooling.deploy.agent import DeployableAgent

class Foo(BaseModel):
    x: int

class DeployableCoinFlipAgent(DeployableAgent):
    foo: Foo

    ...
```

so now can have arbitrary dependencies that we need to know about to initialize the agent. e.g. in this example we need to import and create an instance of `Foo`, so deployment will fail.

examples/cloud_deployment/gcp/deploy.py
```diff
@@ -27,18 +17,6 @@
         secrets={
             "MANIFOLD_API_KEY": f"JUNG_PERSONAL_GMAIL_MANIFOLD_API_KEY:latest"
         },  # Must be in the format "env_var_in_container => secret_name:version", you can create secrets using `gcloud secrets create --labels owner=<your-name> <secret-name>` command.
-        memory=512,
+        memory=256,
+        cron_schedule="0 */2 * * *",
```
evangriffiths, line 21: much cleaner :). Removing `remove_deployed_gcp_function` though - living a high-risk lifestyle!

prediction_market_agent_tooling/deploy/gcp/utils.py
```diff
@@ -11,9 +11,9 @@ def gcloud_deploy_cmd(
     gcp_function_name: str,
     source: str,
     entry_point: str,
-    labels: dict[str, str],
-    env_vars: dict[str, str],
-    secrets: dict[str, str],
+    labels: dict[str, str] | None,
```
evangriffiths, line 14: Could just make them default empty dicts to save a few chars below on if statements

examples/cloud_deployment/gcp/deploy.py
kongzii: That would solve compatibility for GCP, for the price of way more complicated local "deployment", but what if we add another option in the future?  maybe we will be able to make deployment to olas easier, or even AWS or something similar.

prediction_market_agent_tooling/deploy/agent.py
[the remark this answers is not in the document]
kongzii: But even if `run` is a static function, isn't the problem still the same? Even if `run` is static, they are still able to do 

```python
class DeployableCoinFlipAgent(DeployableAgent):
    foo: Foo
```

so, nothing changes, or does it?

Doing`{self.__class__.__name__}()` assumes that the class doesn't take any init arguments, but if `run` is static, the class needs to be initialised somewhere, so it's the same issue.

What we could do is 

```python
class DeployableAgent:
   @staticmethod
   def create_agent() -> "DeployableAgent":
      """Override this method if your agents needs custom initialization"""
      return DeployableAgent()
```

and use  `{self.__class__.__name__}.create_agent()`

examples/cloud_deployment/gcp/deploy.py
kongzii: Should it be auto-removed after it's scheduled to run every 2 hours? I thought that was there only for the example purposes.

prediction_market_agent_tooling/deploy/gcp/utils.py
kongzii: Isn't defaulting to dicts/lists/... forbidden pattern in Python, because of mutability issues?

evangriffiths: 😱 I wasn't aware of that before! TIL...
Explained well here: https://chat.openai.com/share/045afe93-640a-43e7-91b7-24c7e32a197e

Not an issue in this case I guess, as these dicts are only read, not written to. But thanks for the lesson and I retract my suggestion!

examples/cloud_deployment/gcp/deploy.py
evangriffiths: I guess the example was there originally as something that demos all the deployment functionality but destroys everything at the end, kind of like a test that the user could run. Just guarding against the user running it without realising that they're spawning a forever-alive agent! No big deal though

prediction_market_agent_tooling/deploy/agent.py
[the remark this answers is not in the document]
evangriffiths: I would say that because the static method can't access `self`, it doesn't depend on `Foo` to run it.

examples/cloud_deployment/gcp/deploy.py
evangriffiths: Yeah fair point.

prediction_market_agent_tooling/deploy/agent.py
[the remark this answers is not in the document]
evangriffiths: Just tried with this branch, and e.g.
```
from pydantic import BaseModel

from prediction_market_agent_tooling.deploy.agent import DeployableAgent
from prediction_market_agent_tooling.markets.data_models import AgentMarket
from prediction_market_agent_tooling.markets.markets import MarketType

class Foo(BaseModel):
    answer: bool

class FooAgent(DeployableAgent):
    def __init__(self, foo: Foo):
        self.foo = foo

    def answer_binary_market(self, market: AgentMarket) -> bool:
        return self.foo.answer

if __name__ == "__main__":
    agent = FooAgent(foo=Foo(answer=True))
    agent.deploy_gcp(
        repository="git+https://github.com/gnosis/prediction-market-agent-tooling.git@peter/refactor-deployment",
        market_type=MarketType.MANIFOLD,
        memory=1024,
    )
```

which produced
```
import functions_framework
from <some path> import *

from prediction_market_agent_tooling.markets.markets import MarketType


@functions_framework.http
def main(request) -> str:
    FooAgent().run(market_type=MarketType.MANIFOLD)
    return "Success"
```
(the `<some path>` string was a bit funky, but ignoring for now). This failed with:
```
    FooAgent().run(market_type=MarketType.MANIFOLD)
TypeError: FooAgent.__init__() missing 1 required positional argument: 'foo'
```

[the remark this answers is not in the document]
kongzii: Yeah, I'd expect it to fail, as `foo` is the required argument in your `FooAgent`.

So you could do

```python
class FooAgent(DeployableAgent):
    def __init__(self, foo: Foo):
        self.foo = foo

    def answer_binary_market(self, market: AgentMarket) -> bool:
        return self.foo.answer

    @staticmethod
    def create_agent() -> "FooAgent":
        return FooAgent(foo=...)
```

And instead of calling `FooAgent()` in the generated function, we would call `FooAgent.create_agent()`.

I think our situation is very similar to the MLFlow, https://mlflow.org/docs/latest/_modules/mlflow/pyfunc/model.html#PythonModel, they also have some "parent class" that needs to be subclassed, and then this subclass can be deployed by MLFlow-tools without any worries. They don't allow to change __init__ as well, instead there are special methods that are auto-called in the deployment.

---

But back to the staticmethod, let's say we have this:


```python
class DeployableAgent:
   ...
   
   @staticmethod
   def run(market_type: MarketType, _place_bet: bool = True) -> None:
        available_markets = [
            x.to_agent_market() for x in get_binary_markets(market_type)
        ]
        markets = self.pick_markets(available_markets)
        for market in markets:
            result = self.answer_binary_market(market)
            if _place_bet:
                print(f"Placing bet on {market} with result {result}")
                place_bet(
                    market=market.original_market,
                    amount=get_tiny_bet(market_type),
                    outcome=result,
                    omen_auto_deposit=True,
                )
```

Now we don't have `self` anymore, so we need to change it, but how? We can't do

```
result = DeployableAgent.answer_binary_market(market)
```

because that way, anything subclassed won't be used, the parent abstract class will always be used.

And we also can't do

```
   @staticmethod
   def run(agent: DeployableAgent, market_type: MarketType, _place_bet: bool = True) -> None:
       ...
```

because we would be again at the square 0, we would need to initialise the agent inside of the generated `main` function, and that initialisation could fail, so we would need something as `create_agent` I proposed above anyway.

Or did you have something else in mind?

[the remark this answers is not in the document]
evangriffiths: Ah you're right! Happy to go with this option then. We can add a guard to the base class to throw an exception if the derived class tries to override the __init__ method:

```
class Base:
    def __init__(self):
        pass

    def __init_subclass__(cls, **kwargs):
        if cls.__init__ is not Base.__init__:
            raise TypeError("Cannot override __init__ method of Base class")


class Derived(Base):
    # Raises a TypeError
    def __init__(self):
        pass
```

evangriffiths (approved):

prediction_market_agent_tooling/deploy/agent.py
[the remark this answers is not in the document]
kongzii: Wau I didn't know about `__init_subclass__` :o Added!

PR #17, opened 2024-02-16
Support for closed markets benchmarks + a little cleaning up

kongzii:
## Summary by CodeRabbit

- **New Features**
	- Enhanced prediction capabilities in benchmarking agents with new methods.
	- Added `MarketResolution` enum for improved prediction evaluation based on market resolution.
	- Introduced new enums and fields in market and prediction models for advanced filtering and resolution handling.
- **Refactor**
	- Refined prediction and market evaluation logic for clarity and efficiency.
	- Simplified prediction object structure in test scenarios.
- **Tests**
	- Updated test cases to align with the new prediction and market model changes.

### DIFF
--- a/prediction_market_agent_tooling/benchmark/agents.py
+++ b/prediction_market_agent_tooling/benchmark/agents.py
@@ -68,10 +68,7 @@ class AbstractBenchmarkedAgent:
             time_restriction_up_to=time_restriction_up_to,
         )
         if not is_predictable:
-            return Prediction(
-                is_predictable=is_predictable,
-                time_restriction_up_to=time_restriction_up_to,
-            )
+            return Prediction(is_predictable=is_predictable)
         return self.predict_restricted(
             market_question=market_question,
             time_restriction_up_to=time_restriction_up_to,
--- a/prediction_market_agent_tooling/benchmark/agents.py
+++ b/prediction_market_agent_tooling/benchmark/agents.py
@@ -1,8 +1,8 @@
 import random
 import typing as t
+from datetime import datetime
 
 from prediction_market_agent_tooling.benchmark.utils import (
-    EvaluatedQuestion,
     OutcomePrediction,
     Prediction,
 )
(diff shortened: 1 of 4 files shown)


prediction_market_agent_tooling/benchmark/agents.py
```diff
@@ -13,51 +13,84 @@ def __init__(self, agent_name: str, max_workers: t.Optional[int] = None):
         self.agent_name = agent_name
         self.max_workers = max_workers  # Limit the number of workers that can run this worker in parallel threads
 
-    def evaluate(self, market_question: str) -> EvaluatedQuestion:
-        raise NotImplementedError
+    def is_predictable(self, market_question: str) -> bool:
```
kongzii, line 16: Reduced the `evaluate research predict` pattern to just the `is-predictable predict` pattern, where `is-predictable` defaults to True now. So agents are free to be implemented anyhow they want.

prediction_market_agent_tooling/benchmark/agents.py
```diff
@@ -13,51 +13,84 @@ def __init__(self, agent_name: str, max_workers: t.Optional[int] = None):
...
 
-    def predict(
-        self, market_question: str, researched: str, evaluated: EvaluatedQuestion
+    def check_and_predict(self, market_question: str) -> Prediction:
+        is_predictable = self.is_predictable(market_question=market_question)
+        if not is_predictable:
+            return Prediction(is_predictable=is_predictable)
+        return self.predict(market_question=market_question)
+
+    def is_predictable_restricted(
```
kongzii, line 34: Plus, there are `_restricted` versions of them, they are called instead of the original ones, if we are doing benchmark on resolved markets.

prediction_market_agent_tooling/benchmark/utils.py
```diff
@@ -34,6 +43,16 @@ def _validate_outcome_prices(cls, value: list[float] | None) -> list[float] | No
...
 
+    @validator("created_time")
+    def _validate_created_time(cls, value: datetime) -> datetime:
+        if value.tzinfo is None:
+            value = value.replace(tzinfo=pytz.UTC)
+        return value
+
+    @property
+    def is_resolved(self) -> bool:
+        return self.resolution is not None
```
kongzii, line 54: Removed the original `is_resolved` and changed it to this property to not have two sources of the same information.

prediction_market_agent_tooling/benchmark/utils.py
```diff
@@ -48,19 +67,29 @@ def no_outcome_price(self) -> float:
...
 
+    @property
+    def probable_resolution(self) -> MarketResolution:
+        return (
+            self.resolution
+            if self.resolution is not None
+            else MarketResolution.YES
+            if self.p_yes > 0.5
+            else MarketResolution.NO
+        )
```
kongzii, line 78: I added `probable_resolution`, which can be used always, no matter if it is an open or resolved market.

Motivation is that if the market is resolved, it doesn't mean that if the final p_yes is >0.5, the market was resolved to True.

prediction_market_agent_tooling/benchmark/utils.py
```diff
@@ -48,19 +67,29 @@ def no_outcome_price(self) -> float:
...
 class OutcomePrediction(BaseModel):
     p_yes: float
     confidence: float
     info_utility: t.Optional[float]
 
     @property
-    def binary_answer(self) -> bool:
-        return self.p_yes > 0.5
+    def probable_resolution(self) -> MarketResolution:
+        return MarketResolution.YES if self.p_yes > 0.5 else MarketResolution.NO
```
kongzii, line 88: Added it to the outcome prediction as well, just for consistency + we don't need to always compare >0.5.

prediction_market_agent_tooling/benchmark/utils.py
```diff
@@ -233,7 +298,3 @@ def get_llm_api_call_cost(
     model_cost += model_costs[model]["completion_tokens"] * completion_tokens
     model_cost /= 1000
     return model_cost
-
-
-def should_not_happen(message: str, E: t.Type[Exception] = RuntimeError) -> t.NoReturn:
```
kongzii, line 238: This was duplicate, it's defined in the other utils as well.

evangriffiths (approved):

PR #19, opened 2024-02-19
Fix manifold bet profit calculation

evangriffiths:
Fixes a bug in manifold monitoring code where per-bet profit was calculated based on a user's market 'position'. However, this was incorrect, as a user can have multiple bets on a given market, but only one position. This meant:

1. Profit/loss was being double-counted in the case where a user has >1 bet on a given market
2. If an agent has `start_time`, `s` and has placed at least one bet on a given market after `s`, its profit for that market took into account bets also made before `s`

Result, before:
Screenshot 2024-02-19 at 12 44 37

After:
Screenshot 2024-02-19 at 12 47 02

## Summary by CodeRabbit

- **New Features**
	- Enhanced functionality to calculate total fees and profits for bets in the prediction market tool.

- **Refactor**
	- Improved profit calculation logic for more accurate outcomes.
	- Streamlined the process of assigning market outcomes to bets.

### DIFF
--- a/prediction_market_agent_tooling/markets/data_models.py
+++ b/prediction_market_agent_tooling/markets/data_models.py
@@ -6,6 +6,7 @@ from enum import Enum
 from pydantic import BaseModel
 from web3 import Web3
 
+from prediction_market_agent_tooling.benchmark.utils import should_not_happen
 from prediction_market_agent_tooling.gtypes import (
     USD,
     ChecksumAddress,
@@ -21,6 +22,12 @@ from prediction_market_agent_tooling.gtypes import (
 class Currency(str, Enum):
     xDai = "xDai"
     Mana = "Mana"
+
+
+class Resolution(str, Enum):
+    YES = "YES"
+    NO = "NO"
+    CANCEL = "CANCEL"
 
 
 class BetAmount(BaseModel):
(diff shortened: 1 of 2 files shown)


prediction_market_agent_tooling/markets/data_models.py
```diff
@@ -257,6 +260,19 @@ class ManifoldBet(BaseModel):
     createdTime: datetime
     outcome: str
 
+    def get_profit(self, market_outcome: bool) -> ProfitAmount:
+        outcome_bool = self.outcome == "YES"
```
kongzii, line 264: Could this be `True if self.outcome == "YES" else False if self.outcome == "NO" else should_not_happen()`? Just to catch potential errors if anything changes in their APIs.

prediction_market_agent_tooling/markets/manifold.py
```diff
@@ -124,19 +123,15 @@ def manifold_to_generic_resolved_bet(bet: ManifoldBet) -> ResolvedBet:
     if not market.resolutionTime:
         raise ValueError(f"Market {market.id} has no resolution time.")
 
-    # Get the profit for this bet from the corresponding position
-    positions = get_market_positions(market.id, bet.userId)
-    bet_position = next(p for p in positions if p.contractId == bet.contractId)
-    profit = bet_position.profit
-
+    market_outcome = market.resolution == "YES"
```
kongzii, line 126: Similarly here?

kongzii (approved):

PR #20, opened 2024-02-19
entrypoint_function_name variable for GCP

kongzii:


evangriffiths (approved):

evangriffiths:
`lastBetTime` pytest CI failure is unrelated snd should be fixed on main now

PR #23, opened 2024-02-20
Refactor prediction_market_agent_tooling/markets

evangriffiths:
Just a refactor, no functional changes. Directory structure before:
```
% tree             
.
├── prediction_market_agent_tooling
│   ├── markets
│   │   ├── data_models.py
│   │   ├── manifold.py
│   │   ├── markets.py
│   │   └── omen.py
```

after:
```
% tree
.
├── prediction_market_agent_tooling
│   ├── markets
│   │   ├── agent_market.py
│   │   ├── data_models.py
│   │   ├── manifold
│   │   │   ├── api.py
│   │   │   ├── data_models.py
│   │   │   └── manifold.py
│   │   ├── markets.py
│   │   └── omen
│   │       ├── data_models.py
│   │       └── omen.py
```

Each market type now has a class derived from `AgentMarket` that is used to take actions. Now if you want to add support for a new market, you only have to create a new directory under `markets` and implement this new derived class (and also add the market to the `MARKET_TYPE_MAP`).

Unfortunately there isn't a clean separation in Omen between the API functions and the `OmenAgentMarket` class, so they both have to live inside `omen.py` to avoid circular import issues. This can fix fixed later on if desired.

## Summary by CodeRabbit

- **New Features**
    - Introduced the `AgentMarket` class for a unified market template across prediction markets.
    - Added new data models for Manifold and Omen markets to facilitate interaction with market data.
- **Refactor**
    - Refactored betting logic in binary markets by integrating it directly into the `AgentMarket` class.
    - Simplified market type handling and retrieval logic with `AgentMarket` subclasses for Manifold and Omen markets.
- **Bug Fixes**
    - Updated import statements across various files to reflect new module paths and class locations.
- **Documentation**
    - Removed unused imports, classes, methods, and docstrings to clean up the codebase.


prediction_market_agent_tooling/markets/manifold/manifold.py
```diff
@@ -0,0 +1,45 @@
...
+        if amount.currency != self.currency:
+            raise ValueError(f"Manifold bets are made in Mana. Got {amount.currency}.")
+        place_bet(
+            amount=Mana(amount.amount),
+            market_id=self.id,
+            outcome=outcome,
+        )
+
+    @staticmethod
+    def from_data_model(model: ManifoldMarket) -> "ManifoldAgentMarket":
```
kongzii, line 33: Should this method be defined as `NonImplemented` in AgentMarket?

prediction_market_agent_tooling/markets/manifold/manifold.py
```diff
@@ -0,0 +1,45 @@
...
+)
+from prediction_market_agent_tooling.markets.manifold.data_models import ManifoldMarket
+
+
+class ManifoldAgentMarket(AgentMarket):
+    """
+    Manifold's market class that can be used by agents to make predictions.
+    """
+
+    currency: Currency = Currency.Mana
```
kongzii, line 18: This could be classvar to avoid mis-initialisation.

prediction_market_agent_tooling/markets/manifold/manifold.py
evangriffiths: No it doesn't have to be - these methods are only intended to be called inside the respective markets/<market>/ modules - not from the base class

prediction_market_agent_tooling/markets/manifold/manifold.py
kongzii:
```
t.ClassVar[Currency] = Currency. Mana
```

should work, but will need the same change in the parent class too

evangriffiths: Ah nice. So it doesn't fail on initialisation, but it does fail `mypy`

PR #24, opened 2024-02-21
Omen market maker functions

kongzii:
## Summary by CodeRabbit

- **New Features**
	- Introduced a script for creating markets on the Omen prediction platform with customizable details.
	- Added tools for interacting with Realitio contracts, creating markets, and managing conditions.
- **Enhancements**
	- Improved configuration flexibility by updating the default GNOSIS_RPC_URL setup.
	- Enhanced smart contract interaction with new utilities, retry mechanisms, and additional constants.


prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -103,15 +117,80 @@ def get_binary_markets(limit: int) -> list[AgentMarket]:
...
+    # This doesn't have a fixed contract address, as this is something created by the factory below.
+
+with open(
+    os.path.join(
+        os.path.dirname(os.path.realpath(__file__)),
+        "../abis/omen_fpmm_factory.abi.json",
+    )
+) as f:
+    # Contract ABI taken from https://gnosisscan.io/address/0x9083A2B699c0a4AD06F63580BDE2635d26a3eeF0#code.
+    OMEN_FPMM_FACTORY_ABI = ABI(f.read())
```
kongzii, line 129: I have an idea how to refactor this, so we won't have a ton of global constant and file openings here. But let's take a look at that in a following PR.

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -414,31 +504,30 @@ def omen_buy_outcome_tx(
...
-        market=market,
+        market_maker_contract_address=market.market_maker_contract_address_checksummed,
+        collateral_token_contract_address=market.collateral_token_contract_address_checksummed,
         amount_wei=amount_wei,
         from_address=from_address_checksummed,
         from_private_key=from_private_key,
         tx_params={"nonce": nonce},
     )
     nonce = Nonce(nonce + ONE_NONCE)  # Increase after each tx.
-    check_tx_receipt(approve_tx_receipt)
```
kongzii, line 426: This was always called anyway, so I moved into the utility function.

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -554,3 +639,260 @@ def binary_omen_sell_outcome_tx(
...
+    omen_approve_market_maker_to_spend_collateral_token_tx(
+        web3=web3,
+        market_maker_contract_address=OMEN_FPMM_FACTORY_CONTRACT_ADDRESS,
+        collateral_token_contract_address=DEFAULT_COLLATERAL_TOKEN_CONTRACT_ADDRESS,
+        amount_wei=initial_funds_wei,
+        from_address=from_address_checksummed,
+        from_private_key=from_private_key,
+        tx_params={"nonce": nonce},
+    )
+    nonce = Nonce(nonce + ONE_NONCE)  # Increase after each tx.
```
kongzii, line 818: I would also like to refactor this, so we don't have to do it manually like this after each tx. Again in the following PR.

prediction_market_agent_tooling/tools/gnosis_rpc.py
```diff
@@ -1,7 +1,9 @@
+import os
+
 import requests
 from web3.types import Wei
 
-GNOSIS_RPC_URL = "https://rpc.gnosischain.com/"
+GNOSIS_RPC_URL = os.getenv("GNOSIS_RPC_URL", "https://gnosis-rpc.publicnode.com")
```
kongzii, line 6: One of the bugs I was dealing with was in the end caused by unstable RPC. Seems like this one works better.

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -554,3 +639,260 @@ def binary_omen_sell_outcome_tx(
...
+
+    # Create the market.
+    create_market_receipt_tx = call_function_on_contract_tx(
+        web3=web3,
+        contract_address=OMEN_FPMM_FACTORY_CONTRACT_ADDRESS,
+        contract_abi=OMEN_FPMM_FACTORY_ABI,
+        from_address=from_address,
+        from_private_key=from_private_key,
+        function_name="create2FixedProductMarketMaker",
+        function_params=dict(
```
kongzii, line 875: `function_params` should be a list, but sometimes there are so many arguments (like here) that it's hard to follow them. We can use dictionary and then take `.values()`, because dicts are sorted in Python.

prediction_market_agent_tooling/markets/omen/omen.py
evangriffiths: I thought about that as well. Could make `call_function_on_contract_tx` take in the nonce as an arg, and return `TxReceipt` and the updated nonce.

evangriffiths (approved):

PR #43, opened 2024-02-23
Fix monitor_app

evangriffiths:
Fix code in market monitoring. `streamlit run examples/monitor/monitor.py` was throwing:
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for Market
resolution
  Input should be 'yes', 'no' or 'cancel' [type=enum, input_value='mkt', input_type=str]
```

### DIFF
--- a/prediction_market_agent_tooling/benchmark/utils.py
+++ b/prediction_market_agent_tooling/benchmark/utils.py
@@ -36,6 +36,7 @@ class CancelableMarketResolution(str, Enum):
     YES = "yes"
     NO = "no"
     CANCEL = "cancel"
+    MKT = "mkt"
 
 
 class Market(BaseModel):
@@ -65,6 +66,13 @@ class Market(BaseModel):
     @property
     def is_resolved(self) -> bool:
         return self.resolution is not None
+
+    @property
+    def has_successful_resolution(self) -> bool:
+        return self.resolution not in [
+            CancelableMarketResolution.CANCEL,
+            CancelableMarketResolution.MKT,
+        ]
 
     @property
     def is_cancelled(self) -> bool:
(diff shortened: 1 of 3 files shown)


prediction_market_agent_tooling/benchmark/utils.py
```diff
@@ -66,6 +67,13 @@ def _validate_created_time(cls, value: datetime) -> datetime:
     def is_resolved(self) -> bool:
         return self.resolution is not None
 
+    @property
+    def has_successful_resolution(self) -> bool:
```
kongzii, line 71: Should this be used instead of `is_resolved` in prediction_market_agent_tooling/benchmark/benchmark.py? Could you switch it there if you think so as well?

kongzii (approved):

prediction_market_agent_tooling/monitor/monitor.py
```diff
@@ -80,10 +83,10 @@ def monitor_market(open_markets: list[Market], resolved_markets: list[Market]) -
             [
                 (
                     1
-                    if m.resolution == "YES"
+                    if m.resolution == CancelableMarketResolution.YES
                     else (
                         0
-                        if m.resolution == "NO"
+                        if m.resolution == CancelableMarketResolution.NO
```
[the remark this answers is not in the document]
kongzii: Nice catch, thanks!

prediction_market_agent_tooling/benchmark/utils.py
evangriffiths: Good spot. It's only needed in the __init__ method of the Benchmarker where markets are validated. Safe to use `is_resolved` otherwise. Have updated.

PR #47, opened 2024-02-23
Use check in subprocess run

kongzii:
If check is True and the exit code was non-zero, it raises a CalledProcessError.


evangriffiths (approved):
Nice one!

PR #53, opened 2024-02-26
Option to change deployed agent's bet strategy

kongzii:
Solves https://github.com/gnosis/prediction-market-agent-tooling/issues/51

## Summary by CodeRabbit

- **New Features**
	- Introduced dynamic bet amount calculation based on market conditions.
	- Added functionality to calculate minimum bet amounts for winning in prediction markets.
	- Enhanced market and bet models with probability calculations and improved type hinting.
- **Bug Fixes**
	- Fixed the fixed bet amount issue by replacing it with a dynamic calculation method.
- **Tests**
	- Added tests for new betting calculation functionality across different market types and scenarios.


prediction_market_agent_tooling/markets/agent_market.py
```diff
@@ -30,6 +37,14 @@ def place_bet(self, outcome: bool, amount: BetAmount) -> None:
     def get_binary_markets(limit: int) -> list["AgentMarket"]:
         raise NotImplementedError("Subclasses must implement this method")
 
+    def get_outcome_str(self, outcome_index: int) -> str:
```
kongzii, line 40: Noticed this isn't used in the other class, seems like it makes more sense here along with `get_outcome_index`

prediction_market_agent_tooling/markets/bets.py
```diff
@@ -0,0 +1,24 @@
...
+    return bet_amount
+
+
+def minimum_bet_to_win_manifold(
+    answer: bool, amount_to_win: float, market: AgentMarket
+) -> int:
+    if market.currency != Currency.Mana:
+        raise ValueError(f"Manifold bets are made in Mana. Got {market.currency}.")
+    # Manifold lowest bet is 1 Mana, so we need to ceil the result.
+    return ceil(minimum_bet_to_win(answer, amount_to_win, market))
```
kongzii, line 24: For example, on https://manifold.markets/dglid/-will-the-ncaa-establish-any-new-po, with 52% p_yes, we need to bet 2 Mana in order to win 3 Mana. It doesn't make sense (except for benchmark purposes of course), to bet 1 or 2, because we would get only 1 or 2 back.

prediction_market_agent_tooling/markets/bets.py
```diff
@@ -0,0 +1,24 @@
+from math import ceil
```
evangriffiths, line 1: Maybe rename file to betting_strategies.py?

prediction_market_agent_tooling/markets/bets.py
```diff
@@ -0,0 +1,24 @@
...
+) -> float:
+    """
+    Estimates the minimum bet amount to win the given amount based on the current market price.
+    """
+    share_price = market.p_yes if answer else market.p_no
+    bet_amount = amount_to_win / (1 / share_price - 1)
+    return bet_amount
+
+
+def minimum_bet_to_win_manifold(
```
evangriffiths, line 18: Could we keep manifold-specific code in the manifold/ sub-directory. This function could actually be a method on the `ManifoldAgentMarket` class, like `get_tiny_bet_amount`

evangriffiths (approved):
Approved, but with a couple comments :)))))

prediction_market_agent_tooling/markets/bets.py
kongzii: Sure, renamed

prediction_market_agent_tooling/markets/bets.py
kongzii: Yes, much cleaner this way, implemented!

PR #59, opened 2024-02-27
Enable monitoring of Omen market

evangriffiths:
Refactoring code in monitor_app so it no longer hard-codes manifold market monitoring, and allows market monitoring for a general `MarketType`.

Involved mainly:
1. moving things from generic `Market` class in benchmarking code to the `AgentMarket` class and its subclasses.
2. Finding out which magic variables to use in the Omen subgraph query to reliably determine if the market is open/closed/resolved and what the p_yes is.

## Summary by CodeRabbit

- **New Features**
    - Enhanced Omen market functionalities with new data models and properties.
    - Improved market retrieval and sorting logic with new constants and enums.
- **Refactor**
    - Updated market retrieval and monitoring logic.
    - Consolidated market type handling and retrieval processes.
- **Tests**
    - Introduced new testing capabilities for Manifold and Omen prediction markets.
    - Comprehensive tests for market functionalities, including probabilities and resolved bets validation.
- **Chores**
    - Updated imports and refactored code to align with new functionalities and data models.

PR #61, opened 2024-02-28
Agent to replicate markets to Omen

kongzii:
## Summary by CodeRabbit

- **New Features**
    - Enhanced agent deployment with advanced configuration options and agent selection logic improvements.
    - Improved market analysis tools for better categorization and predictability analysis.
    - Added functionality for replicating markets on the Omen prediction market platform.
- **Enhancements**
    - Increased memory allocation for deployment from 256MB to 512MB.
    - Added new parameters like `timeout` and `cron_schedule` for improved deployment flexibility.
- **Security**
    - Implemented `SecretStr` type for API keys to enhance security.
- **Bug Fixes**
    - Fixed timezone validation in the `DeployedAgent` class.


prediction_market_agent_tooling/benchmark/utils.py
```diff
@@ -63,6 +66,12 @@ def _validate_created_time(cls, value: datetime) -> datetime:
             value = value.replace(tzinfo=pytz.UTC)
         return value
 
+    @validator("close_time")
+    def _validate_close_time(cls, value: datetime) -> datetime:
+        if value.tzinfo is None:
+            value = value.replace(tzinfo=pytz.UTC)
+        return value
```
kongzii, line 73: TODO: Use `add_timezone_validator` from the other PR when it's merged.

prediction_market_agent_tooling/markets/manifold/manifold.py
```diff
@@ -44,7 +44,7 @@ def from_data_model(model: ManifoldMarket) -> "ManifoldAgentMarket":
             id=model.id,
             question=model.question,
             outcomes=model.outcomes,
-            p_yes=Probability(model.pool.YES),
+            p_yes=Probability(model.probability),
```
kongzii, line 47: Fix :O Before I forgot that pool.YES actually isn't probability.

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -242,6 +218,39 @@ class Arbitrator(str, Enum):
...
+            outcomes
+            outcomeTokenAmounts
+            outcomeTokenMarginalPrices
+            fee
+        }
+    }"""
+
+    if not include_creator:
+        # If we aren't filtering by query, we need to remove it from where, otherwise "creator: null" will return 0 results.
+        query = query.replace("creator: $creator,", "")
```
kongzii, line 249: Do you know about some nicer way to do this?

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -500,11 +518,6 @@ def omen_buy_outcome_tx(
     # Get the index of the outcome we want to buy.
     outcome_index: int = market.get_outcome_index(outcome)
 
-    # Get the current nonce for the given from_address.
-    # If making multiple transactions quickly after each other,
-    # it's better to increae it manually (otherwise we could get stale value from the network and error out).
-    nonce: Nonce = web3.eth.get_transaction_count(from_address_checksummed)
```
kongzii, line 506: I removed the manual nonce creation, and instead of that, I added retries on the transaction function. So if it fails because of invalid nonce, it will wait a bit, network will update and then it will fetch a new nonce. Because for some reason, this manual nonce increasing didn't work correctly when I created N markets in a row.

prediction_market_agent_tooling/tools/gnosis_rpc.py
```diff
@@ -3,9 +3,7 @@
 import requests
 from web3.types import Wei
 
-GNOSIS_RPC_URL = os.getenv(
-    "GNOSIS_RPC_URL", "https://lb.nodies.app/v1/406d8dcc043f4cb3959ed7d6673d311a"
-)
+GNOSIS_RPC_URL = os.getenv("GNOSIS_RPC_URL", "https://gnosis-rpc.publicnode.com")
```
kongzii, line 6: On the hunt for the most stable RPC!

prediction_market_agent_tooling/tools/cache.py
```diff
@@ -0,0 +1,17 @@
...
+from joblib import Memory
+
+from prediction_market_agent_tooling.config import APIKeys
+
+MEMORY = Memory(APIKeys().CACHE_DIR, verbose=0)
+
+T = TypeVar("T", bound=Callable[..., Any])
+
+
+def persistent_inmemory_cache(func: T) -> T:
```
kongzii, line 13: TODO: Remove it from EVO and use it from here.

prediction_market_agent_tooling/tools/is_predictable.py
```diff
@@ -0,0 +1,48 @@
...
+"{question}"
+
+Then, write down what is the future event of the question, what it reffers to and when that event will happen if the question contains it.
+
+Then, give your final decision, write either "yes" or "no" about whether the question is answerable.
+"""
+
+
+@persistent_inmemory_cache
+def is_predictable(
```
kongzii, line 27: TODO: Remove it from EVO and use it from here.

tests/test_benchmark.py
```diff
@@ -62,7 +61,18 @@ def test_benchmark_run(
     dummy_agent: DummyAgent, dummy_agent_no_prediction: DummyAgentNoPrediction
 ) -> None:
     benchmarker = bm.Benchmarker(
-        markets=get_markets(number=1, source=MarketSource.MANIFOLD),
```
kongzii, line 65: Just fixing tests

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -112,13 +114,14 @@ def {entrypoint_function_name}(request) -> str:
...
 
-        monitor_agent = DEPLOYED_AGENT_TYPE_MAP[market_type].from_api_keys(
-            name=gcp_fname,
-            deployableagent_class_name=self.__class__.__name__,
-            start_time=start_time or datetime.utcnow(),
-            api_keys=gcp_resolve_api_keys_secrets(api_keys),
-        )
-        env_vars |= monitor_agent.model_dump_prefixed()
+        if dump_monitor_agent:
+            monitor_agent = DEPLOYED_AGENT_TYPE_MAP[market_type].from_api_keys(
```
kongzii, line 118: This is a hot-feature to allow deployment of agents that aren't meant for predicting markets and so won't be in the monitor. I created https://github.com/gnosis/prediction-market-agent-tooling/issues/63 to do it in a nicer way.

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -242,6 +218,39 @@ class Arbitrator(str, Enum):
 """
 
 
+def construct_query_get_fixed_product_markets_makers(include_creator: bool) -> str:
+    query = """query getFixedProductMarketMakers($first: Int!, $outcomes: [String!], $creator: Bytes = null) {
```
evangriffiths, line 222: Doesn't look like you need a default value for creator, as you always include it in the `"variables"` dict

prediction_market_agent_tooling/markets/omen/omen.py
evangriffiths: No, it's annoying, but can't think of a better way!

pyproject.toml
```diff
@@ -30,6 +30,8 @@ langchain-community = ">=0.0.19"
 scikit-learn = "^1.4.0"
 tabulate = "^0.9.0"
 types-pytz = "^2024.1.0.20240203"
+langchain = "^0.1.9"
```
evangriffiths, line 33: concerned about adding langchain as a core dependency of this repo, as this might make it harder to install PMAT in other agent projects (e.g. this wouldn't be compatible with the mech repo).

Since `DeployableReplicateToOmenAgent` isn't naturally a prediction market agent in the way that `DeployableAgent` was originally designed, how do you feel about moving that class from `prediction_market_agent_tooling/markets` to `examples`, and then this could become a dev dependency?

prediction_market_agent_tooling/markets/categorize.py
```diff
@@ -0,0 +1,35 @@
+from langchain.prompts import ChatPromptTemplate
+from langchain.schema.output_parser import StrOutputParser
+from langchain_openai import ChatOpenAI
+
+from prediction_market_agent_tooling.benchmark.utils import Market
```
evangriffiths, line 5: I'd like to be moving away from having this fragmentation of using different 'Market' classes (benchmark/utils.py in some places, markets/agent_market.py in other places). Trying to do this in https://github.com/gnosis/prediction-market-agent-tooling/pull/59 for example.

prediction_market_agent_tooling/tools/is_predictable.py
```diff
@@ -0,0 +1,48 @@
...
+- The answer is probably Google-able, after the event happened.
+- The question can not be about itself.
+
+Follow a chain of thought to evaluate if the question is answerable:
+
+First, write the parts of the following question:
+
+"{question}"
+
+Then, write down what is the future event of the question, what it reffers to and when that event will happen if the question contains it.
```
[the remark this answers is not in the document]
kongzii: It seems like my internal language model embedded this word badly years ago 😄 

Thanks!

prediction_market_agent_tooling/markets/categorize.py
kongzii: Changed `infer_category` to accept only the question as string, no need for market class. So when you finish your PR there shouldn't be any problems 🙏

pyproject.toml
kongzii: The issue is that we can't move it to `examples`, because then it won't be importable in the GCP function:

```python
from {path_to_agent_file.replace("/", ".").replace(".py", "")} import *
```

(`from examples.blabla import *` will fail)

But I think we can make it an optional dependency anyway, and if someone wants to use these LLM-powered features, he just needs to make it compatible, wdyt?

prediction_market_agent_tooling/markets/omen/omen.py
kongzii: True, I just leave it in case we use `construct_query_get_fixed_product_markets_makers(include_creator=False)` somewhere else

prediction_market_agent_tooling/tools/is_predictable.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    Evaluate if the question is actually answerable.
+    """
+    llm = ChatOpenAI(model=engine, temperature=0.0)
+
+    prompt = ChatPromptTemplate.from_template(template=prompt_template)
+    messages = prompt.format_messages(question=question)
+    completion = llm(messages, max_tokens=256).content
+
+    if "yes" in completion.lower():
```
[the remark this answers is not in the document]
kongzii: Good catch, I added a little prompt trick there.

prediction_market_agent_tooling/markets/categorize.py
evangriffiths: Yay cool. But it is still used in prediction_market_agent_tooling/markets/omen/replicate/replicate.pyYay cool. But it is still used in prediction_market_agent_tooling/markets/omen/replicate/replicate.py

pyproject.toml
evangriffiths: > because then it won't be importable in the GCP function:

Do you mean importable into examples/cloud_deployment/gcp/deploy.py ? I think that's okay - in this case IMO it's preferable to have code duplication by making a separate example than by adding complexity to the deployment of betting agents example

prediction_market_agent_tooling/markets/categorize.py
kongzii: There it's used because it's coming from `get_markets` function. I thought that's what you are refactoring and once that's done, it would just work? Or maybe you mean something else and I'm just missing it.

pyproject.toml
kongzii: I meant that I can't move `DeployableReplicateToOmenAgent` out of `prediction_market_agent_tooling/` because then, our generated GCP function code won't work:

Screenshot by Dropbox Capture

It would be something like `from examples. cloud_deployment.gcp.deploy import *`, but `examples` is not part of our package, so it will fail to import.

That's why both agent examples are inside of our package:

- prediction_market_agent_tooling/deploy/agent_example.py
- prediction_market_agent_tooling/markets/omen/replicate/agent_example.py

kongzii: I know it isn't nice but it was the only way I could make it work, but if you have a nicer one I'm open to it!

One thing I can think of is having another folder that would be distributed too (not sure if it's possible but sounds like it should be), called `prediction_market_agent_tooling_examples` and the agent examples could be there.

kongzii: > by adding complexity to the deployment of betting agents example

Or if you mean that `dump_monitor_agent=agent_name != AgentName.replicate,` inside of `examples/cloud_deployment/gcp/deploy.py` isn't very nice, then yes I agree on that too, I thought as part of https://github.com/gnosis/prediction-market-agent-tooling/issues/63, we could have something like:

1. DeployableAgent class
2. DeployableBettingAgent, subclass of (1)

It sounded like something useful for the future, as not all agents have to be prediction market ones. And then `examples/cloud_deployment/gcp/deploy.py` can be split into two.

prediction_market_agent_tooling/markets/categorize.py
evangriffiths: Well it still won't be replaceable with AgentMarket.get_binary_markets after https://github.com/gnosis/prediction-market-agent-tooling/pull/59 because there isn't a PolymarketAgentMarket subclass yet. For the interest of getting this merged sooner than later I'm happy for this to go in like it is, but it should be replaced down the line.

kongzii: Then I can replace it as soon as everything needed is in place 👏

pyproject.toml
evangriffiths: Oo, I hadn't thought about a world of DeployableAgents that aren't betting agents. Yes maybe we do want that - will have to think about it more.

But in the mean time we could still make `DeployableReplicateToOmenAgent` a separate example outside of the package - `deploy_to_gcp` is general enough to support it via `function_file` and `requirements_file` args - just can't do it with `DeployableReplicateToOmenAgent.deploy_gcp` currently

evangriffiths (approved):
No further requested changes, LGTM!!

PR #69, opened 2024-03-06
Do not wrap xDai if there are enough of them

kongzii:
## Summary by CodeRabbit

- **New Features**
	- Added functionality to retrieve ERC20 token balance.
- **Enhancements**
	- Improved market deposit process to conditionally deposit collateral based on token balance.
- **Bug Fixes**
	- Updated the deposit transaction function to correctly handle scenarios with insufficient balance.


evangriffiths (approved):
Good spot!

PR #79, opened 2024-03-08
Funding of market

kongzii:
## Summary by CodeRabbit

- **New Features**
	- Enhanced funding and fund removal capabilities for Omen markets in the prediction market tooling.
- **Tests**
	- Expanded test coverage for market creation, funding, and fund removal operations.

### DIFF
--- a/tests/markets/test_omen.py
+++ b/tests/markets/test_omen.py
@@ -1,4 +1,3 @@
-import time
 from datetime import datetime
 
 import numpy as np
@@ -23,6 +22,7 @@ from prediction_market_agent_tooling.markets.omen.omen import (
     omen_remove_fund_market_tx,
     pick_binary_market,
 )
+from prediction_market_agent_tooling.tools.contract import wait_until_nonce_changed
 from prediction_market_agent_tooling.tools.utils import check_not_none
 from prediction_market_agent_tooling.tools.web3_utils import xdai_to_wei
 from tests.utils import RUN_PAID_TESTS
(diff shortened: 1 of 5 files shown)


tests/markets/test_omen.py
```diff
@@ -92,6 +95,35 @@ def test_omen_create_market() -> None:
...
+    remove_fund = omen_outcome_type(xdai_to_wei(xdai_type(0.01)))
+    keys = APIKeys()
+    omen_fund_market_tx(
+        market=market,
+        funds=funds,
+        from_address=keys.bet_from_address,
+        from_private_key=keys.bet_from_private_key,
+        auto_deposit=True,
+    )
+    time.sleep(10)  # Wait for the transaction to be mined.
```
evangriffiths, line 117: Is there something we could poll to see if it's ready?
e.g. 
```
def wait_until_nonce_has_updated(current_nonce: Nonce) -> None:
    while nonce_has_not_updated(current_nonce):
        time.sleep(0.1)
```

kongzii: We are actually doing something like that in https://github.com/gnosis/prediction-market-agent-tooling/blob/2b2cfc36a20c32cb5a6325a9e6e6cdbe06e257f3/prediction_market_agent_tooling/tools/web3_utils.py#L130.

I thought that that should be it. But somehow it isn't enough (maybe free RPC's node is behind a bit?).

kongzii: I'll try the nonce trick

evangriffiths (approved):

tests/markets/test_omen.py
kongzii: Seems like it works!! 🎉

PR #81, opened 2024-03-08
Add Brier Score metric and table to monitoring

evangriffiths:
Before:
Screenshot 2024-03-09 at 00 59 17

After:
Screenshot 2024-03-09 at 00 53 22

Noticing a couple existing bugs(?) to investigate and fix in future PRs:
- Manifold brier score seems too low. [This](https://www.metaculus.com/notebooks/15359/predictive-performance-on-metaculus-vs-manifold-markets/) suggests it should be ~0.1 as apposed to ~0.03
- Something's not right with resolved Omen markets - most have [0, 0] outcome tokens.

## Summary by CodeRabbit

- **New Features**
    - Introduced a feature to compute boolean outcomes for market resolutions.
    - Added functionality to calculate and display Brier scores for market predictions.
    - Enhanced market monitoring with the addition of market outcome bias metrics.
- **Refactor**
    - Adjusted the monitoring application to improve information presentation and behavior.

### DIFF
--- a/prediction_market_agent_tooling/markets/agent_market.py
+++ b/prediction_market_agent_tooling/markets/agent_market.py
@@ -43,6 +43,14 @@ class AgentMarket(BaseModel):
     def p_no(self) -> float:
         return 1 - self.p_yes
 
+    @property
+    def boolean_outcome(self) -> bool:
+        if not self.has_successful_resolution():
+            raise ValueError(
+                "Market must have successful resolution to compute boolean outcome."
+            )
+        return self.resolution == Resolution.YES
+
     def get_bet_amount(self, amount: Decimal) -> BetAmount:
         return BetAmount(amount=amount, currency=self.currency)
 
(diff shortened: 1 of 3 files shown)


prediction_market_agent_tooling/markets/agent_market.py
```diff
@@ -43,6 +43,14 @@ class AgentMarket(BaseModel):
     def p_no(self) -> float:
         return 1 - self.p_yes
 
+    @property
+    def boolean_outcome(self) -> bool:
+        if not self.has_successful_resolution():
+            raise ValueError(
+                "Market must have successful resolution to compute boolean outcome."
+            )
+        return self.resolution == Resolution.YES
```
kongzii, line 52:
```suggestion
        return True if self.resolution == Resolution.YES else False if self.resolution == Resolution.NO else should_not_happen(f"Probably a bug in self.has_successful_resolution")
```

Just to prevent any unnoticed bugs 🙏

kongzii (approved):

PR #98, opened 2024-03-13
Add redeeming positions to Omen market

gabrielfior:
Adds `redeem_position` to Omen markets, allowing funds deposited by agents to be claimable.
Note that the function can be called N times without failing, however only the first successful claim will result in the `collateralToken` being returned to the agent.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `Condition` class to enhance market modeling.
	- Added functionality for redeeming positions in Omen markets, including full position redemption with detailed transaction handling.
	- Implemented a script to retrieve and display public keys of deployed agents for specific market types using the Omen protocol.
- **Tests**
	- Added a test to verify the functionality of redeeming a full position in an Omen market.


tests/markets/test_omen.py
```diff
@@ -186,3 +187,18 @@ def test_resolved_omen_bets(a_bet_from_address: str) -> None:
     # Verify that all bets convert to generic resolved bets.
     for bet in resolved_bets:
         bet.to_generic_resolved_bet()
+
+
+@pytest.mark.skipif(not RUN_PAID_TESTS, reason="This test costs money to run.")
+def test_omen_redeem_positions() -> None:
```
gabrielfior, line 193: Ran this test manually, transaction was successful -> https://gnosisscan.io/tx/0x78967554380a6eda84e6c7bb576d9c75d56d950edc4bcdf2726f6bf0b62be1bd#eventlog

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -82,6 +85,15 @@ def place_bet(
             auto_deposit=omen_auto_deposit,
         )
 
+    def redeem_positions(
+        self, bet_from_address: ChecksumAddress, from_private_key: PrivateKey
```
kongzii, line 89: Just a small nite, it seems that `bet_` is a little misleading here as we aren't going to bet in this function.

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -337,6 +357,30 @@ def omen_buy_outcome_tx(
...
+    condition_id: str,
+    parent_collection_id: str,
+    index_sets: t.List[int],
+    web3: Web3 | None = None,
+) -> TxReceipt:
+    conditional_token_contract = OmenConditionalTokenContract()
+    return conditional_token_contract.send(
+        from_address=from_address,
+        from_private_key=from_private_key,
+        function_name="redeemPositions",
```
kongzii, line 373: The idea with these contract classes was that they will map onchain contract methods one-to-one.

For example, `setApprovalForAll` method:

Screenshot by Dropbox Capture


So afterwards, we can do something like 

```python
conditional_token_contract = OmenConditionalTokenContract()
conditional_token_contract.reedemPositions(...)
```

Anywhere in the code, and it will be properly typed.

Could you move this `.send` stuff to its own method on the contract class, and then you probably don't need this extra `omen_redeem_positions_tx` as it will be one-liner in `omen_redeem_full_position_tx`?

tests/markets/test_omen.py
```diff
@@ -3,6 +3,7 @@
 import numpy as np
 import pytest
 from web3 import Web3
+from web3.types import TxReceipt
```
kongzii, line 6: Just a note: you can import `TxReceipt` (and other useful typing stuff)  in `from prediction_market_agent_tooling.gtypes import ...` on the line 9 down.

In that file, we implement some custom types (Mana currency, PrivateKey, ...) and import these existing types from web3 library just for sake of easier importing in our codebase:

Screenshot by Dropbox Capture

(up to you, not necessary, just letting you know as it saves a few hits on your keyboard 😄 )

tests/markets/test_omen.py
```diff
@@ -186,3 +188,18 @@ def test_resolved_omen_bets(a_bet_from_address: str) -> None:
...
+        "0xBA125828EC00267BBB70564D5558B891EABDAB9B".lower()
+    )  # Market on which agent previously betted on
+    market = OmenAgentMarket.from_data_model(get_market(market_id))
+    keys = APIKeys()
+    tx_receipt: TxReceipt = omen_redeem_full_position_tx(
+        market=market,
+        from_address=keys.bet_from_address,
+        from_private_key=keys.bet_from_private_key,
+    )
+    assert tx_receipt["transactionHash"] is not None
```
kongzii, line 205: [check_tx_receipt](https://github.com/gnosis/prediction-market-agent-tooling/blob/2b2cfc36a20c32cb5a6325a9e6e6cdbe06e257f3/prediction_market_agent_tooling/tools/web3_utils.py#L61) is automatically called in each call of `self.send(...)` in our contract classes [here](https://github.com/gnosis/prediction-market-agent-tooling/blob/2b2cfc36a20c32cb5a6325a9e6e6cdbe06e257f3/prediction_market_agent_tooling/tools/web3_utils.py#L132).

If this check is necessary, could you add it to `check_tx_receipt` so we always check for it? Or if what we are already doing is enough, this can be removed.

tests/markets/test_omen.py
gabrielfior: Makes sense, thanks for the pointers

prediction_market_agent_tooling/markets/omen/data_models.py
```diff
@@ -36,6 +36,15 @@ def get_boolean_outcome(outcome_str: str) -> bool:
     raise ValueError(f"Outcome `{outcome_str}` is not a valid boolean outcome.")
 
 
+class Condition(BaseModel):
+    id: HexAddress
```
kongzii, line 40: conditionId is currently typed as [HexBytes](https://github.com/gnosis/prediction-market-agent-tooling/blob/a62e7808bc0811f7ed814bcf32fb1e8aace3b3d4/prediction_market_agent_tooling/markets/omen/omen_contracts.py#L68), are you sure it should be HexAddress?

```suggestion
    id: HexBytes
```

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -689,6 +716,110 @@ def omen_fund_market_tx(
...
+def build_parent_collection_id() -> HexStr:
+    return HASH_ZERO  # Taken from Olas
+
+
+def omen_redeem_full_position_tx(
+    market: OmenAgentMarket,
+    from_address: ChecksumAddress,
+    from_private_key: PrivateKey,
+    web3: Web3 | None = None,
+    unwrap_collateral_after_redeeming_positions: bool = True,
```
kongzii, line 728:
```suggestion
    unwrap_collateral_after_redeeming_positions: bool,
```

I think we shouldn't default to True in arguments that cost extra money

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -689,6 +716,110 @@ def omen_fund_market_tx(
...
+    if not market.is_resolved():
+        print("Cannot redeem winnings if market is not yet resolved. Exiting.")
+        return
+
+    amount_wei = get_conditional_tokens_balance_for_market(market, from_address, web3)
+    if amount_wei == 0:
+        print("No balance to claim. Exiting.")
+        return
+
+    conditional_token_contract = OmenConditionalTokenContract()
```
kongzii, line 754:
```suggestion
```

Already defined on line 735

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -689,6 +716,110 @@ def omen_fund_market_tx(
...
+    We derive the withdrawable balance from the ConditionalTokens contract through CollectionId -> PositionId (which
+    also serves as tokenId) -> TokenBalances.
+    """
+    conditional_token_contract = OmenConditionalTokenContract()
+    parent_collection_id = build_parent_collection_id()
+    balance = 0
+
+    for index_set in market.condition.index_sets:
+        collection_id: bytes = conditional_token_contract.call(
+            "getCollectionId",
```
kongzii, line 789: Could you define `getCollectionId` as method on `OmenConditionalTokenContract`, same as with `redeemPositions`?

The idea of contract classes was to implement their interface so we don't have to call manually their methods in an untyped way.

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -689,6 +716,110 @@ def omen_fund_market_tx(
...
+            [parent_collection_id, market.condition.id, index_set],
+            web3=web3,
+        )
+        # Note that collection_id is returned as bytes, which is accepted by the contract calls downstream.
+        position_id: int = conditional_token_contract.call(
+            "getPositionId",
```
kongzii, line 795: Same here 🙏

prediction_market_agent_tooling/markets/omen/omen_contracts.py
```diff
@@ -72,6 +72,29 @@ def getConditionId(
         )
         return id_
 
+    def redeemPositions(
+        self,
+        from_address: ChecksumAddress,
+        from_private_key: PrivateKey,
+        collateral_token_address: str,
```
kongzii, line 79:
```suggestion
        collateral_token_address: ChecksumAddress,
```

prediction_market_agent_tooling/markets/omen/omen_contracts.py
```diff
@@ -72,6 +72,29 @@ def getConditionId(
...
+        condition_id: str,
```
kongzii, line 80:
```suggestion
        condition_id: HexBytes,
```

prediction_market_agent_tooling/markets/omen/omen_contracts.py
```diff
@@ -72,6 +72,29 @@ def getConditionId(
...
+        parent_collection_id: str,
```
kongzii, line 81:
```suggestion
        parent_collection_id: HexStr,
```

`build_parent_collection_id` returns `HexStr`, so I guess this should be HexStr too?

tests/markets/test_omen.py
```diff
@@ -186,3 +187,17 @@ def test_resolved_omen_bets(a_bet_from_address: str) -> None:
     # Verify that all bets convert to generic resolved bets.
     for bet in resolved_bets:
         bet.to_generic_resolved_bet()
+
+
+# @pytest.mark.skipif(not RUN_PAID_TESTS, reason="This test costs money to run.")
```
kongzii, line 192: I guess this should be uncommented until we have testing chain in place?

(you don't have to comment these while testing it locally, just run tests as `RUN_PAID_TESTS=1 pytest tests/`)

```suggestion
@pytest.mark.skipif(not RUN_PAID_TESTS, reason="This test costs money to run.")
```

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -689,6 +716,110 @@ def omen_fund_market_tx(
...
+            [market.collateral_token_contract_address_checksummed, collection_id],
+            web3=web3,
+        )
+        balance_for_position: int = conditional_token_contract.call(
+            "balanceOf", [from_address, position_id], web3=web3
```
kongzii, line 800: And here 🙏

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -689,6 +716,110 @@ def omen_fund_market_tx(
...
+
+
+def withdraw_collateral_token(
+    amount_wei: Wei,
+    from_address: ChecksumAddress,
+    from_private_key: PrivateKey,
+    web3: Web3 | None = None,
+) -> TxReceipt:
+    collateral_token = OmenCollateralTokenContract()
+    from_address_checksummed = Web3.to_checksum_address(from_address)
```
kongzii, line 814:
```suggestion
```

No need to convert it to checksummed, as it's already typed to be checksummed in the arguments (`from_address: ChecksumAddress,`)

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -689,6 +716,110 @@ def omen_fund_market_tx(
...
+    return collateral_token.withdraw(
```
kongzii, line 815: Seems like this function `withdraw_collateral_token` only calls `withdraw` method of collateral token, it such case, I guess it can be removed and just the `withdraw` method can be used in `omen_redeem_full_position_tx`?

prediction_market_agent_tooling/markets/omen/omen.py
kongzii: Nevermind, just saw you already did this one in the next PR <3

tests/markets/test_omen.py
kongzii: Nevermind, just saw you already did this in the next PR <3

prediction_market_agent_tooling/markets/omen/omen.py
kongzii: nevermind, just saw the next PR, even nicer to just remove it 👍

PR #102, opened 2024-03-15
Allow positions to be redeemed during agent operation

gabrielfior:
- Added method `withdraw_all_wxdai_as_xdai` to deploy/agent.py
- Removed wxDAI withdrawal from `redeemPositions` method (since they are independent)
- Tested local execution from agent

## Summary by CodeRabbit

- **New Features**
	- Introduced new methods to enhance the functionality of prediction markets, including position redemption and pre/post bet processing.
	- Added API functions for interacting with the Omen prediction market, facilitating access to market information and user data.
	- Implemented a `balanceOf` method for retrieving position balances in prediction markets.

- **Enhancements**
	- Improved the `place_bet` method in the Omen prediction market for better handling of bets.
	- Made the `limit` parameter optional in the Polymarket `get_binary_markets` method, increasing flexibility.
	- Enhanced market deployment tools with new filtering capabilities and structured bet processing.

- **Bug Fixes**
	- Fixed an issue in the Omen prediction market's position redemption process by incorporating a balance check.

- **Refactor**
	- Removed unnecessary imports and updated methods across various modules for better efficiency and readability.
	- Streamlined the process of fetching market details and user positions from the Omen prediction market.


prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -158,15 +168,60 @@ def get_markets(
...
+            )
+        return None
+
+    def pre_processing(self, market_type: MarketType) -> None:
+        """
+        Executes actions that occur before bets are placed.
+        """
+        self.redeem_positions_from_markets(market_type)
+        # This is carried out only for wxDAI (which is the only collateral token on all Omen markets)
+        self.withdraw_all_wxdai_as_xdai()
```
gabrielfior, line 219: Note that this is done once for each MarketType, hence this method is here and not part of the market implementation (we don't want to unwrap small amounts of wxDAI each time, rather it's more efficient to unwrap the entire balance once).
Note that we also explicity check for the balance to be withdrawn (for safety).

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -181,5 +236,13 @@ def run(self, market_type: MarketType, _place_bet: bool = True) -> None:
                     outcome=result,
                 )
 
+    def post_processing(self) -> None:
+        pass
+
+    def run(self, market_type: MarketType, _place_bet: bool = True) -> None:
```
gabrielfior, line 242: I did a bit of re-org in this method - happy to get some feedback here and potentially adjust this.

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -158,15 +168,60 @@ def get_markets(
...
             raise ValueError(f"Unknown market type: {market_type}")
         # Fetch the soonest closing markets to choose from
-        available_markets = cls.get_binary_markets(limit=limit, sort_by=sort_by)
+        available_markets = cls.get_binary_markets(
+            limit=limit, sort_by=sort_by, filter_by=filter_by
+        )
         return available_markets
 
-    def run(self, market_type: MarketType, _place_bet: bool = True) -> None:
+    def redeem_positions_from_markets(
```
kongzii, line 182: This is a general `DeployableAgent` class, but redeeming is specific to Omen only.

Right now, we don't have market-specific deployable agent classes, so we probably should.

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -158,15 +168,60 @@ def get_markets(
...
+        market_type: MarketType,
+    ) -> None:
+        # We can only redeem positions from resolved markets.
+        resolved_markets = self.get_markets(market_type, filter_by=FilterBy.RESOLVED)
+        for market in resolved_markets:
+            print(f"Redeeming position from market {market.id}")
+            market.redeem_positions()
+        return None
+
+    def withdraw_all_wxdai_as_xdai(
```
kongzii, line 193: Same here, only Omen uses (w)xDai. 

Btw I think we don't have to do this at all in our agents, they will convert xdai back to wxdai on every iteration anyway.

prediction_market_agent_tooling/deploy/agent.py
kongzii: I suggested it above but I think it's clearer here.

Currently, we would:

1. Redeem positions
2. Withdraw wxDai to xDai
3. Convert xDai to wxDai
4. Bet

I think we can just skip (2.).

prediction_market_agent_tooling/deploy/agent.py
kongzii: It isn't very clear what's `pre_processing` and `post_processing`, maybe `before_process_bets` and `after_process_bets`?

prediction_market_agent_tooling/markets/agent_market.py
```diff
@@ -62,6 +62,12 @@ def get_tiny_bet_amount(self) -> BetAmount:
     def place_bet(self, outcome: bool, amount: BetAmount) -> None:
         raise NotImplementedError("Subclasses must implement this method")
 
+    def redeem_positions(self) -> None:
```
kongzii, line 65: Similarly here, we have 3 markets right now (omen, polymarket, manifold) and only Omen needs this afaik. 

So it probably make more sense to keep it omen specific, wdyt @evangriffiths ?

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -158,15 +168,60 @@ def get_markets(
...
+        )
         return available_markets
 
-    def run(self, market_type: MarketType, _place_bet: bool = True) -> None:
+    def redeem_positions_from_markets(
+        self,
+        market_type: MarketType,
+    ) -> None:
+        # We can only redeem positions from resolved markets.
+        resolved_markets = self.get_markets(market_type, filter_by=FilterBy.RESOLVED)
```
kongzii, line 187: `get_markets` will currently [return only 20 of them ](https://github.com/gnosis/prediction-market-agent-tooling/blob/df89a9d87f32e5a95f604a2855b5dd0489cac5df/prediction_market_agent_tooling/deploy/agent.py#L159) and their aren't filtered for our agent, so we would miss a lot of them.

However, after a quick look at [Omen graph ](https://github.com/gnosis/prediction-market-agent-tooling/blob/f1391b935e44284f2535dac604f2282f52c4f255/prediction_market_agent_tooling/markets/omen/omen.py#L58), it's not clear to me what's the best way to list markets where we placed  bets.

So any ideas are welcome.

prediction_market_agent_tooling/deploy/agent.py
gabrielfior: Just to make sure I understand: your suggestion is then, instead of, after redeeming the positions, we convert the wxDAI to xDAI (in the `withdraw_all_wxdai_as_xdai` method) and then convert xDAI -> wxDAI ([here](https://github.com/gnosis/prediction-market-agent-tooling/blob/main/prediction_market_agent_tooling/markets/omen/omen.py#L350) , we simply leave the wxDAI balance in the collateral token contract after redeeming, correct?

kongzii: > we simply leave the wxDAI balance in the collateral token contract after redeeming

Yeah that's what I meant, if you guys don't see any issues arising from it

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -163,15 +173,39 @@ def get_markets(
...
```
evangriffiths, line 192: Here you're only interested in the markets that your agent has placed a bet on, right? In that case I think you should use `get_resolved_omen_bets` and then you can use the `fpmm` property of each bet.

You can convert it to a OmenAgentMarket with `OmenAgentMarket.from_data_model(bet.fpmm)` I think

kongzii: Oh cool! I totally forget about this one.

prediction_market_agent_tooling/deploy/agent.py
evangriffiths: Looks like this is unused now, and it can be removed

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -87,13 +89,27 @@ def place_bet(
...
+        keys = APIKeys()
         return omen_redeem_full_position_tx(
             market=self,
-            from_private_key=from_private_key,
+            from_private_key=keys.bet_from_private_key,
+        )
+
+    def before_process_bets(self) -> None:
+        # We can only redeem positions from resolved markets.
+        resolved_markets = self.get_binary_markets(
```
evangriffiths, line 101: Ah [this](https://github.com/gnosis/prediction-market-agent-tooling/pull/102/files#r1528720076) comment from above actually applies to this call.

evangriffiths: And this method will have to be a `staticmethod` I believe

gabrielfior: Yes, static is the right call, done thx

prediction_market_agent_tooling/deploy/agent.py
gabrielfior: Using `get_resolved_omen_bets` now, thanks for the heads-up @evangriffiths

prediction_market_agent_tooling/markets/agent_market.py
gabrielfior: Implemented logic in `OmenAgentMarket` as discussed during stand-up

prediction_market_agent_tooling/deploy/agent.py
gabrielfior: renamed

prediction_market_agent_tooling/deploy/agent.py
gabrielfior: removed

prediction_market_agent_tooling/deploy/agent.py
gabrielfior: After discussion with @kongzii we decided to fix the max number of markets retrieved as 1000.
This is just a quick fix and will be implemented properly on #106

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -90,12 +86,76 @@ def place_bet(
...
+            filter_by=FilterBy.RESOLVED,
+            sort_by=SortBy.CLOSING_SOONEST,
+        )
+        for market in resolved_markets:
+            print(f"Redeeming position from market {market.id}")
+            market.redeem_positions()
+        return None
+
+    def after_process_bets(self) -> None:
+        pass
```
kongzii, line 157: It seems to me this is here in order to save time, because by default it iterates resolved markets, right?

In that case, shouldn't it be the default behaviour to not iterate anything (in both preprocess and posprocess) and add that iteration + processing only for the Omen market?

Because right now this is skipped for Omen markets, but it isn't skipped for Manifold & Polymarket, so we would need to keep in mind to overwrite that processing logic in every subclass.

Wdyt @evangriffiths @gabrielfior ?

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -783,8 +661,8 @@ def get_conditional_tokens_balance_for_market(
             [market.collateral_token_contract_address_checksummed, collection_id],
             web3=web3,
         )
-        balance_for_position: int = conditional_token_contract.call(
-            "balanceOf", [from_address, position_id], web3=web3
+        balance_for_position: int = conditional_token_contract.balanceOf(
```
kongzii, line 664: Cool thanks! Can you do this for `getPositionId` and `getCollectionId` above as well?

prediction_market_agent_tooling/markets/omen/omen_graph_queries.py
```diff
@@ -0,0 +1,395 @@
...
+            json={
+                "query": query,
+                "variables": {
+                    "creator": better_address.lower(),
+                    "id_gt": id_gt,
+                    "first": OMEN_QUERY_BATCH_SIZE,
+                },
+            },
+            headers={"Content-Type": "application/json"},
+        ).json()
```
kongzii, line 184: `response_to_model` could be used here instead of doing it manually a few lines below 🙏

prediction_market_agent_tooling/markets/omen/omen_graph_queries.py
```diff
@@ -0,0 +1,395 @@
...
+        FixedProductMarketMakersResponse,
+    )
+    return [
+        m
+        for m in markets.data.fixedProductMarketMakers
+        if not excluded_questions or m.question not in excluded_questions
+    ]
+
+
+def validate_market_for_filter_condition(
```
kongzii, line 331: Seems unused

prediction_market_agent_tooling/markets/omen/omen_graph_queries.py
```diff
@@ -0,0 +1,395 @@
...
+                "variables": {
+                    "creator": better_address.lower(),
+                    "creationTimestamp_gte": to_int_timestamp(start_time),
+                    "creationTimestamp_lte": to_int_timestamp(end_time),
+                    "id_gt": id_gt,
+                    "first": OMEN_QUERY_BATCH_SIZE,
+                },
+            },
+            headers={"Content-Type": "application/json"},
+        ).json()
```
kongzii, line 370: Similarly here, `response_to_model` can be used instead of doing it manually

tests/markets/test_omen_graph_queries.py
```diff
@@ -0,0 +1,11 @@
...
+from prediction_market_agent_tooling.markets.omen.omen_graph_queries import (
+    get_user_positions,
+)
+
+
+def test_get_user_positions() -> None:
+    keys = APIKeys()
+    user_positions = get_user_positions(keys.bet_from_address)
+    # We assume that the agent has at least 1 historical position
+    assert user_positions is not None
```
kongzii, line 11: It's typed to return list, so this will always be true. But if it's fixed, won't this test fail for addresses without any bet?

```
def get_user_positions(
    better_address: ChecksumAddress,
) -> t.List[OmenUserPosition]:
```

prediction_market_agent_tooling/markets/omen/data_models.py
```diff
@@ -45,6 +45,22 @@ def index_sets(self) -> t.List[int]:
...
 
+class Question(BaseModel):
+    id: HexAddress
+    answerFinalizedTimestamp: t.Optional[int] = None
+    currentAnswer: t.Optional[str] = None
+
+
+class OmenPosition(BaseModel):
+    id: HexAddress
+    conditionIds: t.List[HexAddress]
```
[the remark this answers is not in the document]
gabrielfior: Actually it should work with both variants.
image

Nevertheless I agree to switch it to HexBytes to keep it compatible with `getConditionId`

prediction_market_agent_tooling/markets/omen/omen_graph_queries.py
gabrielfior: I would keep it like this for now, as this will anyways be simplified with subgrounds :pray: 
Thanks for the heads-up!

prediction_market_agent_tooling/markets/omen/data_models.py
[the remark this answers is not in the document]
kongzii: Python doesn't care about types at runtime (Pydantic does), so I trust the code works, but if want Mypy to be useful for us, we need to take care of these small details. E.g. HexAddress (address of some contract) vs HexBytes (just general bytes in hex format) represent something different.

prediction_market_agent_tooling/markets/omen/omen_graph_queries.py
gabrielfior: Same argument as above

tests/markets/test_omen_graph_queries.py
gabrielfior: Great catch

prediction_market_agent_tooling/deploy/agent.py
gabrielfior: PS I discussed about this briefly with @kongzii , the problem with using `get_resolved_bets` for fetching markets is that, in some cases, no `creator` or `from_address` is passed, hence all trades from Omen are retrieved - highly inneficient.
Hence I'm reverting back to the original implementation, which is limited to 1000 markets. This will be improved once #106 is implemented.

prediction_market_agent_tooling/markets/agent_market.py
gabrielfior: Actually we still need `redeem_positions` on the general `AgentMarket`, since the method `get_binary_markets` returns the base class.
One could do some tricks to trick mypy and to change the return statement, but in my mind it makes the most sense to add this functionality to the base class (`AgentMarket`) as a reminder for other market implementations (Manifold for ex).

prediction_market_agent_tooling/deploy/agent.py
evangriffiths: > the problem with using get_resolved_bets for fetching markets is that, in some cases, no creator or from_address is passed

@gabrielfior Sorry I don't understand. Could you explain this bit?

kongzii: To clarify: We discussed to just query 1000 markets, because (by our luck), they will contain all the markets where we placed bet, so we can just filter it afterwards. 

> no creator or from_address is passed, hence all trades from Omen are retrieved

However, @gabrielfior I still don't understand this. As we talked about yesterday, why would we fetch all the bets from the market, if no `bet from address` filter is specified? In that case it indeed doesn't make sense, and fetching of bets can be skipped.

I see two flows we should have in the final implementation:

1. Query for markets generally. Optionally filtered by creator. There is no need to first fetch bets in this flow.
2. Query for markets where we placed bets. In this case, we should fetch our bets, and then fetch corresponding markets.

prediction_market_agent_tooling/markets/agent_market.py
kongzii: I see that subclasses (`ManifoldAgentMarket`, `OmenAgentMarket` and `PolymarketAgentMarket`) are typed as `get_binary_markets(...) -> list["AgentMarket"]`, however, that isn't necessary. You can fix the types there to `ManifoldAgentMarket`, `OmenAgentMarket` and `PolymarketAgentMarket` respectively. If you look into their implementations, you will see they indeed return these subclassed markets, so it's ok to fix the typing. No tricking is required; it will actually be the correct way.

> as a reminder for other market implementations (Manifold for ex).

There is no reminder required because neither Manifold nor Polymarket require redeeming. Omen is the only market out of all (afaik), who requires this.

prediction_market_agent_tooling/deploy/agent.py
kongzii: But what I didn't expect is that tests now take 2 hours to run 😱 

Screenshot by Dropbox Capture

I guess these 1000 markets don't come with information wherever we bet there? So you need to fetch bets for all of them?

prediction_market_agent_tooling/markets/omen/omen_graph_queries.py
kongzii: yep sounds good

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -90,12 +86,76 @@ def place_bet(
...
+    def was_bet_outcome_correct(self) -> bool:
+        keys = APIKeys()
+        # We fetch all bets irrespective of market
+        resolved_omen_bets = get_resolved_omen_bets(
+            start_time=datetime(2024, 1, 1),
+            end_time=None,
+            better_address=keys.bet_from_address,
+        )
+        resolved_bets_for_market = [
+            bet for bet in resolved_omen_bets if bet.fpmm.id == self.id
```
kongzii, line 98: Just noticed this here, if bets are coming with `fpmm: OmenMarket`, which is exactly the omen market you get from `get_omen_markets(...) -> list[OmenMarket]`, why you didn't like first fetching bets for some `better address` and then just getting the markets out of it, without any extra calls?

prediction_market_agent_tooling/deploy/agent.py
kongzii: I just noticed this on the other line: https://github.com/gnosis/prediction-market-agent-tooling/pull/102/files#r1533511063

Screenshot by Dropbox Capture

prediction_market_agent_tooling/markets/omen/omen_graph_queries.py
```diff
@@ -0,0 +1,385 @@
...
+            outcomeTokenMarginalPrices
+            fee
+            category
+        }
+    }
+}
+"""
+
+
+_QUERY_GET_SINGLE_FIXED_PRODUCT_MARKET_MAKER = """
```
kongzii, line 98: This is now duplicated in https://github.com/gnosis/prediction-market-agent-tooling/blob/1a4a150e58a15eded75130e11fada8758d906aed/prediction_market_agent_tooling/markets/omen/omen.py#L201

prediction_market_agent_tooling/markets/omen/omen_graph_queries.py
kongzii: sure sounds good

prediction_market_agent_tooling/markets/omen/omen_graph_queries.py
gabrielfior: thanks, removed

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -750,13 +604,6 @@ def omen_redeem_full_position_tx(
         web3=web3,
     )
 
-    if unwrap_collateral_after_redeeming_positions:
-        withdraw_collateral_token(
```
kongzii, line 754: If this is removed, then `withdraw_collateral_token` becomes unsused and can be removed as well 🙏

prediction_market_agent_tooling/markets/agent_market.py
gabrielfior: @kongzii fixed during our pair-programming session

prediction_market_agent_tooling/deploy/agent.py
gabrielfior: Implemented per our discussion - now 1. fetches bets for creator and
2. uses the market objects contained in such bets (i.e. resolved markets where we placed bets), so that `redeemPositions` is called in this object.

prediction_market_agent_tooling/markets/omen/omen.py
gabrielfior: removed

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -163,15 +175,49 @@ def get_markets(
...
+
+        if market_type == MarketType.OMEN:
+            keys = APIKeys()
+            resolved_omen_bets = get_resolved_omen_bets(
+                start_time=datetime(2020, 1, 1),
+                end_time=None,
+                better_address=keys.bet_from_address,
+            )
+            bets_per_market_id: t.Dict[HexAddress, t.List[OmenBet]] = defaultdict(list)
+            market_id_to_market: t.Dict[HexAddress, OmenMarket] = {}
```
kongzii, line 202: No need to change this, just FYI: Similarly as commented in my PR, you can use just `list[...], dict[..., ...], set[...], tuple[...]`. No need for typing library anymore

prediction_market_agent_tooling/markets/omen/omen_graph_queries.py
```diff
@@ -0,0 +1,400 @@
...
+        query = query.replace("creator: $creator,", "")
+
+    return query
+
+
+def get_omen_markets(
+    outcomes: list[str],
+    sort_by: SortBy,
+    filter_by: FilterBy,
+    first: int = 1000,  # max limit for markets
```
kongzii, line 295: Let's not default to 1000; both `get_{manifold, polymarket}_markets` don't have a default value for limit, so let's keep it consistent, please. No need for this, or is there?

```suggestion
    first: int
```

kongzii: Also, if we change order of arguments, it will create breaking changes for the repositories who use PMAT as a dependency, although it's unnecessary. Can you change it back to 

```
def get_omen_markets(
    first: int,
    outcomes: list[str],
    sort_by: SortBy,
    filter_by: FilterBy,
    created_after: t.Optional[datetime] = None,
    creator: t.Optional[HexAddress] = None,
) -> list[OmenMarket]:
```

please

prediction_market_agent_tooling/markets/omen/omen_graph_queries.py
```diff
@@ -0,0 +1,400 @@
...
+    created_after: t.Optional[datetime] = None,
+    creator: t.Optional[HexAddress] = None,
+    excluded_questions: set[str] | None = None,
+) -> list[OmenMarket]:
+    """
+    We return the first 1000 markets, since we don't have pagination in place (yet).
```
kongzii, line 301: Will be outdated if the definition is changed back, let's query whatever the user want to query.

prediction_market_agent_tooling/markets/omen/omen_graph_queries.py
```diff
@@ -0,0 +1,400 @@
...
+    # Initialize id_gt for the first batch of bets to zero
+    id_gt: str = "0"
+    all_bets: list[OmenBet] = []
+    query = _QUERY_GET_FIXED_PRODUCT_MARKETS_MAKER_TRADES
+
+    if not market_id:
+        query = query.replace("fpmm: $market_id", "")
+
+    if not filter_by_answer_finalized_not_null:
+        query = query.replace("fpmm_: {answerFinalizedTimestamp_not: null},", "")
```
kongzii, line 355: I'd suggest having this logic in extra function as above, like `construct_query_get_fixed_product_markets_makers_trades` or something, but if this is getting refactored with subgrounds as well, then no need to change it.

prediction_market_agent_tooling/markets/omen/omen_graph_queries.py
```diff
@@ -0,0 +1,400 @@
...
+    end_time: t.Optional[datetime],
+    better_address: ChecksumAddress,
+) -> list[OmenBet]:
+    # We filter by answer_finalized is not None on a subgraph level (for faster fetch times),
+    # however further assertions are performed on a data model level
+    bets = get_omen_bets(
+        start_time=start_time,
+        end_time=end_time,
+        better_address=better_address,
+        filter_by_answer_finalized_not_null=True,
```
kongzii, line 398: Nice!!

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -163,15 +175,49 @@ def get_markets(
...
+                # We keep track of the unique markets
+                if bet.fpmm.id not in market_id_to_market:
+                    market_id_to_market[bet.fpmm.id] = bet.fpmm
+
+            # We redeem positions for each unique resolved market where the
+            # agent has placed bets.
+            for market_id, omen_bets in bets_per_market_id.items():
+                market_data_model = market_id_to_market[market_id]
+                market = OmenAgentMarket.from_data_model(market_data_model)
+                market.redeem_positions(omen_bets)
```
kongzii, line 215: Sorry for yet another idea, but as I see `from_data_model()`, it's not necessary. `redeem_positions` can be implemented directly on the `OmenMarket` class. Maybe it even makes more sense.

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -90,10 +85,51 @@ def place_bet(
             auto_deposit=omen_auto_deposit,
         )
 
-    def redeem_positions(self, from_private_key: PrivateKey) -> None:
+    def was_bet_outcome_correct(self, resolved_omen_bets: t.List[OmenBet]) -> bool:
+        resolved_bets_for_market = [
+            bet for bet in resolved_omen_bets if bet.fpmm.id == self.id
+        ]
```
kongzii, line 91:
```suggestion
        if not resolved_bets_for_market:
            raise ValueError(f"No bets for this market.")
        ]
```

prediction_market_agent_tooling/markets/omen/omen_contracts.py
```diff
@@ -108,6 +142,13 @@ def does_condition_exists(
     ) -> bool:
         return self.getOutcomeSlotCount(condition_id) > 0
 
+    def payoutDenominator(self, condition_id: HexAddress) -> int:
```
kongzii, line 145: `getConditionId` return `HexBytes`

```suggestion
    def payoutDenominator(self, condition_id: HexBytes) -> int:
```

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -163,15 +175,49 @@ def get_markets(
...
         return available_markets
 
-    def run(self, market_type: MarketType, _place_bet: bool = True) -> None:
+    def before(self, market_type: MarketType) -> None:
+        """
+        Executes actions that occur before bets are placed.
+        """
+
+        if market_type == MarketType.OMEN:
+            keys = APIKeys()
```
kongzii, line 195: Could you refactor out all of this logic inside of this `if` to a function, for example `reedem_all_markets_tx` into `prediction_market_agent_tooling/markets/omen/omen.py`? It's a big chunk of code that serves 1 purpose, so I think it's useful to have it as a reusable function.

prediction_market_agent_tooling/markets/omen/omen_graph_queries.py
gabrielfior: Changed back. Reason was that `first` had a default option, not needed anymore.

prediction_market_agent_tooling/deploy/agent.py
gabrielfior: That's true, but I kindly disagree with this implementation.
The reason for that is that OmenMarket is more of a container (it has just properties and validators). The OmenAgentMarket is a lot more powerful and already contains logic for placing bets, hence I think that the logic for finalizing the betting process (i.e. redeeming positions if bets were correct) seems like a good fit for that class as well.
Happy to discuss this in more detail in the stand-up if needed.

prediction_market_agent_tooling/markets/omen/omen.py
gabrielfior: Sorry, I don't follow the suggestion.
In the current implementation, if there are for some reason NO bets for that market (which cannot occur due to filtering on a subgraph level), the method `was_bet_outcome_correct` will return False, which will make the method `redeemPositions` exit.
Curious about why do you think raising a `ValueError` is beneficial here.

prediction_market_agent_tooling/deploy/agent.py
kongzii: > Happy to discuss this in more detail in the stand-up if needed.

Nope, your reasoning sounds good! Happy with leaving it like this.

prediction_market_agent_tooling/markets/omen/omen.py
kongzii: That's just my sceptical view of the code world 😄 If someone is asking if that bet was correct, and he/she provides 0 bets for the given market, it sounds like a mistake to report. (for example, if we mistakenly provide bets from another market, there is no way to detect it without this change).

>  (which cannot occur due to filtering on a subgraph level)

yep but this is a general method that can be used by anyone anywhere, no guaranties about where are the bets coming from.

---

My reasoning is like this: If this can not happen, nothing bad comes out of this suggestion. But if there will be a bug,  this will help to catch it as soon as possible.

---

wdyt?

gabrielfior: Sounds reasonable, implemented

kongzii:
LGTM, I'll fix the typing

PR #114, opened 2024-03-19
Fix getting manifold market when closeTime has invalid datetime timestamp

evangriffiths:
Manifold market monitoring was failing due to a market having an invalid timestamp:
Screenshot 2024-03-19 at 14 30 56

This patches out code to work around the problem by adding a validator to the `ManifoldMarket` class

## Summary by CodeRabbit

- **Refactor**
	- Enhanced timestamp handling in market data models to ensure validity before conversion.
	- Updated `BaseModel` import to include `field_validator` instead of `validator`.
	- Added `clip_timestamp` method to the `ManifoldMarket` class.

### DIFF
--- a/prediction_market_agent_tooling/markets/manifold/data_models.py
+++ b/prediction_market_agent_tooling/markets/manifold/data_models.py
@@ -2,7 +2,7 @@ import typing as t
 from datetime import datetime
 from decimal import Decimal
 
-from pydantic import BaseModel
+from pydantic import BaseModel, validator
 
 from prediction_market_agent_tooling.benchmark.utils import should_not_happen
 from prediction_market_agent_tooling.gtypes import Mana, Probability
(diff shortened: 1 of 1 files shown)


prediction_market_agent_tooling/markets/manifold/data_models.py
```diff
@@ -73,6 +73,16 @@ def is_resolved_non_cancelled(self) -> bool:
     def __repr__(self) -> str:
         return f"Manifold's market: {self.question}"
 
+    @validator("closeTime", pre=True)
```
kongzii, line 76:
```suggestion
    @field_validator("closeTime", mode="before")
```

prediction_market_agent_tooling/markets/manifold/data_models.py
```diff
@@ -2,7 +2,7 @@
 from datetime import datetime
 from decimal import Decimal
 
-from pydantic import BaseModel
+from pydantic import BaseModel, validator
```
kongzii, line 5:
```suggestion
from pydantic import BaseModel, field_validator
```

kongzii (commented):
They deprecated `validator`, otherwise LGTM

kongzii (approved):

PR #121, opened 2024-03-21
Resolving replications -- first half, mainly Polymarket fetching logic

kongzii:
## Summary by CodeRabbit

- **New Features**
	- Introduced support for enhanced market resolution methods across various prediction market platforms.
	- Added new utility functions for resolving markets and searching for market resolutions on Manifold, Omen, and Polymarket platforms.
	- Implemented a Google search function to aid in market research and resolution.
- **Enhancements**
	- Updated the handling of market URLs and base URLs for Manifold, Omen, and Polymarket to improve integration and data fetching.
	- Added new data models for Polymarket to accurately represent market-related data based on website responses.
- **Refactor**
	- Refactored market type retrieval and import logic for better efficiency and readability.
- **Tests**
	- Added tests for Polymarket utilities, including URL finding and market resolution determination.


prediction_market_agent_tooling/tools/google.py
```diff
@@ -0,0 +1,49 @@
...
+
+from prediction_market_agent_tooling.config import APIKeys
+from prediction_market_agent_tooling.tools.cache import persistent_inmemory_cache
+
+
+@tenacity.retry(
+    stop=tenacity.stop_after_attempt(3), wait=tenacity.wait_fixed(1), reraise=True
+)
+@persistent_inmemory_cache
+def search_google(
```
kongzii, line 14: Not sure how you will feel about having google search in PMAT. But I now think that market replication should be an agent in PMA (not PMAT), and I would like to move it, but first, I'd like to finish it in this repo.

Afterwards, this can be removed, or we can keep it here as a handy tool.

kongzii: Also, this google searching seems to be effective, but having state with replicated URLs would be even better of course. I think once we have agents with state, it should be used here.

prediction_market_agent_tooling/markets/omen/omen_resolve_replicated.py
```diff
@@ -0,0 +1,81 @@
...
+    # TODO: Finish this.
+    pass
+
+
+def find_resolution_on_other_markets(market: OmenMarket) -> Resolution | None:
+    resolution: Resolution | None = None
+
+    for market_type in MarketType:
+        # We are going to resolve it on Omen, so we can't find the answer there.
+        if market_type == MarketType.OMEN:
```
gabrielfior, line 64: This is more of a stylistic approach, but since Py 3.10 there is the match command (https://docs.python.org/3/whatsnew/3.10.html#pep-634-structural-pattern-matching), which in my opinion makes sense here. What do you think?

prediction_market_agent_tooling/markets/polymarket/data_models_web.py
```diff
@@ -0,0 +1,375 @@
...
+    liquidityAmm: float
+    liquidityClob: float
+    makerBaseFee: int | None
+    takerBaseFee: int | None
+    negRisk: t.Any | None
+    negRiskRequestID: t.Any | None
+    negRiskMarketID: t.Any | None
+    events: list[Event]
+    markets: list[Market1]
+    lower_bound_date: t.Any | None
```
gabrielfior, line 148: For my understanding - why is t.Any | None favorable vs t.Optional[any]?

prediction_market_agent_tooling/markets/omen/omen_resolve_replicated.py
kongzii: Personally, I don't care about what someone uses, `if ... elif ... else` or `match`. Similarly with other Python constructs, as long as they are valid, they don't bring bugs, and there isn't any clear advantage or disadvantage, then I don't see a reason to force one or the other. 

But sure, why not, switched to `match` here.

prediction_market_agent_tooling/markets/omen/omen_resolve_replicated.py
```diff
@@ -0,0 +1,81 @@
...
+
+        else:
+            raise ValueError(
+                f"Unknown market type {market_type} in replication resolving."
+            )
+
+        if resolution is not None:
+            return resolution
+
+    return resolution
```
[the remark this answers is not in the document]
kongzii: Not sure what do you mean, but if you talk about

```
if resolution is not None:
            return resolution
```

then we need this extra return statement, because we want to return as early as possible. If the answer is found on manifold, there is no need to search for it on polymarket.

Another option would be to break the loop, but the final outcome is the same.

prediction_market_agent_tooling/markets/polymarket/data_models_web.py
kongzii: `SomeType | None` is the same thing as `t.Optional[SomeType]`, which is the same as `t.Union[SomeType, None]`.

`|` was introduced later on, and `Type1 | Type2 | Type3 | None` is the same as `t.Optional[t.Union[Type1, Type2, Type3]]`.

I like to use `|` because you don't need to import typing library in order to use it and it saves a few hits on the keyboard 😄 

---

Btw in this case, it could be just `t.Any` (because `Any` means literally `Any`, so including `None`). I just wanted to be explicit that we don't know what type these fields are, but we do know they can be `None`.

prediction_market_agent_tooling/markets/omen/omen_resolve_replicated.py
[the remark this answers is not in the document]
gabrielfior: Thanks for clarifying. Agreed.

prediction_market_agent_tooling/markets/polymarket/data_models_web.py
```diff
@@ -0,0 +1,375 @@
...
+        elif (
+            outcome_to_outcome_price[POLYMARKET_TRUE_OUTCOME] == 0.0
+            and outcome_to_outcome_price[POLYMARKET_FALSE_OUTCOME] == 1.0
+        ):
+            return Resolution.NO
+
+        else:
+            raise ValueError(
+                f"Unexpected outcome prices {outcome_to_outcome_price} or outcomes {self.outcomes}, please debug."
+            )
```
gabrielfior, line 197: I find this resolution logic hard to follow. Could you please add some explanatory comments? Thanks

kongzii: Added, indeed polymarket is hard to follow overall 😄

gabrielfior (approved):

PR #124, opened 2024-03-22
Allow `answer_binary_market` to return None

kongzii:
## Summary by CodeRabbit

- **Refactor**
	- Updated handling of market answers in the prediction market agent tooling to allow skipping markets without answers.

### DIFF
--- a/prediction_market_agent_tooling/deploy/agent.py
+++ b/prediction_market_agent_tooling/deploy/agent.py
@@ -52,7 +52,7 @@ class DeployableAgent:
         """
         return markets[:1]
 
-    def answer_binary_market(self, market: AgentMarket) -> bool:
+    def answer_binary_market(self, market: AgentMarket) -> bool | None:
         """
         Answer the binary market. This method must be implemented by the subclass.
         """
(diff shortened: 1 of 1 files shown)


prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -52,7 +52,7 @@ def pick_markets(self, markets: list[AgentMarket]) -> list[AgentMarket]:
         """
         return markets[:1]
 
-    def answer_binary_market(self, market: AgentMarket) -> bool:
+    def answer_binary_market(self, market: AgentMarket) -> bool | None:
```
gabrielfior, line 55: What do you think of having typing.Optional[bool] as return type?

kongzii: It is the same thing. Is there any reason not to use the new syntax?

gabrielfior: Consistency. So let me rephrase:
-> This syntax is great.
-> We should also fix the other occurances of t.Optional[X] for the new syntax. I can create an issue for that.

PR #133, opened 2024-03-25
Add subgrounds support

gabrielfior:
- Add subgrounds support to Omen and Realitio
- Splitted Omen tests into `test_omen.py` and `test_omen_subgraph_handler.py`

## Summary by CodeRabbit

- **New Features**
	- Introduced `OmenSubgraphHandler` for enhanced interaction with Omen markets and bets, improving data retrieval efficiency.
	- Added a new utility function for converting datetime to integer timestamp.
- **Refactor**
	- Updated various components to utilize `OmenSubgraphHandler`, streamlining market and bet information retrieval.
	- Improved type validation and annotation for handling data models more robustly.
- **Tests**
	- Added new tests for `OmenSubgraphHandler` functionality, ensuring reliability in fetching and filtering market data.
	- Refactored existing tests to align with the latest changes in data retrieval methods.


prediction_market_agent_tooling/markets/omen/data_models.py
```diff
@@ -79,7 +92,7 @@ class OmenMarket(BaseModel):
     collateralVolume: Wei
     usdVolume: USD
     collateralToken: HexAddress
-    outcomes: list[str]
+    outcomes: Possibly_Empty_List_of_Strings
```
gabrielfior, line 95: This was required since it was not possible to convert `None` into `list[str]`

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -0,0 +1,307 @@
...
+        """subgrounds return a weird key as a dict key"""
+        items = []
+        for result_chunk in result:
+            for k, v in result_chunk.items():
+                # subgrounds might pack all items as a list, indexed by a key, or pack it as a dictionary (if one single element)
+                if type(v) is dict:
+                    items.extend([v])
+                else:
+                    items.extend(v)
+        return items
```
gabrielfior, line 230: I apologize for the excessive use of t.Any, but the format returned by subgrounds is not trivial - it returns a dict with a single hex key (like `h7f1de`), and belonging to that key one might get either a. a list of dicts or b. a single dict, hence the complicated logic inside this function.

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -0,0 +1,302 @@
...
+    OMEN_TRADES_SUBGRAPH = "https://api.thegraph.com/subgraphs/name/protofire/omen-xdai"
+    CONDITIONAL_TOKENS_SUBGRAPH = (
+        "https://api.thegraph.com/subgraphs/name/gnosis/conditional-tokens-gc"
+    )
+    REALITYETH_GRAPH_URL = (
+        "https://api.thegraph.com/subgraphs/name/realityeth/realityeth-gnosis"
+    )
+
+    # We define here as str for easier filtering.
+    INVALID_ANSWER_STR = HexBytes(
```
kongzii, line 37: It's called `_STR`, but the type is `bytes` (HexBytes). Before, this was written as 

```
INVALID_ANSWER = 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
```

so the type was `string` and not `bytes`.

```suggestion
    INVALID_ANSWER_STR = HexStr(
```

Could you double check or just correct me if you know already, which one is correct, please?

We can either remove `_STR` suffix or change the type.

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -0,0 +1,302 @@
...
+    def __init__(self) -> None:
+        self.sg = Subgrounds()
+        # Load the subgraph
+        self.trades_subgraph = self.sg.load_subgraph(self.OMEN_TRADES_SUBGRAPH)
+        self.conditional_tokens_subgraph = self.sg.load_subgraph(
+            self.CONDITIONAL_TOKENS_SUBGRAPH
+        )
+        self.realityeth_subgraph = self.sg.load_subgraph(self.REALITYETH_GRAPH_URL)
+
+    def _get_fields_for_bets(self, bets_field: t.Any) -> list[FieldPath]:
```
kongzii, line 50: There are two ways to find out what type something should be:

With mypy, you can add `reveal_type(x)` to the code and run `mypy`, it will print it out.

However, that works only if the library is typed, if it isn't, you can do it in runtime, for example:

```python
print(
    type(OmenSubgraphHandler().trades_subgraph.Query.fixedProductMarketMaker(id="bla"))
)
```

which will print

```
<class 'subgrounds.subgraph.fieldpath.FieldPath'>
```

So

```suggestion
    def _get_fields_for_bets(self, bets_field: FieldPath) -> list[FieldPath]:
```

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -0,0 +1,302 @@
...
+            bets_field.collateralAmount,
+            bets_field.collateralAmountUSD,
+            bets_field.feeAmount,
+            bets_field.outcomeIndex,
+            bets_field.outcomeTokensTraded,
+            bets_field.transactionHash,
+        ]
+        return fields_for_bets + fields_for_markets
+
+    def _get_fields_for_answers(self, answers_field: t.Any) -> list[FieldPath]:
```
kongzii, line 72:
```suggestion
    def _get_fields_for_answers(self, answers_field: FieldPath) -> list[FieldPath]:
```

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -0,0 +1,302 @@
...
+            answers_field.question.id,
+            answers_field.question.user,
+            answers_field.question.updatedTimestamp,
+            answers_field.question.questionId,
+            answers_field.bondAggregate,
+            answers_field.lastBond,
+            answers_field.timestamp,
+        ]
+
+    def _get_fields_for_markets(self, markets_field: t.Any) -> list[FieldPath]:
```
kongzii, line 85:
```suggestion
    def _get_fields_for_markets(self, markets_field: FieldPath) -> list[FieldPath]:
```

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -0,0 +1,302 @@
...
+        ]
+
+        if creator:
+            where_stms.append(fpmm.creator == creator)
+
+        if created_after:
+            where_stms.append(fpmm.creationTimestamp > to_int_timestamp(created_after))
+
+        if opened_before:
+            where_stms.append(fpmm.openingTimestamp > to_int_timestamp(opened_before))
```
kongzii, line 135: So cool!

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -0,0 +1,302 @@
...
+
+        if filter_by == FilterBy.RESOLVED:
+            where_stms.append(fpmm.answerFinalizedTimestamp != None)
+            where_stms.append(fpmm.currentAnswer != None)
+            # We cannot add the same type of filter twice, it gets overwritten, hence we use nested filter.
+            where_stms.append(
+                fpmm.question.currentAnswer
+                != "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"
```
kongzii, line 143: Just checking if `INVALID_ANSWER_STR` shouldn't be used here? (probably after fixing it to `HexStr`)

`INVALID_ANSWER_STR` seems unused now.

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -0,0 +1,302 @@
...
+            )
+        elif filter_by == FilterBy.OPEN:
+            where_stms.append(fpmm.currentAnswer == None)
+
+        if excluded_questions is not None:
+            for question_title in excluded_questions:
+                where_stms.append(fpmm.question.title != question_title)
```
kongzii, line 150: Previously you said that we can't use one filter twice because it will get overwritten. Is this somehow different for some reason?

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -0,0 +1,302 @@
...
+
+        return where_stms
+
+    def _build_sort_direction(self, sort_by: SortBy) -> str:
+        sort_direction = "asc"
+        match sort_by:
+            case SortBy.NEWEST:
+                sort_direction = "desc"
+            case _:
+                pass
```
kongzii, line 160: Logic was different before:

```
def ordering_from_sort_by(sort_by: SortBy) -> tuple[str, str]:
    """
    Returns 'orderBy' and 'orderDirection' strings for the given SortBy.
    """
    if sort_by == SortBy.CLOSING_SOONEST:
        return "creationTimestamp", "desc"  # TODO make more accurate
    elif sort_by == SortBy.NEWEST:
        return "creationTimestamp", "desc"
    else:
        raise ValueError(f"Unknown sort_by: {sort_by}")
```

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -0,0 +1,302 @@
...
+        return omen_markets
+
+    def get_omen_market(self, market_id: HexAddress) -> OmenMarket:
+        markets = self.trades_subgraph.Query.fixedProductMarketMaker(id=market_id)
+        fields = self._get_fields_for_markets(markets)
+        result: t.Any = self.sg.query_json(fields)
+        items = self._parse_items_from_json(result)
+        omen_markets = [OmenMarket.model_validate(i) for i in items]
+
+        if len(omen_markets) > 1:
```
kongzii, line 208:
```suggestion
        if len(omen_markets) != 1:
```

Could be 0 as well

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
kongzii: Yep I get the Any in these cases. But left a few suggestions as there are some assumptions in this function.

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -0,0 +1,302 @@
...
+            raise ValueError(
+                f"Fetched wrong number of markets. Expected 1 but got {len(omen_markets)}"
+            )
+
+        return omen_markets[0]
+
+    def _parse_items_from_json(self, result: t.Any) -> t.List[t.Any]:
```
kongzii, line 215:
```suggestion
    def _parse_items_from_json(self, result: list[dict[str, t.Any]]) -> t.List[dict[str, t.Any]]:
```

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -0,0 +1,302 @@
...
+        """subgrounds return a weird key as a dict key"""
+        items = []
+        for result_chunk in result:
+            for k, v in result_chunk.items():
+                # subgrounds might pack all items as a list, indexed by a key, or pack it as a dictionary (if one single element)
+                if type(v) is dict:
```
kongzii, line 221:
```suggestion
                if isinstance(v, dict):
```

See https://chat.openai.com/share/182af04d-62b0-4a58-9971-67b4ba9390b0

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -0,0 +1,302 @@
...
+    ) -> list[OmenBet]:
+        if not end_time:
+            end_time = utcnow()
+
+        trade = self.trades_subgraph.FpmmTrade
+        where_stms = [
+            trade.type == "Buy",
+            trade.creator == better_address.lower(),
+            trade.creationTimestamp >= to_int_timestamp(start_time),
+            trade.creationTimestamp <= to_int_timestamp(end_time),
```
kongzii, line 260: Again, I thought this won't work as it will overwrite the filter?

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -595,11 +598,15 @@ def redeem_positions_from_all_omen_markets() -> None:
     Redeems positions from all resolved Omen markets.
     """
     keys = APIKeys()
-    resolved_omen_bets = get_resolved_omen_bets(
+    omen_subgraph_handler = OmenSubgraphHandler()
+    bets = omen_subgraph_handler.get_bets(
```
kongzii, line 602: Couldn't we use `get_resolved_bets` here?

```suggestion
    bets = omen_subgraph_handler.get_resolved_bets(
```

prediction_market_agent_tooling/markets/omen/data_models.py
gabrielfior: https://api.thegraph.com/subgraphs/name/protofire/omen-xdai/graphql?query=query+%7B%0A++fixedProductMarketMakers%28first%3A5%2C+where%3A+%7B%0A++++outcomes%3A+null%0A++%7D%29+%7B%0A++++id%0A++++title%0A++%7D%0A%7D

```json
fixedProductMarketMakers": [
      {
        "id": "0x013e60bf17417f35f98a15b8cf2e3656bf41b2ce",
        "title": "Will there be a ceasfire in Ukraine in the next< 3 weeks",
        "outcomes": null
      },
      {
        "id": "0x05afc073f58d43b17d4531b7717df213df6d498c",
        "title": "What will be the total amount of \"Base Fee\" ETH burnt at the end of August 2021? [ETH]",
        "outcomes": null
      },
      {
        "id": "0x071bc23726c6363e2c017185ff431c4d2d9ddf8d",
        "title": "Scalar market? [USD]",
        "outcomes": null
      },
```

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
gabrielfior: Completely missed this type! Thanks, adding now

prediction_market_agent_tooling/markets/omen/omen.py
gabrielfior: good catch

prediction_market_agent_tooling/markets/omen/data_models.py
kongzii: Interesting! So these are two scalar markets, and one market with only one option: `Invalid`.

I think that before, we were using get_markets always with `outcomes=[OMEN_TRUE_OUTCOME, OMEN_FALSE_OUTCOME]` filter, so that's why we didn't see them.

I think we can continue with this assumption that we always use `get binary omen markets`, otherwise there would be a lot of type-correcting around the codebase and that's not required anyway right now.

gabrielfior: The reason why I went with the pydantic Validator instead of list[str] | None (which is the correct type) was not to adapt the usages of outcomes to the case where it's None.
But I think this case it's the correct thing to implement, so adapting this.

gabrielfior: Per our discussion, adding another filter for avoiding having outcomes `null`.

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
gabrielfior: Filters are overwritten if they are the same tuple, a tuple being (COLUMN, operator), like (market.id, 'eq'). In this case, we are using different operators, hence it's "translated" to different filters, see below
```
query {
  x8438f62a8a475d79: fpmmTrades(first: 9223372036854775807, where: {type: Buy, creator: "0x95aa7cc38f8ff36efecf2fefa6f4850320dd32f7", creationTimestamp_gte: "1625073159", creationTimestamp_lte: "1625073162"}) {
```

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -595,11 +598,13 @@ def redeem_positions_from_all_omen_markets() -> None:
...
     keys = APIKeys()
-    resolved_omen_bets = get_resolved_omen_bets(
-        start_time=datetime(2020, 1, 1),
-        end_time=None,
+    omen_subgraph_handler = OmenSubgraphHandler()
+    bets = omen_subgraph_handler.get_resolved_bets(
         better_address=keys.bet_from_address,
+        start_time=datetime(2020, 1, 1),
     )
+    resolved_omen_bets = [b for b in bets if b.fpmm.is_resolved]
```
kongzii, line 606:
```suggestion
```

As we are using `get_resolved_bets` above, this should be unnecessary, right?

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -595,11 +598,13 @@ def redeem_positions_from_all_omen_markets() -> None:
     Redeems positions from all resolved Omen markets.
     """
     keys = APIKeys()
-    resolved_omen_bets = get_resolved_omen_bets(
-        start_time=datetime(2020, 1, 1),
-        end_time=None,
+    omen_subgraph_handler = OmenSubgraphHandler()
+    bets = omen_subgraph_handler.get_resolved_bets(
```
kongzii, line 602:
```suggestion
    resolved_omen_bets = omen_subgraph_handler.get_resolved_bets(
```

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -0,0 +1,305 @@
...
+
+    OMEN_TRADES_SUBGRAPH = "https://api.thegraph.com/subgraphs/name/protofire/omen-xdai"
+    CONDITIONAL_TOKENS_SUBGRAPH = (
+        "https://api.thegraph.com/subgraphs/name/gnosis/conditional-tokens-gc"
+    )
+    REALITYETH_GRAPH_URL = (
+        "https://api.thegraph.com/subgraphs/name/realityeth/realityeth-gnosis"
+    )
+
+    # We define here as str for easier filtering.
```
kongzii, line 36:
```suggestion
    # We define here for easier filtering.
```

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -0,0 +1,305 @@
...
+        created_after: t.Optional[datetime] = None,
+        opened_before: t.Optional[datetime] = None,
+        excluded_questions: set[str] | None = None,
+    ) -> list[subgrounds.subgraph.Filter]:
+        fpmm = self.trades_subgraph.FixedProductMarketMaker
+        where_stms = [
+            fpmm.isPendingArbitration == False,
+            fpmm.outcomes == outcomes,
+            fpmm.title != None,
+            fpmm.outcomes != None,  # we always want non-null outcomes
```
kongzii, line 126:
```suggestion
```

Seems unnecessary as there is `fpmm.outcomes == outcomes,` above and `outcomes` is typed to never be null.

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
kongzii: Makes sense, thanks!

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
gabrielfior: thanks, fixed this

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
gabrielfior: Excellent catch. It was indeed being overwritten - wrote a test for it `test_get_markets_exclude_questions`, now working.

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
gabrielfior: Currently it always sorts by creationTimestamp, and we have 2 cases
-> if sort_by == NEWEST, return in desc order (newest markets first)
-> if sort_by != NEWEST, returns in asc order (oldest markets first). This was my interpretation of the sort_by for getting the markets that are closing_soonest, since there is no indication of when a market will close (from the subgraph), and I infer that the oldest markets have highest likelihood of closing sooner.
Open to changes for this method though.

kongzii: > and I infer that the oldest markets have highest likelihood of closing sooner.

I didn't notice this before! Yep, it makes total sense that the older market will close sooner than the new market.

However, in that case, let's do:

```python
match sort_by:
            case SortBy.NEWEST:
                sort_direction = "desc"
            case SortBy.CLOSING_SOONEST:
                sort_direction = "asc"
            case _:
                raise ValueError(f"Unknown sort_by: {sort_by}") 
```

I should have added more context before, sorry:

My main concert was not raising the exception. Because for example, Polymarket doesn't support any sorting at all:

https://github.com/gnosis/prediction-market-agent-tooling/blob/513f85cc55da6347758be822c1e16f1c5fd6b5fe/prediction_market_agent_tooling/markets/polymarket/polymarket.py#L58

And if we add a new SortBy in the future, and we won't have this `raise Value` error here, code will "work" without any complaints, but we will be unknowingly processing different markets.

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
gabrielfior: True, removing that

prediction_market_agent_tooling/markets/omen/omen.py
kongzii: I did take a look and it seems that the method `get_resolved_bets` is doing exact same thing as this line that I'm suggesting to remove.

Am I missing something?

Screenshot by Dropbox Capture

gabrielfior: My bad, this logic is indeed duplicated. Removed this.

kongzii (approved):
Cooool!

PR #140, opened 2024-03-28
Claim bonded xDai on Realitio as part of replicator

kongzii:
## Summary by CodeRabbit

- **New Features**
	- Added functionality for claiming bonds on Realitio questions and handling withdrawals.
	- Enhanced market resolution process to include claiming back markets based on others' transactions.
- **Enhancements**
	- Improved logging messages in the `finalize_markets` function.
	- Added new properties (`contentHash`, `url`, `id`, `answer`, `createdBlock`) to data models for better tracking and interaction with prediction markets.
- **Refactor**
	- Renamed a function for clarity in its extended functionality.
	- Updated various tools and scripts for consistency and to utilize new features and methods.


prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -135,6 +140,19 @@ def _build_where_statements(
         if opened_before:
             where_stms["openingTimestamp_lt"] = to_int_timestamp(opened_before)
 
+        if filter_by == FilterBy.RESOLVED:
+            finalized = True
+            resolved = True
```
kongzii, line 145: I needed more granual filtering so I split the original `FilterBy.RESOLVED` into these two flags.

Theoretically, only resolved=True should be needed, but we were doing `where_stms["answerFinalizedTimestamp_not"] = None` before as well, so just to be safe I included `finalized=True` as well here.

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -146,21 +164,6 @@ def _build_where_statements(
...
-                where_stms["resolutionTimestamp_not"] = None
-            else:
-                where_stms["resolutionTimestamp"] = None
-
-        where_stms["question_"] = {}
-        if filter_by == FilterBy.RESOLVED:
-            where_stms["answerFinalizedTimestamp_not"] = None
-            where_stms["currentAnswer_not"] = None
-            # We cannot add the same type of filter twice, it gets overwritten, hence we use nested filter.
-            where_stms["question_"]["currentAnswer_not"] = self.INVALID_ANSWER
```
kongzii, line 160: Technically, market can be resolved as invalid. So invalid resolution doesn't mean market isn't resolved. But we will see if removal of 

```
where_stms["question_"]["currentAnswer_not"] = self.INVALID_ANSWER
```

won't cause a problem. 🙏

prediction_market_agent_tooling/markets/omen/omen_resolve_replicated.py
```diff
@@ -69,14 +90,112 @@ def omen_finalize_and_resolve_all_markets_based_on_others_tx(
...
+            # This is from the original Olas implementation (https://github.com/kongzii/trader/blob/700af475a4538cc3d5d22caf9dec9e9d22d72af1/packages/valory/skills/market_manager_abci/graph_tooling/requests.py#L297),
+            # but it's most probably wrong (see comment above).
+            history_hashes.append(answers_objects[i + 1].question.historyHash)
+
+        # last-to-first, the address of each answerer or commitment sender
+        addresses.append(Web3.to_checksum_address(answer.question.user))
+        # last-to-first, the bond supplied with each answer or commitment
+        bonds.append(answer.lastBond)
+        # last-to-first, each answer supplied, or commitment ID if the answer was supplied with commit->reveal
+        answers.append(answer.answer)
```
gabrielfior, line 181: From the comments I couldn't 100% follow - do we have a valid working solution (I see you based your implementation off of omen-exchange UI repo) or is further clarification needed?
If working, I suggest rephrasing the comments to make it clear where improvements can be made and/or which points are still unclear.

prediction_market_agent_tooling/tools/gnosis_rpc.py
```diff
@@ -2,13 +2,13 @@
...
 
-from prediction_market_agent_tooling.gtypes import ChainID, Wei
+from prediction_market_agent_tooling.gtypes import ChainID, HexAddress, Wei
 
 GNOSIS_NETWORK_ID = ChainID(100)  # xDai network.
 GNOSIS_RPC_URL = os.getenv("GNOSIS_RPC_URL", "https://gnosis-rpc.publicnode.com")
 
 
-def get_balance(address: str) -> Wei:
+def get_balance(address: HexAddress) -> Wei:
```
gabrielfior, line 11: I would expect a web3 argument here, so that we can use that instead of the default GNOSIS_RPC_URL.
This will be useful afterwards when we have local blockchain instances.

prediction_market_agent_tooling/markets/omen/data_models.py
```diff
@@ -310,19 +310,30 @@ class FixedProductMarketMakersResponse(BaseModel):
...
+    @property
+    def url(self) -> str:
+        return f"https://reality.eth.limo/app/#!/question/{self.id}"
+
 
 class RealityAnswer(BaseModel):
-    answer: str
-    bondAggregate: Wei
-    lastBond: Wei
+    id: str
```
gabrielfior, line 330: Here as well, I would expect the fields on `omen_subgraph_handler` to be expanded by these new fields.

prediction_market_agent_tooling/tools/gnosis_rpc.py
kongzii: This is an older function, I just fixed type as I noticed it.

It uses requests instead of Web3

prediction_market_agent_tooling/markets/omen/omen_resolve_replicated.py
kongzii: What is not clear?

This works for questions with just 1 answer, otherwise it will raise NotImplementedError:

```
if len(answers_objects) > 1:
        # As you can see below, we need `history_hash` for every answer.
        # The trouble is, that historyHash is updated after each new answer and the contract holds only the latest one.
        # So if we have more than 1 answer, we missing the historyHash n-1 of them and this would fail.
        # You can find how to calculate history hash at https://realitio.github.io/docs/html/contract_explanation.html#answer-history-entries.
        # At the moment, we support only 1 answer, as for that one answer we will have the hash.
        raise NotImplementedError()

```

prediction_market_agent_tooling/markets/omen/omen_resolve_replicated.py
```diff
@@ -69,14 +90,112 @@ def omen_finalize_and_resolve_all_markets_based_on_others_tx(
...
+def claim_bonds_on_realitio_question(
+    question: RealityQuestion,
+    from_private_key: PrivateKey,
+    auto_withdraw: bool,
+) -> None:
+    public_key = private_key_to_public_key(from_private_key)
+    realitio_contract = OmenRealitioContract()
+
+    # Get all answers for the question.
+    answers_objects = OmenSubgraphHandler().get_answers(question_id=question.questionId)
```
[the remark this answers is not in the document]
kongzii: It needs to be initialised somewhere anyway; it doesn't really matter if it's done here or in some wrapped function.

So I'd just keep it. If the best practice is to remove singleton from it, let's issue that.

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
kongzii: I tested it and it seems fine:

```python
from prediction_market_agent_tooling.markets.omen.omen import (
    get_omen_binary_markets,
    SortBy,
    FilterBy,
)
x = get_omen_binary_markets(None, SortBy.NONE, FilterBy.NONE)
print(len(x))
```

output: `3175`.

currentAnswer is typed as `str | None`, so that should accept invalid answer without any problems:

https://github.com/gnosis/prediction-market-agent-tooling/blob/366d9f508d7706795ccc8bc74fecefd629751a62/prediction_market_agent_tooling/markets/omen/data_models.py#L100

gabrielfior: Understood. Happy with this then.

prediction_market_agent_tooling/markets/omen/omen_resolve_replicated.py
gabrielfior:
```
 else:
            # TODO: See `if len(answers_objects) > 1` above.
            # This is from the original Olas implementation (https://github.com/kongzii/trader/blob/700af475a4538cc3d5d22caf9dec9e9d22d72af1/packages/valory/skills/market_manager_abci/graph_tooling/requests.py#L297),
            # but it's most probably wrong (see comment above).
            history_hashes.append(answers_objects[i + 1].question.historyHash)
```
The part above to me is not clear. If the code runs into this `else` condition, then it seems to me that a wrong historyHash is appended to the history hashes, although it's the same logic as in Olas. If that's the case (that the implementation is wrong in Olas), then we should clarify that before merging.
Please let me know if this is clear otherwise happy to chat.

prediction_market_agent_tooling/markets/omen/omen_resolve_replicated.py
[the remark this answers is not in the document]
gabrielfior: Created an issue and leaving this as is. OK.

prediction_market_agent_tooling/markets/omen/data_models.py
gabrielfior: Nope, seems fine too me, my mistake.

prediction_market_agent_tooling/markets/omen/omen_resolve_replicated.py
kongzii: It's the case, but what do you want to clarify?

There is the `if` statement above that I mentioned that will `raise NotImplementedError`, so this code won't be executed. If we ever find the need for it, we can circle back to it.

PR #149, opened 2024-04-02
Add prob_uncertainty util

kongzii:
## Summary by CodeRabbit

- **New Features**
    - Added a new function `prob_uncertainty` for calculating market uncertainty in the prediction market tooling.
- **Refactor**
    - Enhanced timezone validation in utility tools.
- **Tests**
    - Included a test function `test_prob_uncertainty` to validate the `prob_uncertainty` function.

### DIFF
--- a/prediction_market_agent_tooling/tools/utils.py
+++ b/prediction_market_agent_tooling/tools/utils.py
@@ -9,6 +9,7 @@ import pytz
 import requests
 from loguru import logger
 from pydantic import BaseModel, ValidationError
+from scipy.stats import entropy
 
 from prediction_market_agent_tooling.gtypes import (
     DatetimeWithTimezone,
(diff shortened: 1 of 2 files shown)


evangriffiths (commented):
The function you've got there looks a lot like https://en.wikipedia.org/wiki/Binary_entropy_function. A market isn't exactly a bernoulli process (because it's not a repeated by sampling from a random variable), but I think this is still a good fit. You can calculate it in python with:

```
from scipy.stats import entropy
p = 0.3
entropy([p, 1-p])
```

kongzii:
> The function you've got there looks a lot like https://en.wikipedia.org/wiki/Binary_entropy_function. A market isn't exactly a bernoulli process (because it's not a repeated by sampling from a random variable), but I think this is still a good fit. You can calculate it in python with:
> 
> ```
> from scipy.stats import entropy
> p = 0.3
> entropy([p, 1-p])
> ```

It's doing something similar, but 0.5 is far away from the expected result:

```
>>> p = 0.5; entropy([p, 1 - p])
0.6931471805599453
```

So we would need to scale it between [0.0, 1.0] after using entropy.

Do you like that approach mode than the current one?

(mine is GPT generated so I don't have a strong opinion 😄 )

evangriffiths:
Ah you can scale it with `entropy([p, 1 - p], base=2)`,

```
>>> entropy([0.5, 0.5], base=2)
1.0
```

evangriffiths (approved):

PR #153, opened 2024-04-02
Update function signatures for answer_binary_market

evangriffiths:
## Summary by CodeRabbit

- **Refactor**
	- Modified the `answer_binary_market` method in the prediction market agent tooling to allow returning `None` in addition to `bool`, enhancing decision-making flexibility for agents.


evangriffiths:
Just imagining a careless developer (i.e. me in the future!) copy-pasting `DeployableCoinFlipAgent`'s definition when making a new agent, without looking at the base class, and therefore just assuming that `answer_binary_market` can't return `None`.

PR #154, opened 2024-04-02
Fix removeFunds logic on Omen

gabrielfior:
Added extra logic for fetching minimum amount we should pass when removing funds

## Summary by CodeRabbit

- **Refactor**
	- Improved the `get_conditional_tokens_balance_for_market` function to return detailed balance information.
	- Enhanced the `omen_remove_fund_market_tx` function for better position merging based on minimum balances.

### DIFF
--- a/prediction_market_agent_tooling/markets/omen/omen.py
+++ b/prediction_market_agent_tooling/markets/omen/omen.py
@@ -528,7 +528,10 @@ def omen_redeem_full_position_tx(
         logger.debug("Cannot redeem winnings if market is not yet resolved. Exiting.")
         return
 
-    amount_wei = get_conditional_tokens_balance_for_market(market, from_address, web3)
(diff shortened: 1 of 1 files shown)


prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -528,7 +528,10 @@ def omen_redeem_full_position_tx(
         logger.debug("Cannot redeem winnings if market is not yet resolved. Exiting.")
         return
 
-    amount_wei = get_conditional_tokens_balance_for_market(market, from_address, web3)
+    amount_per_index = get_conditional_tokens_balance_for_market(
```
gabrielfior, line 531: This method has been changed so that balances are returned per index set, not summed - this allows the method to be reused

kongzii (approved):

PR #155, opened 2024-04-02
Gabriel/omen subgraph handler singleton

gabrielfior:
Made the Subgraph a Singleton, since it takes some time to load and does not hold state differentiable between instances.

## Summary by CodeRabbit

- **Refactor**
	- Improved the `OmenSubgraphHandler` class for better resource management.
	- Added a metaclass `SingletonMeta` for enhanced functionality.

### DIFF
--- a/prediction_market_agent_tooling/tools/singleton.py
+++ b/prediction_market_agent_tooling/tools/singleton.py
@@ -1,13 +1,18 @@
-class SingletonMeta(type):
+import typing as t
+
+_T = t.TypeVar("_T")
+
+
+class SingletonMeta(type, t.Generic[_T]):
     """
     The Singleton class can be implemented in different ways in Python. Some
     possible methods include: base class, decorator, metaclass. We will use the
     metaclass because it is best suited for this purpose.
     """
 
-    _instances = {}
+    _instances: dict[t.Any, _T] = {}
 
-    def __call__(cls, *args, **kwargs):
+    def __call__(cls, *args: t.Any, **kwargs: t.Any) -> _T:
         """
         Possible changes to the value of the `__init__` argument do not affect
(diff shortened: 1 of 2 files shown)


prediction_market_agent_tooling/tools/singleton.py
```diff
@@ -0,0 +1,23 @@
...
+class SingletonMeta(type, t.Generic[_T]):
+    """
+    The Singleton class can be implemented in different ways in Python. Some
+    possible methods include: base class, decorator, metaclass. We will use the
+    metaclass because it is best suited for this purpose.
+    """
+
+    _instances: dict[t.Any, _T] = {}
+
+    def __call__(cls, *args: t.Any, **kwargs: t.Any) -> _T:
```
kongzii, line 15: Shouldn't `cls` be `self` as this is not marked to be `@classmethod`?

kongzii: Screenshot by Dropbox Capture

gabrielfior: It does work with self as well, but this is just a matter of naming convention.
This is a tricky one, since PyCharm suggests cls, the reference where I took this from (https://refactoring.guru/design-patterns/singleton/python/example#example-0) also uses cls, but ChatGpt suggests self.
I accept your argument that `cls` is normally used in a classmethod, so I will change this.

image

kongzii: Yeah, `self` and `cls` is just naming conventions, but with a strong difference in the code about what the variable really holds. Python even introduced `Self` type hint: https://realpython.com/python-type-self/

kongzii (approved):

PR #158, opened 2024-04-04
Refactor redeeming of bets in agent

kongzii:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new way to access bet redemption using private keys.
	- Added functionality to check if a user's position is redeemable in the prediction market.
	- Enhanced market analysis with the ability to retrieve positions based on condition ID.

- **Improvements**
	- Improved clarity in function names related to bet outcome verification and redemption.
	- Optimized logic for determining if a bet outcome was correct and if a market is redeemable by a specific user.


prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -113,44 +114,41 @@ def was_bet_outcome_correct(self, resolved_omen_bets: t.List[OmenBet]) -> bool:
...
-                bet.fpmm.question.answerFinalizedTimestamp is None
-                or bet.fpmm.question.currentAnswer is None
-            ):
+            if not bet.fpmm.is_resolved:
                 continue
 
             # Like Olas, we assert correctness by matching index OR invalid market answer
             if bet.outcomeIndex == int(
-                bet.fpmm.question.currentAnswer, 16
+                check_not_none(bet.fpmm.question.currentAnswer), 16
```
gabrielfior, line 122: Doesn't this throw an exception if `bet.fpmm.question.currentAnswer` == None?
In the prev case, we simply returned False - I would keep that behavior, or is there a reason to stop execution here?

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -289,30 +289,58 @@ def _parse_items_from_json(
...
 
         if total_balance_bigger_than is not None:
             where_stms["totalBalance_gt"] = total_balance_bigger_than
 
+        if position_id_in is not None:
+            where_stms["position_"]["positionId_in"] = [x.hex() for x in position_id_in]
+
         positions = self.conditional_tokens_subgraph.Query.userPositions(
             first=sys.maxsize, where=where_stms
         )
```
gabrielfior, line 346: This is a great addition to the overall Omen logic! Just one observation regarding the filter `position_id_in`:
-> I'm not 100% sure, but I would expect to be a limit in the number of items you could pack into the array to use as filter (my guess is that until ~ 100 should be fine, more than that I would double-check).
-> For the above reason and also in general, I would expect a test to be added testing this method, inside the test class for the OmenSubgraphHandler.

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -107,44 +108,41 @@ def was_bet_outcome_correct(self, resolved_omen_bets: t.List[OmenBet]) -> bool:
...
             ) or bet.outcomeIndex == int(self.INVALID_MARKET_ANSWER, 16):
                 return True
 
         return False
 
-    def check_if_position_was_already_redeemed(self) -> bool:
+    def market_redeemable_by(self, user: ChecksumAddress) -> bool:
```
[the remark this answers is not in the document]
gabrielfior: Makes ton of sense, very cool

prediction_market_agent_tooling/markets/omen/omen.py
kongzii: Ah I see, before there was 

```
if (
                bet.fpmm.question.answerFinalizedTimestamp is None
                or bet.fpmm.question.currentAnswer is None
            ):
                continue
```

but because I used the `is_resolved` property, mypy doesn't know it can't be None and that's why I needed to add `check_not_none`.

I added a comment about it

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
kongzii: Yep, agreed, It's always nice to have a test. Added.

gabrielfior (approved):

PR #162, opened 2024-04-04
Fix getting open markets from Omen and bump to 0.9.1

kongzii:
## Summary by CodeRabbit

- **Refactor**
	- Improved handling and sorting logic in market predictions tooling for enhanced efficiency and accuracy.

### DIFF
--- a/prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
+++ b/prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
@@ -1,6 +1,6 @@
 import sys
 import typing as t
-from datetime import datetime
+from datetime import datetime, timedelta
 
 from eth_typing import ChecksumAddress
 from subgrounds import FieldPath, Subgrounds
@@ -22,7 +22,6 @@ from prediction_market_agent_tooling.tools.web3_utils import ZERO_BYTES
 
 
 class OmenSubgraphHandler(metaclass=SingletonMeta):
-
     """
     Class responsible for handling interactions with Omen subgraphs (trades, conditionalTokens).
     """
@@ -152,9 +151,6 @@ class OmenSubgraphHandler(metaclass=SingletonMeta):
         if opened_before:
             where_stms["openingTimestamp_lt"] = to_int_timestamp(opened_before)
 
-        if opened_after:
-            where_stms["openingTimestamp_gt"] = to_int_timestamp(opened_after)
-
         if liquidity_bigger_than is not None:
             where_stms["liquidityMeasure_gt"] = liquidity_bigger_than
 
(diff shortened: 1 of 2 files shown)


prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -185,6 +181,9 @@ def _build_where_statements(
             else:
                 where_stms["answerFinalizedTimestamp"] = None
 
+        if opened_after:
```
gabrielfior, line 184: I'm not a big fan of having filterBy + other args (`finalized_before, opened_after`, etc). We are slowly building towards an ORM and that is of course tricky.
I would suggest adding as argument, instead of a simple filterBy Enum, a list of filters, which we iterate upon.
So we could have as input, [filterBy.CLOSED, filterBy.FinalizedBefore(INSERT_DATE)], and that would be cleaner from my perspective.

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -202,7 +201,8 @@ def _build_sort_direction(self, sort_by: SortBy) -> str:
             case SortBy.NEWEST:
                 sort_direction = "desc"
             case SortBy.CLOSING_SOONEST:
-                sort_direction = "asc"
+                # `desc` feel unintuitive, but really if we use `asc`, we are getting markets closing in 2030, 2026, etc.
```
gabrielfior, line 204: In that case, then our sort_direction is always "desc" and we don't need this whole code block, right?

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -165,9 +161,9 @@ def _build_where_statements(
             finalized = True
             resolved = True
         elif filter_by == FilterBy.OPEN:
-            where_stms["currentAnswer"] = None
-            finalized = False
-            resolved = False
+            # We can not use `resolved=False` + `finalized=False` here,
```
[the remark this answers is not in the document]
kongzii: That's true, but if the question is opened on Reality in the future, `current answer` must be None, so no need for that filter anymore.

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
kongzii: Yep agreed, this is getting messy, it's already implemented like this, so I'd vote to do the cleanup in separate PR 🙏

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
kongzii: I like to keep it explicit, also if we add a new value to `SortBy`, it will be good to see what's going on here.

evangriffiths: Might as well keep it here because it's going to be added back in a follow-up PR! :D

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
evangriffiths: Fair point - the enum option was sufficient at the beginning, but is getting less suitable! But we should create an issue for this and address in a separate PR as Peter isn't actually adding a new filter option here.

evangriffiths (approved):
LGTM

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
kongzii: Created a issue with some proposed solution: https://github.com/gnosis/prediction-market-agent-tooling/issues/164

Please comment if you have any other ideas about it 🙏

PR #163, opened 2024-04-04
Fix sorting on 'closing soonest' when getting open markets from Omen

evangriffiths:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new attribute to track market opening times more accurately.
	- Enhanced sorting functionality for markets to prioritize by closing time.

- **Bug Fixes**
	- Adjusted market close time handling to improve accuracy in market status.

- **Tests**
	- Added tests to ensure correct handling of market close times and betting strategies.

### DIFF
--- a/prediction_market_agent_tooling/markets/omen/data_models.py
+++ b/prediction_market_agent_tooling/markets/omen/data_models.py
@@ -112,6 +112,9 @@ class OmenMarket(BaseModel):
     creationTimestamp: t.Optional[int] = None
     condition: Condition
     question: Question
+    openingTimestamp: t.Optional[
+        int
+    ] = None  # Don't be fooled - this is the time that the market closes for trading, and the market question is opened on reality.eth!
 
     @property
     def answer_index(self) -> t.Optional[int]:
--- a/prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
+++ b/prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
@@ -202,8 +202,7 @@ class OmenSubgraphHandler(metaclass=SingletonMeta):
             case SortBy.NEWEST:
                 sort_direction = "desc"
             case SortBy.CLOSING_SOONEST:
-                # `desc` feel unintuitive, but really if we use `asc`, we are getting markets closing in 2030, 2026, etc.
-                sort_direction = "desc"
+                sort_direction = "asc"
             case SortBy.NONE:
                 sort_direction = "desc"
             case _:
--- a/prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
+++ b/prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
@@ -1,6 +1,6 @@
 import sys
 import typing as t
-from datetime import datetime
+from datetime import datetime, timedelta
 
 from eth_typing import ChecksumAddress
 from subgrounds import FieldPath, Subgrounds
@@ -22,7 +22,6 @@ from prediction_market_agent_tooling.tools.web3_utils import ZERO_BYTES
 
 
 class OmenSubgraphHandler(metaclass=SingletonMeta):
-
     """
     Class responsible for handling interactions with Omen subgraphs (trades, conditionalTokens).
     """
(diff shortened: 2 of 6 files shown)


tests/markets/omen/test_omen.py
```diff
@@ -152,3 +152,10 @@ def create_market_fund_market_remove_funding() -> None:
         - Assert amount in xDAI is reflected in user's balance
     """
     assert True
+
+
+def test_omen_market_close_time() -> None:
+    market = OmenAgentMarket.from_data_model(pick_binary_market())
+    assert (
+        market.close_time > market.created_time
+    ), "Market close time should be after open time."
```
gabrielfior, line 161: I would also test the sorting part of things that you added - make sure that all elements in the list are properly sorted is trivial, but improves coverage of the codebase.

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -201,8 +201,7 @@ def _build_sort_direction(self, sort_by: SortBy) -> str:
             case SortBy.NEWEST:
                 sort_direction = "desc"
             case SortBy.CLOSING_SOONEST:
-                # `desc` feel unintuitive, but really if we use `asc`, we are getting markets closing in 2030, 2026, etc.
-                sort_direction = "desc"
+                sort_direction = "asc"
```
kongzii, line 204: wdyt about `self._build_sort_direction` returning both the sorting field and sorting direction? something like `return self.trades_subgraph.FixedProductMarketMaker.openingTimestamp, "asc"`, it should be easier and less error-prone then to extend in the future

prediction_market_agent_tooling/markets/omen/data_models.py
```diff
@@ -112,6 +112,9 @@ class OmenMarket(BaseModel):
     creationTimestamp: t.Optional[int] = None
     condition: Condition
     question: Question
+    openingTimestamp: t.Optional[
```
evangriffiths, line 115: note: for some reason `question.openingTimestamp` returns None for all fpmms, whereas this returns a value as expected (even though they should be the same!), so using this field

tests/markets/omen/test_omen.py
evangriffiths: Good point, added

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -176,6 +172,7 @@ def _build_where_statements(
         if resolved is not None:
             if resolved:
                 where_stms["resolutionTimestamp_not"] = None
+                where_stms["currentAnswer_not"] = self.INVALID_ANSWER
```
evangriffiths, line 175: Existing bugfix - resolved markets should not have invalid answers

kongzii: Actually, resolved market can have invalid answer:

Screenshot by Dropbox Capture

Market can be resolved as INVALID if we can't determine the correct answer.

evangriffiths: Do we want to have these markets returned in our API though? If so, the test prediction-market-agent-tooling/tests/markets/test_omen_subgraph_handler.py::test_filter_resolved_markets needs updating instead.

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
evangriffiths: Agreed, fixed

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
kongzii: I thought it's correct to return them as they are resolved markets, but @gabrielfior had a similar comment when you were on vacation.

So, if you both think it's better just to filter them out, let's keep it.

kongzii (approved):

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
evangriffiths: Okay. I guess from what I can see we don't have a use-case for markets resolved with an invalid answer. If we do, then we can change.

Issue #164, opened 2024-04-05
Refactor `_build_where_statements` arguments


kongzii:
Our logic in `_build_where_statements` is getting messy, we have granual, omen-specific filters, but we also accept `FilterBy` (which is a common enum used across all PM platforms).

Because of that, we need to flip the granual filters in favor of `FilterBy`:

![Uploading Screenshot by Dropbox Capture.png…]()

I think what we can do instead, is keep only omen-specific filters in this function (so remove `FilterBy` argument) and then, have a wrapper function around `get_omen_binary_markets` that would accept only `FilterBy` and it would flip the granual filters as needed. This wrapper function would be used at places, where we support all PM markets.

Note: in https://github.com/gnosis/prediction-market-agent-tooling/pull/165 in `OmenMarketAgent.get_binary_markets` we set `liquidity_bigger_than` based on `filter_by`. For this issue, do this for the remaining args `created_after`, `sort_by_field`, `sort_direction`, etc...

kongzii:
Will be closed with https://github.com/gnosis/prediction-market-agent-tooling/pull/168

PR #165, opened 2024-04-05
Filter open omen markets by non-zero liquidity

evangriffiths:
Now `OmenAgentMarket.get_binary_markets(filter_by=FilterBy.OPEN, sort_by=SortBy.CLOSING_SOONEST)` only returns markets that have not had their liquidity removed by the market-maker

## Summary by CodeRabbit

- **Refactor**
	- Updated terminology from `liquidityMeasure` to `liquidityParameter` across various components for clarity.
- **New Features**
	- Enhanced filtering in `get_binary_markets` to dynamically set liquidity parameters based on user input.
- **Tests**
	- Adjusted test cases to reflect the updated parameter naming convention.

### DIFF
--- a/prediction_market_agent_tooling/markets/omen/data_models.py
+++ b/prediction_market_agent_tooling/markets/omen/data_models.py
@@ -99,7 +99,12 @@ class OmenMarket(BaseModel):
     creator: HexAddress
     category: str
     collateralVolume: Wei
-    liquidityMeasure: Wei
+    # Note: there are two similar parameters relating to liquidity:
+    # liquidityParameter and liquidityMeasure. The former appears to match most
+    # closely with the liquidity returned when calling the contract directly
+    # (see OmenAgentMarket.get_liquidity). So we can use it e.g. for filtering
+    # markets, but until better understood, please call the contract directly.
+    liquidityParameter: Wei
     usdVolume: USD
     collateralToken: HexAddress
     outcomes: list[str]
--- a/prediction_market_agent_tooling/markets/omen/omen.py
+++ b/prediction_market_agent_tooling/markets/omen/omen.py
@@ -185,7 +185,7 @@ class OmenAgentMarket(AgentMarket):
         if filter_by == FilterBy.OPEN:
             # We assume here that we are only interested in markets that
             # we can trade on, i.e. ones with non-zero liquidity.
-            liquidity_bigger_than = (wei_type(0),)
+            liquidity_bigger_than = wei_type(0)
         else:
             liquidity_bigger_than = None
 
--- a/prediction_market_agent_tooling/markets/omen/omen.py
+++ b/prediction_market_agent_tooling/markets/omen/omen.py
@@ -182,6 +182,13 @@ class OmenAgentMarket(AgentMarket):
         created_after: t.Optional[datetime] = None,
         excluded_questions: set[str] | None = None,
     ) -> list[AgentMarket]:
+        if filter_by == FilterBy.OPEN:
+            # We assume here that we are only interested in markets that
+            # we can trade on, i.e. ones with non-zero liquidity.
+            liquidity_bigger_than = (wei_type(0),)
+        else:
+            liquidity_bigger_than = None
+
         return [
             OmenAgentMarket.from_data_model(m)
             for m in get_omen_binary_markets(
@@ -190,6 +197,7 @@ class OmenAgentMarket(AgentMarket):
                 created_after=created_after,
                 filter_by=filter_by,
                 excluded_questions=excluded_questions,
+                liquidity_bigger_than=liquidity_bigger_than,
             )
         ]
 
--- a/prediction_market_agent_tooling/markets/omen/omen_replicate.py
+++ b/prediction_market_agent_tooling/markets/omen/omen_replicate.py
@@ -178,11 +178,11 @@ def omen_unfund_replicated_known_markets_tx(
             )
         ):
             logger.info(
-                f"[{idx+1}/{len(markets)}] Skipping unfunding of `{market.liquidityMeasure=} {market.question=}  {market.url=}`, because it's not saturated yet, `{market.p_yes=}`."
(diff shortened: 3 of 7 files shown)


prediction_market_agent_tooling/markets/omen/data_models.py
```diff
@@ -99,7 +99,12 @@ class OmenMarket(BaseModel):
     creator: HexAddress
     category: str
     collateralVolume: Wei
-    liquidityMeasure: Wei
+    # Note: there are two similar parameters relating to liquidity:
+    # liquidityParameter and liquidityMeasure. The former appears to match most
+    # closely with the liquidity returned when calling the contract directly
+    # (see OmenAgentMarket.get_liquidity). So we can use it e.g. for filtering
+    # markets, but until better understood, please call the contract directly.
+    liquidityParameter: Wei
```
evangriffiths, line 107: This parameter appears to be only used in print statements - so we could actually comment this out (but leave in the comment for future reference) to avoid confusion

kongzii: I think it's fine to leave it in the model, it's called `liquidityParameter` so it doesn't seem like it should be **the liquidity**  and there is a nice comment for it at least.

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -182,6 +182,13 @@ def get_binary_markets(
         created_after: t.Optional[datetime] = None,
         excluded_questions: set[str] | None = None,
     ) -> list[AgentMarket]:
+        if filter_by == FilterBy.OPEN:
+            # We assume here that we are only interested in markets that
+            # we can trade on, i.e. ones with non-zero liquidity.
+            liquidity_bigger_than = wei_type(0)
```
kongzii, line 188: So if we call `get_binary_markets` on the agent class, it will return only bettable ones, but `get_omen_binary_markets` itself will return also non-bettable (but open).

Seems a little confusing to me why `FilterBy.OPEN` would behave differently on different places, but if it's only me then let's keep it.

evangriffiths: I don't think it does behave differently - it's just that `liquidity_bigger_than` isn't an arg of `get_binary_markets` but it is (one we want to set) for `get_omen_binary_markets`.

One option could be to make it also an arg of `get_binary_markets` (though that gets a bit messy given the different currencies of the different market types).

Or to make it less confusing `liquidity_bigger_than` could just be a required arg of `get_omen_binary_markets`, so then the user has to explicitly set it and will be less likely to leave it as None and then be confused when there are 0-liquidity markets returned

kongzii: What I meant is that FilterBy is already used in the graph querying here: https://github.com/gnosis/prediction-market-agent-tooling/blob/43ba22c8255869e0ef46f96ce67a705b753a3588/prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py#L163

So we could have there

```
elif filter_by == FilterBy.OPEN:
            # We can not use `resolved=False` + `finalized=False` here,
            # because even closed markets don't need to be resolved yet (e.g. if someone forgot to finalize the question on reality).
            opened_after = utcnow() + timedelta(seconds=42)
            liquidity_bigger_than = wei_type(0)
```

and then, behaviour of `OmenAgentMarket.get_binary_markets(filter_by=FilterBy.OPEN)` would be the same as `get_omen_binary_markets(filter_by=FilterBy.OPEN)`, because right now, these two return different results (first one will return only liquidity>0 markets, but the second one will return all of them, yet they are using the same filter parameter)

evangriffiths: > So we could have there ...
I didn't want to do that because it means overriding `liquidity_bigger_than`, which seems prone to causing more confusion. Do you really want that? Also, what if you actually want that... to filter by open markets, but include zero-liquidity markets in the return type?

> because right now, these two return different results
Yeah I see what you mean. But I think my preferred option is to make `liquidity_bigger_than` a required arg of `get_omen_binary_markets`, so you can't do `get_omen_binary_markets(filter_by=FilterBy.OPEN)`

evangriffiths: What do you reckon?

kongzii: > which seems prone to causing more confusion.

Yep it totally is, but there is already some overriding in place and an issue for that that we should refactor it: https://github.com/gnosis/prediction-market-agent-tooling/issues/164.

So it would be good to have all the weirdness in one place during the refactoring, to not forget about it.

Or if you still prefer this, let's just add a comment to that issue about this to keep in mind

kongzii (approved):

prediction_market_agent_tooling/markets/omen/omen.py
kongzii: Anyway, either keeping it like this and I can add the comment after rmerge, or if you change it then I simply won't, so LGTM 👍

evangriffiths: > but there is already some overriding in place

Ah yes, you mean `opened_after` is also overridden?

> we should refactor it: https://github.com/gnosis/prediction-market-agent-tooling/issues/164.

You mean remove `filter_by` as an arg from `get_omen_binary_markets`, and move the logic up into `OmenAgentMarket.get_binary_markets` to infer the other args (`finalized`, `created_after`, ...) from `filter_by`? Yes, that makes sense, I like the sound of that too! I guess my current implementation is that, but only for `liquidity_bigger_than`.

evangriffiths: If it's alright, I'll merge this in as-is (so I can re-deploy the known-outcome-agent before the weekend), and you or I can pick up that issue soon?

evangriffiths: Added the comment to https://github.com/gnosis/prediction-market-agent-tooling/issues/164

PR #170, opened 2024-04-08
Make opening timestamp non-optional

kongzii:


prediction_market_agent_tooling/markets/omen/data_models.py
```diff
@@ -118,12 +123,25 @@ class OmenMarket(BaseModel):
     resolutionTimestamp: t.Optional[int] = None
     answerFinalizedTimestamp: t.Optional[int] = None
     currentAnswer: t.Optional[str] = None
-    creationTimestamp: t.Optional[int] = None
+    creationTimestamp: int
```
kongzii, line 126: Typed as non-optional on graph:

Screenshot by Dropbox Capture

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -149,7 +160,9 @@ def _build_where_statements(
             where_stms["creationTimestamp_gt"] = to_int_timestamp(created_after)
 
         if opened_before:
-            where_stms["openingTimestamp_lt"] = to_int_timestamp(opened_before)
+            where_stms["question_"]["openingTimestamp_lt"] = to_int_timestamp(
```
evangriffiths, line 163: How come you've changed the 'where statements'?

kongzii: Btw I also tried to change it in https://github.com/gnosis/prediction-market-agent-tooling/blob/7906ccdb5a91f14bdeefe5ee0f67892ef6d3deab/prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py#L218

But I tried the following syntaxes:

- `self.trades_subgraph.FixedProductMarketMaker.question.openingTimestamp`
- `self.trades_subgraph.FixedProductMarketMaker.question__openingTimestamp`

And it doesn't work, maybe @gabrielfior knows?

gabrielfior: My only 2cents are that `fpmm.openingTimestamp` can be null, whereas `question.openingTimestamp` cannot (determined that querying the subgraph manually).
So I'm fine with this change.

kongzii: @gabrielfior That's cool, but do you know how that syntax I asked about should be written?

PR #171, opened 2024-04-08
Fix PolymarketRewards model

kongzii:


prediction_market_agent_tooling/markets/omen/omen_replicate.py
```diff
@@ -56,7 +56,11 @@ def omen_replicate_from_tx(
         10 if market_type == MarketType.POLYMARKET else 100,
         market_type,
         filter_by=FilterBy.OPEN,
-        sort_by=SortBy.NONE,
+        sort_by=(
+            SortBy.NONE
+            if market_type == MarketType.POLYMARKET
+            else SortBy.CLOSING_SOONEST
```
evangriffiths, line 62: how come?

evangriffiths (approved):

prediction_market_agent_tooling/markets/omen/omen_replicate.py
kongzii: I'm going through the logs and in last run it replicated a market that will end in 9 months. I thought if we replicate only small number of markets, it make sense to think short-term.

Would you keep it at random?

kongzii: (even on random markets that close in more than 365 days won't be replicated)

kongzii: (and even with `CLOSING_SOONEST` there will be only ones with at least 48h of open time)

kongzii: And this replicated market that is already known but still open on Manifold 🙈 I'll revert and check it again in a few days.

evangriffiths: 😄

PR #172, opened 2024-04-08
Added subgraph methods to fetch markets by user position

gabrielfior:
Will be needed by microchain agent on PMA repo.

### DIFF
--- a/prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
+++ b/prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
@@ -505,3 +505,24 @@ class OmenSubgraphHandler(metaclass=SingletonMeta):
         result = self.sg.query_json(fields)
         items = self._parse_items_from_json(result)
         return [RealityAnswer.model_validate(i) for i in items]
+
+    def get_markets_from_all_user_positions(
+        self, user_positions: list[OmenUserPosition]
+    ) -> list[OmenMarket]:
+        condition_ids = sum([u.position.condition_ids for u in user_positions], [])
+        markets = self.get_omen_binary_markets(
+            limit=sys.maxsize, condition_id_in=condition_ids
+        )
+        return markets
+
+    def get_market_from_user_position(
+        self, user_position: OmenUserPosition
+    ) -> OmenMarket:
+        """Markets and user positions are uniquely connected via condition_ids"""
+        condition_ids = user_position.position.condition_ids
+        markets = self.get_omen_binary_markets(limit=1, condition_id_in=condition_ids)
+        if len(markets) != 1:
+            raise ValueError(
+                f"Incorrect number of markets fetched {len(markets)}, expected 1."
+            )
+        return markets[0]
--- a/prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
+++ b/prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
@@ -306,26 +306,46 @@ class OmenSubgraphHandler(metaclass=SingletonMeta):
             **optional_params,
         )
 
+        omen_markets = self.do_markets_query(markets)
+        return omen_markets
+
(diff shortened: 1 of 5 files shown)


prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -329,6 +330,25 @@ def get_omen_market(self, market_id: HexAddress) -> OmenMarket:
 
         return omen_markets[0]
 
+    def get_omen_market_by_title(self, market_question_title: str) -> OmenMarket | None:
```
kongzii, line 333: Wdyt about adding `market_question_title` as filter to `get_omen_binary_markets`?

As you say, it's possible to have multiple markets with the same title, so seems like that method suits it well.

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -505,3 +525,26 @@ def get_answers(self, question_id: HexBytes) -> list[RealityAnswer]:
         result = self.sg.query_json(fields)
         items = self._parse_items_from_json(result)
         return [RealityAnswer.model_validate(i) for i in items]
+
+    def get_markets_from_all_user_positions(
+        self, user_positions: list[OmenUserPosition]
+    ) -> list[OmenMarket]:
+        condition_ids: list[HexBytes] = sum(
```
kongzii, line 532: There can be potentially many duplicates

```suggestion
        condition_ids: list[HexBytes] = list(set(sum(
```

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -505,3 +525,26 @@ def get_answers(self, question_id: HexBytes) -> list[RealityAnswer]:
...
+            [u.position.conditionIds for u in user_positions], []
+        )
```
kongzii, line 534:
```suggestion
        )))
```

(to make the suggestion above work)

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
evangriffiths: An alternative suggestion: we could remove the need for this by having `GetMarkets` return the market id, and then we can use `subgraph_handler.get_omen_market_by_market_id` in the microchain agent's fuction directly

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
gabrielfior: Great catch, yes for sure duplicates should be expected. Implemented.

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
gabrielfior: I will go with @evangriffiths suggestion and reduce the amount of code necessary here. If we don't need to query by `market_title`, then no need for this method or added complexity.

tests/markets/omen/test_omen_subgraph_handler.py
```diff
@@ -189,3 +193,58 @@ def test_get_user_positions_with_position_ids(
...
+        "0xffe4bf3e61be010728813a2a61ef422fd7d07b410170b64a5dfced9549f2e057"
+    )
+    positions = omen_subgraph_handler.get_positions(condition_id)
+    assert len(positions) == 2
+
+
+def build_incomplete_user_position_from_condition_ids(
+    condition_ids: list[HexBytes],
+) -> OmenUserPosition:
+    return OmenUserPosition.construct(
```
kongzii, line 246: Interesting one! Didn't know about it.

kongzii (approved):

tests/markets/omen/test_omen_subgraph_handler.py
gabrielfior: It's a hacky way to bypass the model validation - use with care!

PR #179, opened 2024-04-10
Add AgentMarket.get_token_balance, with support for Omen

evangriffiths:
Carrying on with supporting a 1-to-1 mapping between AgentMarket functions and functions required by a 'general agent' (e.g. microchain agent in PMA).

Support for manifold and polymarket to come later

### DIFF
--- a/prediction_market_agent_tooling/markets/omen/omen.py
+++ b/prediction_market_agent_tooling/markets/omen/omen.py
@@ -22,7 +22,11 @@ from prediction_market_agent_tooling.markets.agent_market import (
     FilterBy,
     SortBy,
 )
-from prediction_market_agent_tooling.markets.data_models import BetAmount, Currency
+from prediction_market_agent_tooling.markets.data_models import (
+    BetAmount,
+    Currency,
+    TokenAmount,
+)
 from prediction_market_agent_tooling.markets.omen.data_models import (
     OMEN_BASE_URL,
     OMEN_FALSE_OUTCOME,
@@ -209,12 +213,15 @@ class OmenAgentMarket(AgentMarket):
     def get_index_set(self, outcome: str) -> int:
         return self.get_outcome_index(outcome) + 1
 
-    def get_token_balance(self, user_id: str, outcome: str) -> Decimal:
+    def get_token_balance(self, user_id: str, outcome: str) -> TokenAmount:
         index_set = self.get_index_set(outcome)
         balances = get_conditional_tokens_balance_for_market(
             self, Web3.to_checksum_address(user_id)
         )
-        return Decimal(balances[index_set])
+        return TokenAmount(
+            amount=Decimal(wei_to_xdai(balances[index_set])),
+            currency=Currency.xDai,
+        )
 
 
 def pick_binary_market(
--- a/prediction_market_agent_tooling/markets/omen/omen.py
+++ b/prediction_market_agent_tooling/markets/omen/omen.py
@@ -45,6 +45,7 @@ from prediction_market_agent_tooling.markets.omen.omen_subgraph_handler import (
     OmenSubgraphHandler,
 )
 from prediction_market_agent_tooling.tools.balances import get_balances
+from prediction_market_agent_tooling.tools.contract import wait_until_nonce_changed
 from prediction_market_agent_tooling.tools.utils import check_not_none
 from prediction_market_agent_tooling.tools.web3_utils import (
     add_fraction,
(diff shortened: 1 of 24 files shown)


prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -206,6 +206,23 @@ def get_contract(self) -> OmenFixedProductMarketMakerContract:
...
+        # Find position matching market_outcome
+        index_set = self.get_index_set(outcome)
+        position_for_index_set = next(p for p in positions if index_set in p.indexSets)
+        position_as_int = int(position_for_index_set.id.hex(), 16)
+
+        balance = OmenConditionalTokenContract().balanceOf(
+            from_address=Web3.to_checksum_address(user_id),
+            position_id=position_as_int,
+        )
+        return Decimal(balance)
```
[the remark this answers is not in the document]
gabrielfior: Makes sense coderabbitai! We should check for cases where outcome is not contained in `market.outcomes` and return a 0 balance.

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -206,6 +206,23 @@ def get_contract(self) -> OmenFixedProductMarketMakerContract:
...
+    def get_index_set(self, outcome: str) -> int:
+        return self.get_outcome_index(outcome) + 1
+
+    def get_token_balance(self, user_id: str, outcome: str) -> Decimal:
+        positions = OmenSubgraphHandler().get_positions(self.condition.id)
+
+        # Find position matching market_outcome
+        index_set = self.get_index_set(outcome)
+        position_for_index_set = next(p for p in positions if index_set in p.indexSets)
+        position_as_int = int(position_for_index_set.id.hex(), 16)
```
kongzii, line 218: This seems like a common thing we are doing, wdyt about adding method `as_int` to https://github.com/gnosis/prediction-market-agent-tooling/blob/b081d0a3d764e22c5892d952ec8cf7f862f8f38f/prediction_market_agent_tooling/tools/hexbytes_custom.py#L42?

kongzii: Btw there is this function from @gabrielfior  already: https://github.com/gnosis/prediction-market-agent-tooling/blob/877f281b60084c439ef0c96b009317f4a2b209b0/prediction_market_agent_tooling/markets/omen/omen.py#L537

Seems like something like this should work:

```python
index_set = self.get_index_set(outcome)
balances = get_conditional_tokens_balance_for_market(self, Web3.to_checksum_address(user_id))
return balances[index_set]
```

prediction_market_agent_tooling/markets/omen/omen.py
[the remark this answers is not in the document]
kongzii: > Makes sense. We should check for cases where outcome is not contained in market.outcomes and return a 0 balance.

Returning 0 should mean that the outcome does exist, but the user doesn't have any tokens for it; asking for an outcome that's not in the market feels like a bug.

So I'm good with Evan's current implementation (except for the suggestion in another comment to reuse the existing function). That's doing

```
def get_index_set(self, outcome: str) -> int:
        return self.get_outcome_index(outcome) + 1

def get_outcome_index(self, outcome: str) -> int:
        try:
            return self.outcomes.index(outcome)
        except ValueError:
            raise ValueError(f"Outcome `{outcome}` not found in `{self.outcomes}`.")
```

and so will throw a ValueError if wrong outcome is passed in.

prediction_market_agent_tooling/markets/omen/omen.py
evangriffiths: 🤩 Oo that's handy. Great suggestion!

evangriffiths: > This seems like a common thing we are doing, wdyt about adding method `as_int`

Also a good idea. Will leave to the next person to add another case of it 😆

kongzii: We can just do this, look:

@coderabbitai Please create an issue with the name "Add method `as_int` to the HexBytes class", thanks!

evangriffiths: oo cool feature

evangriffiths: would be even better if it linked to this conversation in the issue description

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -206,6 +206,16 @@ def get_contract(self) -> OmenFixedProductMarketMakerContract:
...
 
+    def get_index_set(self, outcome: str) -> int:
+        return self.get_outcome_index(outcome) + 1
+
+    def get_token_balance(self, user_id: str, outcome: str) -> Decimal:
+        index_set = self.get_index_set(outcome)
+        balances = get_conditional_tokens_balance_for_market(
+            self, Web3.to_checksum_address(user_id)
+        )
+        return Decimal(balances[index_set])
```
kongzii, line 217: Just double check if agent works with Weis and doesn't expect this to be in xDai?

kongzii (approved):

prediction_market_agent_tooling/markets/omen/omen.py
evangriffiths: Hmm, having a bit of a confused moment here. What does a token balance refer to here. Just assume we convert the value to xDai for now. Is it

- "The amount of xDai I bet on that outcome for that market" or
- "The current value of my tokens on that outcome if that market were to resolve in favour of that outcome"
- or something else

?

i.e. if I bet 1 xDai on Yes on a market, then call `get_token_balance` on that market, will it return `1 xDai` or will it depend on the current p_yes of the market?

kongzii: I think it should be amount of shares that we have. So if we buy 26.79 shares for 10 xDai, get_token_balance should always be 26.79 shares, which may become 26.79 xDai if we were correct.

> The current value

Does the value of the shares that were already bought change over time? 😱 

Screenshot by Dropbox Capture

evangriffiths: Wonder if `get_token_balance`'s return type should be `BetAmount` in that case - i.e. have amount and currency

evangriffiths: Or a `TokenAmount`, and then BetAmount and ProfitAmount 
https://github.com/gnosis/prediction-market-agent-tooling/blob/main/prediction_market_agent_tooling/markets/data_models.py#L21

can just be aliases

kongzii: > Or a `TokenAmount`, and then BetAmount and ProfitAmount https://github.com/gnosis/prediction-market-agent-tooling/blob/main/prediction_market_agent_tooling/markets/data_models.py#L21
> 
> can just be aliases

Yeah that looks good

PR #183, opened 2024-04-11
Expose token-selling as method on AgentMarket

evangriffiths:
Only supporting Omen in this PR. Sorry for the tech debt!

### DIFF
--- a/prediction_market_agent_tooling/markets/omen/omen.py
+++ b/prediction_market_agent_tooling/markets/omen/omen.py
@@ -45,6 +45,7 @@ from prediction_market_agent_tooling.markets.omen.omen_subgraph_handler import (
     OmenSubgraphHandler,
 )
 from prediction_market_agent_tooling.tools.balances import get_balances
+from prediction_market_agent_tooling.tools.contract import wait_until_nonce_changed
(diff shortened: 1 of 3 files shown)


prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -92,13 +93,27 @@ def place_bet(
...
         amount_xdai = xDai(amount.amount)
         keys = APIKeys()
-        binary_omen_buy_outcome_tx(
-            amount=amount_xdai,
-            from_private_key=keys.bet_from_private_key,
-            market=self,
-            binary_outcome=outcome,
-            auto_deposit=omen_auto_deposit,
-        )
+        with wait_until_nonce_changed(keys.bet_from_address):
```
evangriffiths, line 96: `wait_until_nonce_changed` seems like a safe thing to have on by default

kongzii: Agreed, if you think it's useful everywhere, it can be done directly around https://github.com/gnosis/prediction-market-agent-tooling/blob/4e9ec041e06d5da2c32d031ac9bd588d9fe50d28/prediction_market_agent_tooling/tools/contract.py#L99

PR #188, opened 2024-04-11
Add `omen_sell_outcome_tx` docstring clarification

evangriffiths:


evangriffiths:
@kongzii 
https://github.com/gnosis/prediction-market-agent-tooling/pull/179#discussion_r1561054006

> Does the value of the shares that we already bought change over time? 😱

The amount you win if the market resolves in the direction you bet in, then no. But in the context of selling the shares, yes it does!

PR #193, opened 2024-04-15
Fix test_balance_for_user_in_market

evangriffiths:
Test is flaky, as it will start failing as the market is resolved. Asserting balance==0 still gives full test coverage.

## Summary by CodeRabbit

- **Tests**
	- Updated the balance check in market tests to ensure accurate comparison.

### DIFF
--- a/tests/markets/omen/test_omen.py
+++ b/tests/markets/omen/test_omen.py
@@ -217,7 +217,7 @@ def test_balance_for_user_in_market() -> None:
         outcome=OMEN_TRUE_OUTCOME,
     )
     assert balance_yes.currency == Currency.xDai
-    assert float(balance_yes.amount) == 0.001959903969410997
+    assert float(balance_yes.amount) == 0
 
     balance_no = market.get_token_balance(
         user_id=user_address,


tests/markets/omen/test_omen.py
```diff
@@ -217,7 +217,7 @@ def test_balance_for_user_in_market() -> None:
         outcome=OMEN_TRUE_OUTCOME,
     )
     assert balance_yes.currency == Currency.xDai
-    assert float(balance_yes.amount) == 0.001959903969410997
+    assert float(balance_yes.amount) == 0
```
kongzii, line 220: how come this will work or that it didn't fail before? :o

kongzii: oh you sold it on the market?

kongzii (approved):

tests/markets/omen/test_omen.py
evangriffiths: Yeah. It was only recently resolved https://aiomen.eth.limo/#/0x59975b067b0716fef6f561e1e30e44f606b08803/finalize

gabrielfior: My 2 cents
-> Test (written by me) was not awesome, as the balance will definitely change in the future (either via selling or via redeeming positions). So my bad.
-> Still, not checking the balance deviates from the purpose of the test - so the balance in the test (0.001959903969410997) was correct for a block in the past (e.g. last week, block 33247199). So the "optimal" way to make the test pass would be
    -> Add an argument block_number to `market.get_token_balance` 
    -> Pass this argument all the way to the contract call, like this `myContract.myMethod.call(block=123, "), all the way here
    https://github.com/gnosis/prediction-market-agent-tooling/blob/ded4fac0fd3816b73e3914c3b604f1085d8d519f/prediction_market_agent_tooling/tools/web3_utils.py#L104

PR #196, opened 2024-04-16
Fix mypy on agent market

kongzii:


prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -75,6 +75,8 @@ class OmenAgentMarket(AgentMarket):
     market_maker_contract_address_checksummed: ChecksumAddress
     condition: Condition
     finalized_time: datetime | None
+    created_time: datetime
+    close_time: datetime
```
kongzii, line 79: On Omen we know that these two times can't be None, it's okay to narrow the type down

evangriffiths (approved):

Issue #148, opened 2024-04-02
Add support for Safe in transactions


kongzii:
Having funds in Safe is safer as we can configure things like maximum withdraw amount per day, recovery, etc. In the future we want our agents to use Safe to store their funds instead of plain private key.

gabrielfior:
Like this approach - I would also note that Olas uses Safe under-the-hood for all agents, so I would assume we can either a. copy-paste everything or, even better, b. use their interface for creating Safe accounts automatically for agents.

gabrielfior:
I think we can definitely reuse something over start from scratch.
In my head, following options (in order of preference) make sense:

1 - Add Open-Autonomy as dependency to PMAT (https://docs.autonolas.network/open-autonomy/api/contracts/gnosis_safe_proxy_factory/contract/) and reuse some of their SafeContracts classes, so that we are able to deploy a Safe (via Safe factory) and send txs to the safe
2 - If above not possible, we can try to re-use old Python scripts to interact with Safe (https://github.com/safe-global/safe-eth-py/tree/69f1ee3263086403f6017effa0841c6a2fbba6d6/gnosis/safe). This is a bit risky (could work only for old versions), but we could test it out. I'm asking the Safe team about option 2 currently.
3 - If none of the above works, we can have a small Typescript server running somewhere, which accepts requests and wraps the Safe functionality.

gabrielfior:
Found an interesting implementation using `safe-eth-py` (https://github.com/safe-global/safe-eth-py/blob/main/gnosis/safe/tests/safe_test_case.py#L225), which is also used by Polywrap's AutoTx (https://github.com/polywrap/AutoTx/blob/main/autotx/utils/ethereum/SafeManager.py). I'll probably follow a similar path.

kongzii:
Do you know if it will be possible to somehow reuse our current code? E.g. all the functions accepting `private key` to also support Safe?

gabrielfior:
Closed by #200

PR #200, opened 2024-04-16
Added Safe support for agents

gabrielfior:
- Added a script for creating a Safe from PMAT for a given agent
- Extended README with instructions on how to create a SAFE for an agent
- Extended `prediction_market_agent_tooling/tools/contract.py` for cases where a Safe is available, so that transactions are executed via Safe


prediction_market_agent_tooling/tools/contract.py
```diff
@@ -53,19 +57,37 @@ def wait_until_nonce_changed(
...
 
     CHAIN_ID: t.ClassVar[ChainID]
     CHAIN_RPC_URL: t.ClassVar[str]
-
-    abi: ABI
-    address: ChecksumAddress
+    signer_private_key: PrivateKey | None = (
+        None  # no private_key required for call methods (reading)
+    )
+    safe_address: ChecksumAddress | None = None  # no Safe address required, only signer_private_key suffices if not using Safe.
```
gabrielfior, line 73: 2 fields added as follows
-> safe_address: can be None, this will normally be loaded from SAFE_ADDRESS env and passed to a Contract instance (via the method below, `build_with_private_key_and_safe`)
-> signer_private_key: can also be None (in case one just wants to call `self.call` methods, i.e. methods that read from the blockchain but don't write)

prediction_market_agent_tooling/tools/contract.py
```diff
@@ -53,19 +57,37 @@ def wait_until_nonce_changed(
...
+    abi: ABI  # = Field(default=ABI(""))
+    address: ChecksumAddress  # = Field(default=ChecksumAddress(HexAddress(HexStr(""))))
 
     _abi_field_validator = field_validator("abi", mode="before")(abi_field_validator)
 
+    @classmethod
+    def build_with_private_key_and_safe(
```
gabrielfior, line 80: This class should be used as a convenience function, whereas the normal __init__ is used for the `read-only` mode.

prediction_market_agent_tooling/tools/contract.py
```diff
@@ -96,22 +117,37 @@ def send(
         """
         Used for changing a state (writing) to the contract.
         """
-        with wait_until_nonce_changed(private_key_to_public_key(from_private_key)):
-            receipt = send_function_on_contract_tx(
+
+        if not self.signer_private_key:
+            raise ValueError(
+                "Signer private key must be set for send functions to be called."
+            )
```
gabrielfior, line 124: This avoids contract classes that were instantiated via `__init__` from calling `send` functions.

scripts/bet_omen.py
```diff
@@ -18,6 +19,7 @@
 def buy(
     amount: str = typer.Option(),
     from_private_key: str = typer.Option(),
+    safe_address: str = typer.Option(default=None),
```
gabrielfior, line 22: Default case for scripts is None for safe_addresses and txs will be completed directly using the private_key as usual.

README.md
```diff
@@ -65,6 +65,14 @@ class DeployableCoinFlipAgent(DeployableAgent):
 DeployableCoinFlipAgent().deploy_gcp(...)
 ```
 
+For deploying a Safe manually for a given agent, run the script below:
+
+```commandline
+poetry run python scripts/create_safe_for_agent.py  --from-private-key <YOUR_AGENT_PRIVATE_KEY> --salt-nonce 42
+```
+This will output the newly created Safe in the terminal, and it can then be copied over to the deployment part (e.g. Terraform).
+Note that `salt_nonce` can be passed so that the created safe is deterministically created for each agent, so that, if the same `salt_nonce` is used, the script will not create a new Safe for the agent, instead it will output the previously existent Safe.
```
kongzii, line 74: nice

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -69,6 +69,7 @@ class OmenAgentMarket(AgentMarket):
     currency: t.ClassVar[Currency] = Currency.xDai
     base_url: t.ClassVar[str] = OMEN_BASE_URL
     creator: HexAddress
+    use_safe: bool = False
```
kongzii, line 72: unused

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -218,9 +225,15 @@ def get_binary_market(id: str) -> "OmenAgentMarket":
             )
         )
 
-    def get_contract(self) -> OmenFixedProductMarketMakerContract:
+    def get_contract(
+        self,
+        signer_private_key: PrivateKey | None = None,
+        safe_address: ChecksumAddress | None = None,
```
kongzii, line 231: This method is returning contract of the market, why do we need to pass private key and safe address here?

```
def get_contract(self) -> OmenFixedProductMarketMakerContract:
        return OmenFixedProductMarketMakerContract(
            address=self.market_maker_contract_address_checksummed
        )
```

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -259,8 +273,15 @@ def omen_buy_outcome_tx(
...
 
-    market_contract: OmenFixedProductMarketMakerContract = market.get_contract()
-    collateral_token_contract = OmenCollateralTokenContract()
+    market_contract: OmenFixedProductMarketMakerContract = market.get_contract(
+        from_private_key, safe_address
+    )
+
+    collateral_token_contract = (
+        OmenCollateralTokenContract.build_with_private_key_and_safe(
+            from_private_key, safe_address
```
kongzii, line 282: Aha, I see, so instead of initing the contract as `c = OmenCollateralTokenContract()` and then passing the private key to the transactional methods as `x.do_something(private_key=private_key)`. Now class itself holds the private key + safe address and then just call `x.do_something()`.

Personally I liked it better before because by looking at `get_contract` above, for example, it's not clear to me why someone would call it with a private key and safe address.

But if @evangriffiths likes this new approach more as well, I'm okay with being over-voted 🙈

prediction_market_agent_tooling/markets/omen/omen_contracts.py
```diff
@@ -69,6 +67,10 @@ def resolve(
         )
 
 
+class DummyContract(ContractOnGnosisChain):
```
kongzii, line 70: unused

prediction_market_agent_tooling/tools/contract.py
```diff
@@ -32,6 +33,9 @@ def abi_field_validator(value: str) -> ABI:
         with open(value) as f:
             value = f.read()
 
+    if "" == value:
+        raise ValueError(f"Invalid ABI: {value}")
```
kongzii, line 37: No need for this, JSON parsing below will already fail:

Screenshot by Dropbox Capture

```suggestion
```

prediction_market_agent_tooling/tools/contract.py
```diff
@@ -53,19 +57,37 @@ def wait_until_nonce_changed(
...
     CHAIN_ID: t.ClassVar[ChainID]
     CHAIN_RPC_URL: t.ClassVar[str]
-
-    abi: ABI
-    address: ChecksumAddress
+    signer_private_key: PrivateKey | None = (
+        None  # no private_key required for call methods (reading)
+    )
+    safe_address: ChecksumAddress | None = None  # no Safe address required, only signer_private_key suffices if not using Safe.
+    abi: ABI  # = Field(default=ABI(""))
```
kongzii, line 74: Why these new comments?

```suggestion
    abi: ABI
```

prediction_market_agent_tooling/tools/contract.py
```diff
@@ -53,19 +57,37 @@ def wait_until_nonce_changed(
...
+    address: ChecksumAddress  # = Field(default=ChecksumAddress(HexAddress(HexStr(""))))
```
kongzii, line 75:
```suggestion
    address: ChecksumAddress
```

prediction_market_agent_tooling/tools/safe.py
```diff
@@ -0,0 +1,126 @@
...
+        )
+    elif fallback_handler != NULL_ADDRESS and not ethereum_client.is_contract(
+        fallback_handler
+    ):
+        raise EnvironmentError(
+            f"Fallback handler address {fallback_handler} "
+            f"does not exist on network {ethereum_network.name}"
+        )
+
+    account_balance: int = ethereum_client.get_balance(account.address)
```
kongzii, line 65:
```suggestion
    account_balance: Wei = ethereum_client.get_balance(account.address)
```

prediction_market_agent_tooling/tools/safe.py
```diff
@@ -0,0 +1,126 @@
...
+    if not account_balance:
+        logger.info(
+            "Client does not have any funds. Let's try anyway in case it's a network without gas costs"
+        )
+    else:
+        ether_account_balance = round(
```
kongzii, line 71: This variable is sometimes defined and sometimes isn't, but it seems like it can be created before the if-else statements

kongzii: Btw we have `wei_to_xdai` function in web3_utils.py file already

prediction_market_agent_tooling/tools/safe.py
```diff
@@ -0,0 +1,126 @@
...
+    if not ethereum_client.w3.eth.get_code(
+        safe_contract_address
+    ) or not ethereum_client.w3.eth.get_code(proxy_factory_address):
+        raise EnvironmentError("Network not supported")
+
+    logger.info(
+        f"Creating new Safe with owners={owners} threshold={threshold} salt-nonce={salt_nonce}"
+    )
+
+    # We ignore mypy below because using the proper class SafeV141 yields an error.
```
kongzii, line 88:
```suggestion
    # We ignore mypy below because using the proper class SafeV141 yields an error and mypy doesn't understand that there is a hacky `__new__` method on this abstract class.
```

Just for clarity for the future readers

prediction_market_agent_tooling/tools/web3_utils.py
```diff
@@ -152,3 +187,58 @@ def send_function_on_contract_tx(
...
+    contract_address: ChecksumAddress,
+    contract_abi: ABI,
+    from_private_key: PrivateKey,
+    safe_address: ChecksumAddress,
+    function_name: str,
+    function_params: Optional[list[Any] | dict[str, Any]] = None,
+    tx_params: Optional[TxParams] = None,
+    timeout: int = 180,
+) -> TxReceipt:
+    if not safe_address:
```
kongzii, line 215: It's typed as `safe_address: ChecksumAddress`, so this check isn't required and can be deleted

```suggestion
```

prediction_market_agent_tooling/tools/web3_utils.py
```diff
@@ -152,3 +187,58 @@ def send_function_on_contract_tx(
...
+        function_params=function_params,
+        tx_params=tx_params,
+    )
+
+    if not web3.HTTPProvider.endpoint_uri:
+        raise EnvironmentError(f"RPC_URL not available in web3 object.")
+    ethereum_client = EthereumClient(
+        ethereum_node_url=URI(web3.HTTPProvider.endpoint_uri)
+    )
+    s = SafeV141(safe_address, ethereum_client)
```
kongzii, line 233: Here using `SafeV141` directly works? Can't it be used above as well?

scripts/create_safe_for_agent.py
```diff
@@ -0,0 +1,42 @@
...
+from gnosis.eth import EthereumClient
+
+from prediction_market_agent_tooling.tools.gnosis_rpc import GNOSIS_RPC_URL
+from prediction_market_agent_tooling.tools.safe import create_safe
+
+
+def create_safe_for_agent(
+    from_private_key: str = typer.Option(),
+    rpc_url: str | None = None,
+    salt_nonce: int = secrets.randbits(256),
```
kongzii, line 15: Is there any reason not to default it to 42 as in the README?

```suggestion
    salt_nonce: int = 42,
```

kongzii: Oh, nevermind, probably because you didn't want to always create the same safe by default.

In that case, beware that `salt_nonce: int = secrets.randbits(256)` is initialised only once, and then it remains the same throughout the whole process. So, if someone decides to use this programmatically for some reason, he will unintentionally get the same safe every time.

Doing `salt_nonce: int | None = None,` and then `salt_nonce = salt_nonce or secrets.randbits(256)` is usually safer.

prediction_market_agent_tooling/tools/contract.py
```diff
@@ -53,19 +57,37 @@ def wait_until_nonce_changed(
...
 
+    @classmethod
+    def build_with_private_key_and_safe(
+        cls: Type[T],
+        signer_private_key: PrivateKey,
+        safe_address: ChecksumAddress | None,
+    ) -> T:
+        # Note that abi and signer are to be defined in the subclasses. If not, pydantic.ValidationError will be thrown.
+        return cls(
+            safe_address=safe_address,  # type: ignore
```
kongzii, line 87: Why is this type ignore necessary? It shouldn't be

prediction_market_agent_tooling/tools/contract.py
gabrielfior: Yep, I had removed this. Good catch, thanks!

prediction_market_agent_tooling/tools/safe.py
gabrielfior: Agree, implemented.

prediction_market_agent_tooling/markets/omen/omen.py
evangriffiths: Yes I prefer the previous approach. I feel like we need a new class for a 'private key that may or may not come with a safe address'

e.g.
```
class PrivateCredentials(BaseModel):
    private_key: PrivateKey
    safe_address: SafeAddress | None

    def has_safe_address(self) -> bool:
        return self.safe_address is not None
    
    def from_api_keys(self, api_keys: ApiKeys) -> "PrivateCredentials":
        return PrivateCredentials(
            private_key=api_keys.private_key.get_secret_value(),
            safe_address=api_keys.safe_address,
        )
```

Then everywhere where we currently pass around a private key, we pass around a `private_credentials` object. And in `ContractBaseClass.send` it does what you've already done - makes one set of calls if `private_credentials.has_safe_address`, otherwise, makes the existing call.

evangriffiths: Then I feel like this task could be split into 2 PRs:
1. replace all uses of `private_key` with `private_credentials`, where `PrivateCredentials` is just a wrapper around `PrivateKey` (i.e. no safe at this point)
2. Add `safe_address` to `PrivateCredentials`, and update `ContractBaseClass.send`

prediction_market_agent_tooling/tools/contract.py
gabrielfior: Function will not be needed anymore, closing this

prediction_market_agent_tooling/markets/omen/omen.py
gabrielfior: Implemented PrivateCredentials as per above

prediction_market_agent_tooling/tools/web3_utils.py
gabrielfior: Good catch! I switched back to `Safe` and wrote a test for it.

prediction_market_agent_tooling/markets/omen/omen.py
gabrielfior: Reverted back to version without keys

gabrielfior:
Closes #148

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -247,72 +262,84 @@ def pick_binary_market(
...
     outcome: str,
     auto_deposit: bool,
+    web3: Web3 | None = None,
 ) -> None:
     """
     Bets the given amount of xDai for the given outcome in the given market.
     """
     amount_wei = xdai_to_wei(amount)
-    from_address_checksummed = private_key_to_public_key(from_private_key)
+    if private_credentials.safe_address is not None:
```
kongzii, line 276: Feels like this will be a common pattern, is that right?

If so, this if-elsing could be a @property on PrivateCredentials class

prediction_market_agent_tooling/tools/contract.py
```diff
@@ -53,14 +54,16 @@ def wait_until_nonce_changed(
         time.sleep(sleep_time)
 
 
+T = TypeVar("T", bound="ContractBaseClass")
```
kongzii, line 57: Is this still needed?

```suggestion
```

prediction_market_agent_tooling/tools/contract.py
```diff
@@ -53,14 +54,16 @@ def wait_until_nonce_changed(
...
+
 class ContractBaseClass(BaseModel):
     """
     Base class holding the basic requirements and tools used for every contract.
     """
 
     CHAIN_ID: t.ClassVar[ChainID]
     CHAIN_RPC_URL: t.ClassVar[str]
-
     abi: ABI
```
kongzii, line 67:
```suggestion

    abi: ABI
```

prediction_market_agent_tooling/tools/web3_utils.py
```diff
@@ -118,7 +157,6 @@ def call_function_on_contract(
 )
 def send_function_on_contract_tx(
     web3: Web3,
-    *,
```
kongzii, line 121: You don't like starts? 🙈

prediction_market_agent_tooling/tools/web3_utils.py
```diff
@@ -152,3 +187,53 @@ def send_function_on_contract_tx(
...
+@tenacity.retry(
+    # Retry only for the transaction errors that match the given patterns,
+    # add other retrieable errors gradually to be safe.
+    retry=tenacity.retry_if_exception_message(
+        match="(.*wrong transaction nonce.*)|(.*Invalid.*)|(.*OldNonce.*)"
+    ),
+    wait=tenacity.wait_chain(*[tenacity.wait_fixed(n) for n in range(1, 10)]),
+    stop=tenacity.stop_after_attempt(9),
+    after=lambda x: logger.debug(
+        f"send_function_on_contract_tx failed, {x.attempt_number=}."
```
kongzii, line 201:
```suggestion
        f"send_function_on_contract_tx_using_safe failed, {x.attempt_number=}."
```

prediction_market_agent_tooling/tools/web3_utils.py
```diff
@@ -152,3 +187,53 @@ def send_function_on_contract_tx(
...
+        raise EnvironmentError(f"RPC_URL not available in web3 object.")
+    ethereum_client = EthereumClient(ethereum_node_url=URI(web3.provider.endpoint_uri))  # type: ignore
+    s = Safe(safe_address, ethereum_client)  # type: ignore
+    safe_tx = s.build_multisig_tx(
+        to=Web3.to_checksum_address(tx_params["to"]),
+        data=HexBytes(tx_params["data"]),
+        value=tx_params["value"],
+    )
+    safe_tx.sign(from_private_key.get_secret_value())
+    safe_tx.call()  # simulate call
```
kongzii, line 235: @coderabbitai Please create an issue to do this call simulation in `send_function_on_contract_tx` as well.

pyproject.toml
```diff
@@ -36,6 +36,10 @@ langchain-openai = { version = "^0.0.5", optional = true}
 google-api-python-client = { version = "2.95.0", optional = true}
 subgrounds = "^1.8.1"
 loguru = "^0.7.2"
+safe-eth-py = "^6.0.0b14"
+eth-account = "0.8.0"
```
kongzii, line 40:
```suggestion
eth-account = "^0.8.0"
```

Could be problematic in the future

kongzii (commented):
Left just minor comments and questions for my understanding, otherwise LGTM

prediction_market_agent_tooling/tools/contract.py
```diff
@@ -147,51 +161,55 @@ class ContractERC20BaseClass(ContractBaseClass):
...
+            web3=web3,
         )
 
     def deposit(
         self,
+        private_credentials: PrivateCredentials,
         amount_wei: Wei,
-        from_private_key: PrivateKey,
         tx_params: t.Optional[TxParams] = None,
+        web3: Web3 | None = None,
```
evangriffiths, line 186: It doesn't look like this function is called anywhere with the arg?

prediction_market_agent_tooling/markets/omen/omen.py
gabrielfior: There was a property like this (suggested by Evan), however mypy wasn't happy because it thought safe could still be of type `Checksum | None`, hence added this test explicity.
Happy with suggestions here though.

prediction_market_agent_tooling/tools/contract.py
gabrielfior: Actually it is - for example, here `prediction_market_agent_tooling/markets/omen/omen.py`

```
    if auto_deposit and collateral_token_balance < amount_wei:
        collateral_token_contract.deposit(
            private_credentials=private_credentials, amount_wei=amount_wei, web3=web3
        )
```
In general, adding web3 as an arg is very helpful when we execute tests and a TestWeb3 will be passed.

prediction_market_agent_tooling/markets/omen/omen.py
kongzii: You had there someting as `if self. safe_address is not None:...` and it thought it's none? Do you have some code to reproduce it? It should be just fine, same as here.

pyproject.toml
gabrielfior: moved back to ^

prediction_market_agent_tooling/tools/web3_utils.py
gabrielfior: Hehe I thought they were never being used, hence I removed - any usage that justifies putting this back on?

prediction_market_agent_tooling/tools/web3_utils.py
```diff
@@ -104,6 +107,42 @@ def call_function_on_contract(
...
+
+    if not tx_params_new.get("nonce", None):
+        from_checksummed = Web3.to_checksum_address(tx_params_new["from"])
+        tx_params_new["nonce"] = web3.eth.get_transaction_count(from_checksummed)
+
+    # Build the transaction.
+    function_call = contract.functions[function_name](*parse_function_params(function_params))  # type: ignore # TODO: Fix Mypy, as this works just OK.
+    tx_params_new = function_call.build_transaction(tx_params_new)
+    gas = web3.eth.estimate_gas(tx_params_new)
+    tx_params_new["gas"] = gas
```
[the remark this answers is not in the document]
gabrielfior: You mean specifying gas? This prevents the transaction from failing due to out-of-gas errors, which occurred when I tried on Tenderly and locally.
Setting the gas also too high is not an issue, as the unused gas will be returned to the sender.

prediction_market_agent_tooling/tools/contract.py
evangriffiths: ah my bad

prediction_market_agent_tooling/tools/web3_utils.py
kongzii: `*` enforces that the following arguments need to be passed as keywords only, it's not something that should be somehow used.

But it can be deleted if that's not desired

prediction_market_agent_tooling/markets/omen/omen_replicate.py
```diff
@@ -145,10 +141,10 @@ def omen_replicate_from_tx(
 
 
 def omen_unfund_replicated_known_markets_tx(
-    from_private_key: PrivateKey,
+    private_credentials: PrivateCredentials,
     saturation_above_threshold: float | None = None,
 ) -> None:
-    from_address = private_key_to_public_key(from_private_key)
+    from_address = private_key_to_public_key(private_credentials.private_key)
```
kongzii, line 147: Here, public key is always used, even if Safe address would be available.

What's the plan to use? Is it possible to always use safe? Or will we differentiate between it? And if so, how do we keep track of when safe is used and when it isn't?

gabrielfior: My suggestion would be to always use the safe (if available). Hence we should always call the functions (balanceOf, etc) as if the Safe was the creator ("from") of the transaction.
I will adjust this function accordingly.

gabrielfior: This has been implemented.

kongzii (commented):
There are still some usages of `bet_from_address` from APIKeys, these should be probably changed as well?

And `bet_from_address` method itself can be removed, to not create confusion

kongzii (approved):

PR #201, opened 2024-04-17
Fix Omen's market probabilities

kongzii:
## Summary by CodeRabbit

- **New Features**
  - Added properties for tracking market activity metrics like `lastActiveDay` and `lastActiveHour`.
  - Introduced methods to retrieve the last trade probabilities for yes and no options.
- **Enhancements**
  - Improved market metrics and statistics display.
  - Enhanced Brier score and market outcome bias calculations.
- **Bug Fixes**
  - Implemented retry logic for graceful handling of failures in data retrieval.
- **Refactor**
  - Enhanced clarity by updating variable and method names.
- **Tests**
  - Expanded testing suite to cover new features and refactored methods.


prediction_market_agent_tooling/benchmark/benchmark.py
```diff
@@ -208,7 +208,7 @@ def _compute_mse(
             return None
         mse = sum(
             [
-                (check_not_none(p.outcome_prediction).p_yes - m.p_yes) ** 2
+                (check_not_none(p.outcome_prediction).p_yes - m.current_p_yes) ** 2
```
kongzii, line 211: Renamed p_yes to current_p_yes so that it's clear that there is some other p_yes as well 😄

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -421,24 +433,28 @@ def get_user_positions(
...
             end_time = utcnow()
 
         trade = self.trades_subgraph.FpmmTrade
-        where_stms = [
-            trade.type == "Buy",
-            trade.creator == better_address.lower(),
-            trade.creationTimestamp >= to_int_timestamp(start_time),
-            trade.creationTimestamp <= to_int_timestamp(end_time),
-        ]
+        where_stms = []
```
evangriffiths, line 447: oo nice, I bet this speeds up the request significantly!

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -421,24 +433,28 @@ def get_user_positions(
...
         self,
-        better_address: ChecksumAddress,
-        start_time: datetime,
+        better_address: ChecksumAddress | None = None,
+        start_time: datetime | None = None,
         end_time: t.Optional[datetime] = None,
-        market_id: t.Optional[str] = None,
+        market_id: t.Optional[ChecksumAddress] = None,
         filter_by_answer_finalized_not_null: bool = False,
+        type_: t.Literal["Buy", "Sell"] | None = "Buy",
```
evangriffiths, line 441: I would say you could hard-code this to `"Buy"` as apposed to passing in as an arg. A `"Sell"` trade isn't a bet

prediction_market_agent_tooling/benchmark/benchmark.py
evangriffiths: Am I right in saying that 
- For Omen, `m.current_p_yes` == `m.get_last_trade_p_yes()` for open markets, but not for closed markets
- For Manifold (and Polymarket?) `m.current_p_yes` == `m.get_last_trade_p_yes()` always

Is there ever a time where we want to use anything other than `m.get_last_trade_p_yes()` for a closed or open market? I can't think of one!

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
kongzii: `Buy` and `Sell` are valid filters on the graph. I needed the `Sell` one as well because market probability can be changed by selling as well.

So I agree that `Sell` isn't a bet, but it's needed. Would you prefer me to rename this to `get_trades`? (it's called trades and not bets anyway on the graph.

prediction_market_agent_tooling/benchmark/benchmark.py
kongzii: Yes that's should be right.

> for a closed or open market? I can't think of one!

Theoretically, no. 

However, the issue with `get_last_trade_p_yes` is that it needs an extra API call on Omen. So if we know that we  only work with open markets (because we filtered for them or something), it's much faster to just use `current_p_yes` instead of calling `get_last_trade_p_yes`

evangriffiths: Hmmm, okay. So what about if we
- revert this back from `AgentMarket.current_p_yes` -> `AgentMarket.p_yes`
- Keep `OmenMarket.current_p_yes`
- Make `get_p_yes_history_cached` a method on `OmenMarket` instead of `OmenAgentMarket`
- In `OmenAgentMarket.from_data_model`, `p_yes` is `model.get_p_yes_history_cached` if outcome tokens == 0, else `model.current_p_yes`?

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
evangriffiths: Ah okay, yes that would be clearer. Then `get_resolved_bets` can call `get_trades` with `type_ == "Buy"`

prediction_market_agent_tooling/benchmark/benchmark.py
kongzii: I was thinking about doing it automatically in `from_data_model` as well, but such method should be very quick to execute. If I add an API calls there, it will take super long time to do it on a bigger batch of markets, probably unnecessary.

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
kongzii: Renamed, wdyt now?

prediction_market_agent_tooling/benchmark/benchmark.py
evangriffiths: Isn't it just a case of moving the slow bit from `OmenAgentMarket.from_data_model` to `OmenAgentMarket.get_last_trade_p_yes`?

The downside of this is that:
- the interface is a bit messier (`current_p_yes` and `get_last_trade_p_yes` properties/methods instead of just `p_yes` property)
- you have to have this caching logic so that multiple calls to `get_last_trade_p_yes` don't query the omen subgraph multiple times

The upside is:
- you save on having to make expensive omen subgraph queries for the case where you do `m = OmenAgentMarket.from_data_model` on closed/zero-liquidity markets, BUT don't ever want to do `m.p_yes`.

Curious if you can see a case where we encounter this upside, as at the moment we only get closed markets in the monitoring code, and there we do want to do `m.p_yes`

kongzii: > Isn't it just a case of moving the slow bit from OmenAgentMarket.from_data_model to OmenAgentMarket.get_last_trade_p_yes?

Yep, but calling `get_last_trade_p_yes` is an informed decision, rather than automatic in the background

> the interface is a bit messier

Yeah that's for sure

Here is the timing btw: if you still want to do it by default, I will surrender and implement it 😄 

Screenshot by Dropbox Capture

evangriffiths: Okay, that's quite a big time difference!

evangriffiths: One other (quite hacky) option could be to:
- keep the interface the same (i.e. `OmenAgentMarket.p_yes`).
- Inside `OmenAgentMarket.from_data_model`, do not make the API call to get the price history, so leave p_yes unset.
- Anywhere where `OmenAgentMarket.from_data_model` is called, in the outer loop make a single API call to get the bets for all relevant markets, and then set the `p_yes` values for each market.

But I don't think that's preferable. Fine, you win!!! 😆

PR #217, opened 2024-04-24
Created Langfuse wrapper for langfuse integration

gabrielfior:
## Summary by CodeRabbit

- **New Features**
	- Integrated Langfuse support across the platform to enhance agent functionality with new language processing capabilities.
	- Added configuration settings for Langfuse API keys and host details.
	- Introduced a `LangfuseWrapper` class for creating Langfuse handlers with specific API keys and session ID.

### DIFF
--- a/prediction_market_agent_tooling/benchmark/agents.py
+++ b/prediction_market_agent_tooling/benchmark/agents.py
@@ -6,9 +6,6 @@ from prediction_market_agent_tooling.benchmark.utils import (
     OutcomePrediction,
     Prediction,
 )
-from prediction_market_agent_tooling.monitor.langfuse.langfuse_wrapper import (
-    LangfuseWrapper,
-)
 
 
 class AbstractBenchmarkedAgent:
@@ -21,7 +18,6 @@ class AbstractBenchmarkedAgent:
         self.model = model
         self.agent_name = agent_name
         self.max_workers = max_workers  # Limit the number of workers that can run this worker in parallel threads
-        self.langfuse_wrapper = LangfuseWrapper(agent_name=self.__class__.__name__)
 
     def is_predictable(self, market_question: str) -> bool:
         """
--- a/prediction_market_agent_tooling/benchmark/agents.py
+++ b/prediction_market_agent_tooling/benchmark/agents.py
@@ -6,6 +6,9 @@ from prediction_market_agent_tooling.benchmark.utils import (
     OutcomePrediction,
     Prediction,
 )
+from prediction_market_agent_tooling.monitor.langfuse.langfuse_wrapper import (
+    LangfuseWrapper,
+)
 
 
 class AbstractBenchmarkedAgent:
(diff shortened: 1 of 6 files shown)


prediction_market_agent_tooling/monitor/langfuse/langfuse_wrapper.py
```diff
@@ -0,0 +1,23 @@
...
+from prediction_market_agent_tooling.config import APIKeys
+from prediction_market_agent_tooling.tools.utils import utcnow
+
+
+class LangfuseWrapper(BaseModel):
+    agent_name: str
+
+    @property
+    def session_id(self) -> str | None:
+        return f"{self.agent_name} - {utcnow()}"
```
kongzii, line 13: session_id will be different each time this property is accessed (even during he same session), is that desired?

prediction_market_agent_tooling/benchmark/agents.py
```diff
@@ -18,6 +21,7 @@ def __init__(
         self.model = model
         self.agent_name = agent_name
         self.max_workers = max_workers  # Limit the number of workers that can run this worker in parallel threads
+        self.langfuse_wrapper = LangfuseWrapper(agent_name=self.__class__.__name__)
```
kongzii, line 24: I think benchmark doesn't need langfuse, that's executed only locally. Do you see some usecase for it?

gabrielfior: Depends on what we are trying to benchmark - if only performance, and that's captured in the Python reports, then happy to exclude this - if we are also keeping track of reasoning, etc, then I keep it.

@evangriffiths could you also weigh in here please?

prediction_market_agent_tooling/monitor/langfuse/langfuse_wrapper.py
gabrielfior: The idea of `session_id` is to group traces together - my idea was to include here a kind of `run_identifier`, so every time we call `deploy_local` we will have an unique identifier `{agent}-{datetime_now}`.
This is then easily filterable, see below

image

kongzii: I get the idea that's cool. But shouldn't session_id be the same for the whole session?

gabrielfior: Refactored using @computed_property and @cache from [Pydantic docs](https://docs.pydantic.dev/2.7/api/fields/#pydantic.fields.computed_field)

prediction_market_agent_tooling/monitor/langfuse/langfuse_wrapper.py
```diff
@@ -0,0 +1,26 @@
...
+from pydantic import BaseModel, computed_field
+
+from prediction_market_agent_tooling.config import APIKeys
+from prediction_market_agent_tooling.tools.utils import utcnow
+
+
+class LangfuseWrapper(BaseModel):
+    agent_name: str
+
+    @computed_field  # type: ignore
```
kongzii, line 13:
```suggestion
    @computed_field  # type: ignore[misc] # Mypy issue: https://github.com/python/mypy/issues/14461
```

kongzii (approved):

PR #222, opened 2024-04-26
Added safe funding to script | added tests

gabrielfior:
- Added auto_funding of Safe to script that creates Safe for agents
- Added tests to functionality above

## Summary by CodeRabbit

- **New Features**
	- Introduced new method to verify ownership of Safe wallets.
	- Enhanced transaction preparation with gas estimation and parameter handling.
	- Added functionality to fund newly created Safe wallets with xDAI and log balances.
	- Implemented a new function for sending xDAI tokens with complete transaction handling.

- **Enhancements**
	- Improved `create_safe_for_agent` to verify successful Safe deployment before funding.

- **Bug Fixes**
	- Adjusted transaction-related function calls and parameters for proper execution and testing.

- **Tests**
	- Updated integration tests to accommodate new functionalities and transaction handling changes.

- **Documentation**
	- Enhanced function documentation to reflect new parameters and functionalities.

### DIFF
--- a/prediction_market_agent_tooling/tools/web3_utils.py
+++ b/prediction_market_agent_tooling/tools/web3_utils.py
@@ -147,11 +147,6 @@ def _prepare_tx_params(
         from_checksummed = Web3.to_checksum_address(tx_params_new["from"])
         tx_params_new["nonce"] = web3.eth.get_transaction_count(from_checksummed)
 
-    # gas
-    if not tx_params_new.get("gas"):
-        gas = web3.eth.estimate_gas(tx_params_new)
-        tx_params_new["gas"] = gas
-
     return tx_params_new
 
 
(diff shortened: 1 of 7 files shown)


prediction_market_agent_tooling/tools/web3_utils.py
```diff
@@ -235,3 +249,39 @@ def send_function_on_contract_tx_using_safe(
...
+
+    # We need gas and gasPrice here (and not elsewhere) because we are not calling
+    # contract.functions.myFunction().build_transaction, which auto-fills some params
+    # with defaults, incl. gas and gasPrice.
+    gas = web3.eth.estimate_gas(tx_params_new)
+    tx_params_new["gas"] = gas
+    tx_params_new["gasPrice"] = web3.eth.gas_price
+
+    # Sign with the private key.
+    signed_tx = web3.eth.account.sign_transaction(
```
kongzii, line 278: Why can't https://github.com/gnosis/prediction-market-agent-tooling/blob/53ed7dd66b0020f54a02008341f5ff47e5f13e43/prediction_market_agent_tooling/tools/web3_utils.py#L158 be used in this function?

kongzii (commented):
Left question if `send_function_on_contract_tx` can't be reused, otherwise lgtm

prediction_market_agent_tooling/tools/web3_utils.py
gabrielfior: The 2 functions are a bit different

-> `send_function_on_contract_tx` builds txParams and calls `contract.function().build_transaction()`, which fills txParams with standard functions
-> `send_xdai_...` builds txParams, does not call `contract.function().build_transaction()`, sets gas in a separate step (not needed when using contract), then signs-, sends- and checks tx.

I could (and will) consolidate some lines (signing, etc) in a common function, but I still think we need 2 methods.

kongzii (approved):

PR #224, opened 2024-04-29
Update default pick markets method

kongzii:


prediction_market_agent_tooling/markets/markets.py
```diff
@@ -52,3 +62,35 @@ def get_binary_markets(
         excluded_questions=excluded_questions,
     )
     return markets
+
+
+def have_bet_on_market_since(
```
evangriffiths, line 67: Would be nice if this were a method on `AgentMarket` class. Would need a little refactor, but this could re-use the code from `DeployedAgent.get_resolved_bets`. Maybe we could create a task for this and not worry about it for this PR?

evangriffiths (approved):

PR #227, opened 2024-04-30
Make mech client optional

kongzii:


pyproject.toml
```diff
@@ -40,12 +40,13 @@ safe-eth-py = "^6.0.0b14"
 eth-account = "^0.8.0"
 prompt-toolkit = "^3.0.43"
 safe-cli = "^1.0.0"
-mech-client = "^0.2.13"
+mech-client = { version = "^0.2.13", optional = true }
```
gabrielfior, line 43: If this is marked as optional, I understand that the PMA agents that use this won't run if not installed with `extras=[mech-client]`, correct?
Should we update PMA for a warning of this or similar? Or is this common practice?

kongzii: PMA will error out and mech need to be added to extas for PMAT dependency, yes.

> Should we update PMA for a warning of this or similar? Or is this common practice?

If there is a breaking change, either in the code or like this one, the practice is to bump the version accordingly.

PMA is dependent on ^0.20.0, so this change won' be picked up automatically because of 

Screenshot by Dropbox Capture

but as soon as someone update the version in PMA, this needs to be fixed yes. ( I'll do that )

pyproject.toml
```diff
@@ -40,12 +40,13 @@ safe-eth-py = "^6.0.0b14"
...
 langfuse = "^2.27.1"
 
 [tool.poetry.extras]
 langchain = ["langchain", "langchain-openai"]
 google = ["google-api-python-client"]
+mech = ["mech-client"]
```
evangriffiths, line 49:
```suggestion
mech-client = ["mech-client"]
```
?

gabrielfior (approved):

pyproject.toml
kongzii: Sure, no problem with that, commited

evangriffiths (approved):

PR #229, opened 2024-05-02
Loguru for GCP

kongzii:


evangriffiths (approved):

prediction_market_agent_tooling/loggers.py
```diff
@@ -0,0 +1,54 @@
+import sys
+import typing as t
+from enum import Enum
+
+from loguru import logger
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+if t.TYPE_CHECKING:
+    from loguru import Logger
```
gabrielfior, line 9: Why is TYPE_CHECKING needed here? I thought Loguru would also be imported by 3rd parties (e.g. PMA), no?

kongzii: There are two ways how to do typing in Python.

1. This Is what we are doing, and 99% of other libraries as well, as it's the preferred way. Typing directly in the code, e.g. we have functions as `def do_something(x: str) -> int: ...`.

2. Is to have just `def do_something(x): ...` and provide typing in the second file called `stub`, see here for more details: https://stackoverflow.com/questions/59051631/what-is-the-use-of-stub-files-pyi-in-python

In the second case, the typing is not available during run-time, and therefore we need this `if TYPE_CHECKING` condition here.

Also note that the `Logger` is used in the strings in the typing, because it won't be available at runtime:

Screenshot by Dropbox Capture

PR #236, opened 2024-05-07
Support to deploy non-betting agents

kongzii:


prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -83,63 +83,31 @@ def p_no(self) -> Probability:
         return Probability(1 - self.p_yes)
 
 
-class DeployableAgent:
-    bet_on_n_markets_per_run: int = 1
-
+class DeployableRunAgent:
```
evangriffiths, line 86: Just my opinion on naming...

Both `DeployableRunAgent` and `DeployableBetAgent` have a `run` method. So I'm not sure on `Run` as a distinguishing part of the name. I like
- `DeployableAgent` (for the base, non betting agent)
- `DeployableTraderAgent` (borrowing from Olas' naming)

wdyt?

kongzii: Sure, looks good, renamed.

evangriffiths (approved):

PR #237, opened 2024-05-07
Remove mech-client as a dependency

evangriffiths:


pyproject.toml
```diff
@@ -40,13 +40,11 @@ safe-eth-py = "^6.0.0b14"
 eth-account = "^0.8.0"
 prompt-toolkit = "^3.0.43"
 safe-cli = "^1.0.0"
-mech-client = { version = "^0.2.13", optional = true }
 langfuse = "^2.27.1"
 
 [tool.poetry.extras]
 langchain = ["langchain", "langchain-openai"]
 google = ["google-api-python-client"]
-mech-client = ["mech-client"]
```
kongzii, line 49: If this was already optional, maybe we can just leave it here? But I'm fine with removing it too.

kongzii (approved):

pyproject.toml
evangriffiths: kamikazeeeeee

PR #272, opened 2024-06-10
OmenAgentMarket.sell_tokens gets token pool balance dynamically

evangriffiths:
... instead of saving it at constriction time.

This fixes an issue where you could do
```
market = get_market()
market.buy_tokens(amount) # Pool balance is now incorrect!
market.sell_tokens(num_tokens) # num_tokens -> collateral calculation is now wrong
```

This therefore fixes the flaky tests_integration/markets/omen/test_omen.py::test_omen_buy_and_sell_outcome

### DIFF
--- a/tests_integration/markets/omen/test_omen.py
+++ b/tests_integration/markets/omen/test_omen.py
@@ -258,4 +258,4 @@ def test_omen_buy_and_sell_outcome(
 
     # Check that we have sold our entire stake in the market.
     remaining_tokens = get_market_outcome_tokens()
-    assert np.isclose(remaining_tokens.amount, 0, atol=1e-3)
+    assert np.isclose(remaining_tokens.amount, 0, atol=1e-6)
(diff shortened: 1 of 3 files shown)


gabrielfior (approved):

gabrielfior:
@evangriffiths still don't think this can be merged since Pytest is not happy, but I'm OK with the code changes

tests_integration/markets/omen/test_omen.py
```diff
@@ -258,4 +258,4 @@ def get_market_outcome_tokens() -> TokenAmount:
 
     # Check that we have sold our entire stake in the market.
     remaining_tokens = get_market_outcome_tokens()
-    assert np.isclose(remaining_tokens.amount, 0, atol=5e-3)
+    assert np.isclose(remaining_tokens.amount, 0, atol=1e-6)
```
evangriffiths, line 261: Much tighter test tolerance now!

PR #278, opened 2024-06-17
Making PMAT dependencies compatible with Olas mechs

gabrielfior:


pyproject.toml
```diff
@@ -34,10 +34,11 @@ google-cloud-secret-manager = "^2.18.2"
 langchain = { version = "^0.1.9", optional = true}
 langchain-openai = { version = "^0.0.5", optional = true}
 google-api-python-client = { version = "2.95.0", optional = true}
-subgrounds = "^1.8.1"
+# Use fork while https://github.com/0xPlaygrounds/subgrounds/pull/57 is not merged
```
kongzii, line 37: I'll leave that for subgrounds maintainers, but fyi this this change

Screenshot by Dropbox Capture

make versioning more strict (`>=2.6.1, <3` instead of `>=2,<3`), so idk why making it more strict would help the Olas conflicts?

kongzii (approved):

pyproject.toml
gabrielfior: This is the version I used and where conflicts were solved - agree that this is more strict.

PR #283, opened 2024-06-24
Add AgentMarket.get_positions_value

evangriffiths:


prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -421,6 +421,62 @@ def get_positions(cls, user_id: str, liquid_only: bool = False) -> list[Position
 
         return positions
 
+    @classmethod
+    def get_positions_value(cls, positions: list[Position]) -> BetAmount:
+        # Two dicts to map from market ids to (1) postions and (2) market.
```
gabrielfior, line 426: typo

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -421,6 +421,62 @@ def get_positions(cls, user_id: str, liquid_only: bool = False) -> list[Position
...
+            no_tokens = 0.0
+            if OMEN_TRUE_OUTCOME in position.amounts:
+                yes_tokens = position.amounts[OutcomeStr(OMEN_TRUE_OUTCOME)].amount
+            if OMEN_FALSE_OUTCOME in position.amounts:
+                no_tokens = position.amounts[OutcomeStr(OMEN_FALSE_OUTCOME)].amount
+
+            # Account for the value of positions in resolved markets
+            if market.is_resolved() and market.has_successful_resolution():
+                valued_tokens = yes_tokens if market.boolean_outcome else no_tokens
+                total_position_value += valued_tokens
```
gabrielfior, line 459: Can you please remind me how one can derive the value of a position on a resolved market?
Let's say there is 100 xDAI in collateral and 100 Y and 100 N tokens were issued. Now, the outcome was true. If Bob had 10 Y tokens, is the value always going to be 10 xDAI (i.e. a "correct outcome token" is always worth 1 xDAI after resolution)?
I wasn't aware of this calculation, hence my question.

evangriffiths: Yes, that's my understanding. Not sure exactly where in the smart contract it is, but it makes sense if you consider your example, and nobody places a trade before the market closes. If e.g. Y is correct, then the 100 Y tokens are worth 100 xDai, and the 100 N tokens are worth zero. So the funder of the market gets all their money back.

tests/markets/omen/test_omen.py
```diff
@@ -145,3 +145,13 @@ def test_get_positions_1() -> None:
...
+
+
+def test_positions_value() -> None:
+    # Pick a user that has active positions
+    user_address = Web3.to_checksum_address(
+        "0x2DD9f5678484C1F59F97eD334725858b938B4102"
+    )
+    positions = OmenAgentMarket.get_positions(user_id=user_address)
+    value = OmenAgentMarket.get_positions_value(positions=positions)
+    assert value.amount > 0
```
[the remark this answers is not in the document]
evangriffiths: @gabrielfior that's a fair point. The test was flakey. I like your suggestion, but I'd rather not introduce a test-time-only arg if possible. An alternative is to just pass a manually constructed `Position` to `OmenAgentMarket.get_positions_value`. In this case the position is based on a historical bet, so we can also use `OmenBet.get_profit` to double-check that our calculations are correct. wdyt?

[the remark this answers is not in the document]
gabrielfior: I like this fixed `Position` option as well, since we are only testing the logic behind it. Works for me!

prediction_market_agent_tooling/markets/omen/omen.py
gabrielfior: Cool, but what happens if we have 100Y and 100N tokens and, 1s before the market is resolved, the funder removed almost all the liquidity? Do you have such a case for me to take a look at?
Note that this case is not part of the `if` condition from line 457, so agree with the logic.

prediction_market_agent_tooling/markets/agent_market.py
```diff
@@ -114,6 +114,16 @@ def get_last_trade_p_no(self) -> Probability | None:
         """
         raise NotImplementedError("Subclasses must implement this method")
 
+    def get_last_trade_yes_outcome_price(self) -> float | None:
+        if last_trade_p_yes := self.get_last_trade_p_yes():
```
kongzii, line 118:
```suggestion
        # Price on prediction markets are, by definition, equal to the probability of an outcome.
        # Just making it explicit in this function.
        if last_trade_p_yes := self.get_last_trade_p_yes():
```

I was confused for a min why doing this, just a comment for future confusers please 😄

prediction_market_agent_tooling/markets/agent_market.py
```diff
@@ -114,6 +114,16 @@ def get_last_trade_p_no(self) -> Probability | None:
...
+            return float(last_trade_p_yes)
+        return None
+
+    def get_last_trade_no_outcome_price(self) -> float | None:
+        if last_trade_p_no := self.get_last_trade_p_no():
```
kongzii, line 123:
```suggestion
        # Price on prediction markets are, by definition, equal to the probability of an outcome.
        # Just making it explicit in this function.
        if last_trade_p_no := self.get_last_trade_p_no():
```

Same here 🙏

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -421,6 +421,62 @@ def get_positions(cls, user_id: str, liquid_only: bool = False) -> list[Position
 
         return positions
 
+    @classmethod
+    def get_positions_value(cls, positions: list[Position]) -> BetAmount:
+        # Two dicts to map from market ids to (1) positions and (2) market.
+        market_ids_positions = {p.market_id: p for p in positions}
```
kongzii, line 427: I think it can happen that agent has multiple positions in the same market, in which case, this will map only to its last position, and the computation will be off.

prediction_market_agent_tooling/markets/omen/omen.py
kongzii: > Cool, but what happens if we have 100Y and 100N tokens and, 1s before the market is resolved, the funder removed almost all the liquidity?

Market creators can not remove liquidity of sold tokens (not sure that's the correct terminology). But to your example, if you hold 100 tokens of correct outcome, market creator can not remove their value. You will be able to exchange them for your promised xDais.

prediction_market_agent_tooling/markets/omen/omen.py
evangriffiths: Good spot! `AgentMarket.get_positions` will only turn one position per market, regardless of how many different trades the user has placed on that market though. So I've just added a check that the user doesn't pass in multiple positions for a market.

kongzii: Seems like `get_positions`  should be modifiable to return all of them (or to aggregate all trades into a single position).

If you don't see some problems there, should we have an issue to tackle in another PR?

Screenshot by Dropbox Capture

evangriffiths: > Seems like get_positions should be modifiable to return all of them (or to aggregate all trades into a single position).

Sorry, I don't understand what you mean here. Can you explain?

kongzii: Maybe I don't understand how it works 🙈  

If I have 10 open trades on 1 market, `AgentMarket.get_positions` will

1. Return a single position, for one of the trades
2. Return a single position, but it will reflect (somehow aggregate) all the trades?

If (1), then I'm proposing that we should have an issue to fix that and return all the trades for all the markets.

kongzii (approved):

prediction_market_agent_tooling/markets/omen/omen.py
evangriffiths: It's (2). It aggregates them.

Screenshot 2024-06-25 at 12 20 55
Screenshot 2024-06-25 at 12 21 46

PR #287, opened 2024-06-25
Add `larger_than` arg to AgentMarket.get_positions

evangriffiths:
Fixes https://github.com/gnosis/prediction-market-agent/issues/226

### DIFF
--- a/prediction_market_agent_tooling/markets/omen/omen.py
+++ b/prediction_market_agent_tooling/markets/omen/omen.py
@@ -376,7 +376,7 @@ class OmenAgentMarket(AgentMarket):
         sgh = OmenSubgraphHandler()
         omen_positions = sgh.get_user_positions(
             better_address=Web3.to_checksum_address(user_id),
-            total_balance_bigger_than=xdai_to_wei(larger_than),
+            total_balance_bigger_than=xdai_to_wei(xDai(larger_than)),
         )
 
         # Sort positions and corresponding markets by condition_id
--- a/prediction_market_agent_tooling/markets/omen/omen.py
+++ b/prediction_market_agent_tooling/markets/omen/omen.py
@@ -367,11 +367,16 @@ class OmenAgentMarket(AgentMarket):
         )
 
     @classmethod
-    def get_positions(cls, user_id: str, liquid_only: bool = False) -> list[Position]:
+    def get_positions(
+        cls,
+        user_id: str,
+        liquid_only: bool = False,
+        larger_than: float = 0,
+    ) -> list[Position]:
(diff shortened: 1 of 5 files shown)


prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -367,11 +367,16 @@ def get_token_balance(
...
+        cls,
+        user_id: str,
+        liquid_only: bool = False,
+        larger_than: float = 0,
+    ) -> list[Position]:
         sgh = OmenSubgraphHandler()
         omen_positions = sgh.get_user_positions(
             better_address=Web3.to_checksum_address(user_id),
-            total_balance_bigger_than=wei_type(0),
+            total_balance_bigger_than=xdai_to_wei(xDai(larger_than)),
```
kongzii, line 379: Mypy should be okay with this as xDai comes from float, but if not, then your implementation is fine of course.

```suggestion
            total_balance_bigger_than=xdai_to_wei(larger_than),
```

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -367,11 +367,16 @@ def get_token_balance(
         )
 
     @classmethod
-    def get_positions(cls, user_id: str, liquid_only: bool = False) -> list[Position]:
+    def get_positions(
+        cls,
+        user_id: str,
+        liquid_only: bool = False,
+        larger_than: float = 0,
```
kongzii, line 374:
```suggestion
        larger_than: xDai = 0,
```

prediction_market_agent_tooling/markets/omen/omen.py
evangriffiths: I did try that, but my is unhappy with this:
```
% mypy
prediction_market_agent_tooling/markets/omen/omen.py:374: error: Argument 3 of "get_positions" is incompatible with supertype "AgentMarket"; supertype defines the argument type as "float"  [override]
prediction_market_agent_tooling/markets/omen/omen.py:374: note: This violates the Liskov substitution principle
prediction_market_agent_tooling/markets/omen/omen.py:374: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
```

From the mypy docs:
> It’s unsafe to override a method with a more specific argument type

So I guess mypy sees `xDai` as a more specific type of `float` (as apposed to just an alias)

kongzii: Whaat, I thought it works the other way around! Okay then, nevermind :(

kongzii (approved):

PR #292, opened 2024-07-01
Add subgraph for omen image mapping

kongzii:


prediction_market_agent_tooling/markets/omen/omen_contracts.py
```diff
@@ -630,6 +630,10 @@ class OmenThumbnailMapping(ContractOnGnosisChain):
         "0xe0cf08311F03850497B0ed6A2cf067f1750C3eFc"
     )
 
+    @staticmethod
+    def construct_ipfs_url(ipfs_hash: IPFSCIDVersion0) -> str:
+        return f"https://ipfs.io/ipfs/{ipfs_hash}"
+
```
gabrielfior, line 636: Can you please remind me what are we using for IPFS pining?
I'm asking this because Olas for ex uses Gateway IPFS, so the URL above would not work. 
For ex,
this works
https://gateway.autonolas.tech/ipfs/f0170122085e6da28d342ee64d40b38c97d01bdab2ed1093ae337aed44e8b6228a13d5749
this doesn't
http://ipfs.io/ipfs/f0170122085e6da28d342ee64d40b38c97d01bdab2ed1093ae337aed44e8b6228a13d5749

so just want to make sure the URL above is always valid.

prediction_market_agent_tooling/markets/omen/omen_contracts.py
```diff
@@ -630,6 +630,10 @@ class OmenThumbnailMapping(ContractOnGnosisChain):
         "0xe0cf08311F03850497B0ed6A2cf067f1750C3eFc"
     )
 
+    @staticmethod
```
gabrielfior, line 633: Another question - example market_id from your subgraph
```
{
        "id": "0x0ef8ad19551fcf5bb852ea73559656cbfb175975",
        "marketAddress": "0x0ef8ad19551fcf5bb852ea73559656cbfb175975",
        "image_hash": "0x1eb5c8fc362f8420850ca8a7496a4d64c0344f6fbaa8e3e1e3bb6dfc6bc2580e"
      },
```
I wanted to retrieve the image using the image_hash above, I tried `https://www.ipfs.com/ipfs/0x1eb5c8fc362f8420850ca8a7496a4d64c0344f6fbaa8e3e1e3bb6dfc6bc2580e` but it didn't return the image. Is it expected?

kongzii: Yes, this is related to what we talked about before on Slack, about Bytes32 and the fact that IPFS hash is longer than 32 bytes by default.

The hash stored in the contract is in bytes32 format, converted from ipfscidv0 here https://github.com/gnosis/prediction-market-agent-tooling/blob/54d2dac4ee23d8584ec792c18dbf65a605ffca98/prediction_market_agent_tooling/markets/omen/omen_contracts.py#L665.

So it needs to be converted back as you can see in this PR here https://github.com/gnosis/prediction-market-agent-tooling/blob/54d2dac4ee23d8584ec792c18dbf65a605ffca98/prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py#L641

gabrielfior: Thanks for clarifying. Calling `bytes32_to_ipfscidv0` yields https://ipfs.io/ipfs/QmQQXiFsPo5cj9ay4PiBnuqWe3pd3LAi5PZX5wzUeY3zwo, works perfectly, thanks.

gabrielfior (approved):

PR #297, opened 2024-07-03
Retry in omen buy outcome

kongzii:


prediction_market_agent_tooling/tools/web3_utils.py
```diff
@@ -157,7 +157,7 @@ def _prepare_tx_params(
     # Retry only for the transaction errors that match the given patterns,
     # add other retrieable errors gradually to be safe.
     retry=tenacity.retry_if_exception_message(
-        match="(.*wrong transaction nonce.*)|(.*Invalid.*)|(.*OldNonce.*)"
+        match="(.*wrong transaction nonce.*)|(.*Invalid.*)|(.*OldNonce.*)|(.*revert.*)"
```
gabrielfior, line 160: If it reverts, it's not worth retrying - there is a condition in the smart contract that is not verified, hence it will fail even after retrying

kongzii: Doesn't `revert` occur also on some random occasions?

For example, this happened today:

Screenshot by Dropbox Capture


And I don't see any good reason for that bet placing to fail.

kongzii: Maybe instead of retrying the exact same transaction, we should retry the whole place_bet?

kongzii: Changed

gabrielfior (approved):

PR #299, opened 2024-07-05
Add partial Metaculus support

evangriffiths:


prediction_market_agent_tooling/markets/metaculus/api.py
```diff
@@ -0,0 +1,94 @@
...
+        f"{METACULUS_API_BASE_URL}/comments/",
+        json={
+            "comment_text": comment_text,
+            "submit_type": "N",
+            "include_latest_prediction": True,
+            "question": question_id,
+        },
+        headers=get_auth_headers(),
+    )
+    response.raise_for_status()
```
gabrielfior, line 35: For reference - you could also use the openapi-python-generator with their YAML file (https://github.com/openapi-generators/openapi-python-client), this way you avoid having to hardcode routes etc.

prediction_market_agent_tooling/markets/metaculus/api.py
```diff
@@ -0,0 +1,94 @@
...
+        url_params["project"] = tournament_id
+    if created_after:
+        url_params["created_time__gt"] = created_after.isoformat()
+    if status:
+        url_params["status"] = status
+
+    url = f"{METACULUS_API_BASE_URL}/questions/"
+    response = requests.get(url, headers=get_auth_headers(), params=url_params)
+    response.raise_for_status()
+    response_json = json.loads(response.content)
```
gabrielfior, line 93: response.json() is more elegant 🤓

prediction_market_agent_tooling/markets/metaculus/data_models.py
```diff
@@ -0,0 +1,89 @@
...
+    username: str
+    question: int
+
+
+class CommunityPredictionStats(BaseModel):
+    full: CommunityPrediction
+    unweighted: CommunityPrediction
+
+
+class MetaculusQuestion(BaseModel):
```
gabrielfior, line 53: This file also gets auto-generated from the YAML

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "prediction-market-agent-tooling"
-version = "0.40.1"
+version = "0.41.0"
```
gabrielfior, line 3: I expect some conflicts with other PRs that are also bumping the version

prediction_market_agent_tooling/monitor/markets/metaculus.py
```diff
@@ -0,0 +1,43 @@
...
+
+    @classmethod
+    def from_all_gcp_functions(
+        cls: t.Type["DeployedMetaculusAgent"],
+        filter_: t.Callable[[Function], bool] = lambda function: function.labels[
+            MARKET_TYPE_KEY
+        ]
+        == MarketType.METACULUS.value,
+    ) -> t.Sequence["DeployedMetaculusAgent"]:
+        return super().from_all_gcp_functions(filter_=filter_)
```
gabrielfior, line 43: Sorry, why do we need this?

kongzii: It's from good old days when we used GCP Functions for deployment. This was used to automatically get all agents betting on some market. Since other agent classes still have this as well, I think it's nice to have it here too.

kongzii (approved):

prediction_market_agent_tooling/markets/metaculus/api.py
evangriffiths: Interesting. Thanks, good to know.

pyproject.toml
evangriffiths: Race you!

prediction_market_agent_tooling/markets/metaculus/api.py
evangriffiths: > (not sure if you two don't like it or it's just too easy to forget we have it 😄 )

Forgot we had it. This is a lovely, thanks for the reminder!

prediction_market_agent_tooling/monitor/markets/metaculus.py
evangriffiths: It's also needed for these tests to pass https://github.com/gnosis/prediction-market-agent-tooling/blob/main/tests/monitor/test_monitor.py 😄

kongzii: > It's also needed for these tests to pass

The best reason ever 😄

prediction_market_agent_tooling/markets/metaculus/data_models.py
```diff
@@ -0,0 +1,89 @@
...
+    q1: float | None = None
+    q2: float | None = None
+    q3: float | None = None
+
+    @property
+    def p_yes(self) -> float:
+        """
+        q2 corresponds to the median, or 'second quartile' of the distribution.
+
+        If no value is provided, we default to 0.5.
```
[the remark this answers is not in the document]
evangriffiths: > What does it mean is no value is provided?

This is the case when a market is new, so has no prediction. I've clarified in the comment. Would you prefer if this logic was moved to `MetaculusAgentMarket.from_data_model`? i.e.

```
if p_yes := model.community_prediction.full.p_yes
  current_p_yes = p_yes
else:
  current_p_yes = 0.5
```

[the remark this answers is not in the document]
kongzii: No no this is fine, makes sense for new markets without predictions

PR #300, opened 2024-07-05
Gabriel/resolved bets in agent market

gabrielfior:
See PR comments here 

https://github.com/gnosis/prediction-market-agent/pull/306/files#r1664311729


prediction_market_agent_tooling/markets/data_models.py
```diff
@@ -12,6 +13,9 @@ class Currency(str, Enum):
     Mana = "Mana"
     USDC = "USDC"
 
+    def __str__(self) -> str:
```
evangriffiths, line 16: This shouldn't be needed - this is the default behaviour for string enums

prediction_market_agent_tooling/markets/data_models.py
```diff
@@ -24,6 +28,9 @@ class TokenAmount(BaseModel):
     amount: float
     currency: Currency
 
+    def __str__(self) -> str:
+        return "Amount {} currency {}".format(self.amount, self.currency)
```
evangriffiths, line 32: not a biggie, but why not use f-strings like everywhere else?

scripts/example_tool_metadata.json
```diff
@@ -0,0 +1,35 @@
+ {
```
evangriffiths, line 1: This looks like it shouldn't be in this PR 😄

prediction_market_agent_tooling/markets/data_models.py
gabrielfior: They output 'Currency.Mana' - if that's OK, then happy to keep it like that.

gabrielfior: removed

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -336,6 +337,20 @@ def get_bets_made_since(
...
+    ) -> list[ResolvedBet]:
+        subgraph_handler = OmenSubgraphHandler()
+        bets = subgraph_handler.get_resolved_bets_with_valid_answer(
+            better_address=better_address,
+            start_time=start_time,
+            end_time=None,
+            market_id=None,
+        )
+        generic_bets = [b.to_generic_resolved_bet() for b in bets]
+        return generic_bets
```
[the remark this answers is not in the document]
evangriffiths: IMO might as well make it as generic as possible here, and pass in an `end_time` arg too.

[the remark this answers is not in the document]
gabrielfior: added end_time

prediction_market_agent_tooling/markets/data_models.py
```diff
@@ -34,6 +38,10 @@ class Bet(BaseModel):
     outcome: bool
     created_time: datetime
     market_question: str
+    market_id: HexAddress
```
kongzii, line 41: This is a general `Bet` class,  so this should be just a string.

```suggestion
    market_id: str
```

And then you don't have to do `market_id=HexAddress(HexStr(market.id)),` in  `manifold_to_generic_resolved_bet`, because Manifold ID is not a HexAddress

prediction_market_agent_tooling/markets/manifold/api.py
```diff
@@ -213,6 +213,7 @@ def manifold_to_generic_resolved_bet(
         outcome=bet.get_resolved_boolean_outcome(),
         created_time=bet.createdTime,
         market_question=market.question,
+        market_id=HexAddress(HexStr(market.id)),
```
kongzii, line 216:
```suggestion
        market_id=market.id,
```

kongzii (approved):

PR #302, opened 2024-07-08
Gabriel/update langchain

gabrielfior:


gabrielfior:
This update is needed by https://github.com/gnosis/prediction-market-agent/pull/307

scripts/example_tool_metadata.json
```diff
@@ -0,0 +1,35 @@
+ {
```
kongzii, line 1: Should this be in `scripts/` folder?

scripts/upload_olas_tool_metadata.py
```diff
@@ -0,0 +1,105 @@
...
+    # ToDo - Add input like this
+    """
+     api_key: str | None = kwargs.get("api_keys", {}).get("openai", None)
+    if api_key is None:
+        return error_response("No api key has been given.")
+
+    gnosis_rpc_url: str | None = kwargs.get("api_keys", {}).get("gnosis_rpc_url", None)
+
+    """
+    print("oi")
```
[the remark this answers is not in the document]
kongzii:
```suggestion
```

scripts/upload_olas_tool_metadata.py
```diff
@@ -0,0 +1,105 @@
...
+                ),
+                "api_keys": SchemaProperty(
+                    type="object", description="RPC_URL", schema=Schema()
+                ),
+            },
+            required=["text"],
+        ),
+    )
+    # ToDo - Output should be a schema
+    output_properties = {
```
kongzii, line 68: This can be Pydantic model as well, can't it?

scripts/upload_olas_tool_metadata.py
```diff
@@ -0,0 +1,105 @@
...
+        ),
+    )
+    metadata = Metadata(
+        name=tool_name, description=description, input=input, output=output
+    )
+
+    filename = "metadata.json"
+    with open(filename, "w") as json_file:
+        # Write the dictionary to the file as JSON
+        json.dump(metadata.model_dump_json(), json_file, indent=4)
```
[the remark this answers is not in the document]
kongzii: Rabbit is correct, model_dump_json already gives the model as JSON-string, so this will be JSON-ized JSON string, is that correct?

scripts/example_tool_metadata.json
gabrielfior: Removed this file completely

scripts/upload_olas_tool_metadata.py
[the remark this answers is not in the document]
gabrielfior: Also removed this file

scripts/upload_olas_tool_metadata.py
gabrielfior: It could, but file deleted

prediction_market_agent_tooling/config.py
```diff
@@ -82,7 +83,7 @@ def bet_from_address(self) -> ChecksumAddress:
         return self.SAFE_ADDRESS if self.SAFE_ADDRESS else self.public_key
 
     @property
-    def openai_api_key(self) -> SecretStr:
+    def openai_api_key(self) -> SecretStrV1:
```
kongzii, line 86: If we do this, there will be troubles with other libraries that are updated to the latest pydantic secretstr version.

Instead of changing `openai_api_key` to `SecretStrV1`, add please property `openai_api_key_secretstr_v1` (or anything similar), you can use https://github.com/gnosis/prediction-market-agent-tooling/blob/bac961f73de2b0dce8b0892a484834dadf3683db/prediction_market_agent_tooling/gtypes.py#L78 to convert it

evangriffiths: > there will be troubles

@kongzii  Could you explain?

evangriffiths: Ohh I see, sorry just being slow this morning. Didn't realise `SecretStrV1` was older than `SecretStr`. Weird then that updating langchain requires using older `SecretStrV1`

gabrielfior: sure, done

kongzii (approved):

PR #303, opened 2024-07-08
Add full manifold market

kongzii:


evangriffiths (approved):
Coool. What's this needed for?

kongzii:
I wanted to use this to get category of manifold's market, to filter our markets about manifold itself (as proposed on TG), unfortunately, they don't publish category in this "full" model either. But maybe it will be handy in the future 🤷

PR #307, opened 2024-07-12
is_predictable_without_description?

kongzii:


prediction_market_agent_tooling/tools/is_predictable.py
```diff
@@ -42,33 +84,67 @@ def is_predictable_binary(
...
         from langchain_openai import ChatOpenAI
     except ImportError:
-        logger.info("langchain not installed, skipping is_predictable_binary")
+        logger.error("langchain not installed, skipping is_predictable_binary")
         return True
 
     llm = ChatOpenAI(
         model=engine,
-        temperature=0.0,
+        temperature=LLM_SUPER_LOW_TEMPERATURE,
```
evangriffiths, line 92: pour quoi?

kongzii: parce que 

```
# t=0 is mathematically impossible and it's not clear how OpenAI (and others) handle it, as a result, even with t=0, gpt-4-turbo produces very different outputs,
# it seems that using a very low temperature is the best way to have as consistent outputs as possible: https://community.openai.com/t/why-the-api-output-is-inconsistent-even-after-the-temperature-is-set-to-0/329541/12
```

prediction_market_agent_tooling/tools/utils.py
```diff
@@ -19,6 +19,10 @@
 
 T = TypeVar("T")
 
+# t=0 is mathematically impossible and it's not clear how OpenAI (and others) handle it, as a result, even with t=0, gpt-4-turbo produces very different outputs,
```
evangriffiths, line 22: > t=0 is mathematically impossible

😮 I didn't know this. Just had a google and apparently the logits are scaled by 1/T, so T==0 means divide-by-0 error! I looked at the docs and for the [chat api](https://platform.openai.com/docs/api-reference/chat/create) it's vague about what T==0 means, which is weird:

> What sampling temperature to use, between 0 and 2. Higher values like 0.8 will make the output more random, while lower values like 0.2 will make it more focused and deterministic.

For the [audio transcription api](https://platform.openai.com/docs/api-reference/audio/createTranscription#audio-createtranscription-temperature) though it says

> If set to 0, the model will use [log probability](https://en.wikipedia.org/wiki/Log_probability) to automatically increase the temperature until certain thresholds are hit.

How is there no good documentation on what T==0 means for chat?!?!?

tests/tools/test_is_predictable.py
```diff
@@ -41,4 +44,54 @@
 def test_is_predictable_binary(question: str, answerable: bool) -> None:
     assert (
         is_predictable_binary(question=question) == answerable
-    ), f"Question is not evaluated correctly, see the completion: {is_predictable_binary}"
+    ), f"Question is not evaluated correctly."
+
+
+@pytest.mark.skipif(not RUN_PAID_TESTS, reason="This test costs money to run.")
```
evangriffiths, line 50: Nice test!

prediction_market_agent_tooling/tools/utils.py
kongzii: > Have you done any experiments in the context of this PR that would suggest T==LLM_SUPER_LOW_TEMPERATURE is more reliable than T==0? If not I think we should leave it as T==0

Yes, a few. If I run

```
ENABLE_CACHE=false RUN_PAID_TESTS=1 pytest tests/ -k test_is_predictable_without_description  -vvv
```

It seems that it's more consistent, but still not 100% the same. Before with `t=0`, more test cases would fail and then pass again.

If I run it now, 1 fails, even though before all of them passed. 

But I don't feel 100% confident about it to replace `t=0` with `t= LLM_SUPER_LOW_TEMPERATURE` everywhere 😄

evangriffiths: Fair enough!

evangriffiths (approved):

kongzii:
After trying to restart tests for 5x, I' just merging because the failures aren't related to this PR :-D

PR #308, opened 2024-07-12
Add description to agent market

kongzii:


prediction_market_agent_tooling/markets/metaculus/metaculus.py
```diff
@@ -25,6 +25,9 @@ class MetaculusAgentMarket(AgentMarket):
 
     have_predicted: bool
     base_url: t.ClassVar[str] = METACULUS_API_BASE_URL
+    description: str | None = (
```
evangriffiths, line 28: They do have `Resolution Criteria` and `Fine Print`, which are similar to manifold descriptions. Could do `description = resolution_criteria` or `description = resolution_criteria + "\n" + fine_print`.

Happy if you create an issue for a separate PR though.

e.g. see https://www.metaculus.com/questions/25755/biden-announces-drop-out-before-july-15-2024/

kongzii: Seems like they aren't available in the MetaculusQuestion model: https://github.com/gnosis/prediction-market-agent-tooling/blob/1d242ca73a4855d59dc61c1faf92307eb9153c41/prediction_market_agent_tooling/markets/metaculus/data_models.py#L54

@coderabbitai Please create an issue to check out if we can get this extra information from some API.

evangriffiths: Ah you're right, from a quick look it seems the fields are only available on the question creation and updating endpoints `https://www.metaculus.com/api2/schema/redoc/#tag/questions/operation/questions_list`

evangriffiths (approved):

PR #320, opened 2024-07-23
Create markets with sDai as collateral

kongzii:


prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -615,7 +622,10 @@ def omen_sell_outcome_tx(
 
     market_contract: OmenFixedProductMarketMakerContract = market.get_contract()
     conditional_token_contract = OmenConditionalTokenContract()
-    collateral_token = OmenCollateralTokenContract()
+    collateral_token_contract = market_contract.get_collateral_token_contract()
+    assert isinstance(
+        collateral_token_contract, ContractERC20OnGnosisChain
+    ), "TODO: Implement for the ERC-4626 case."
```
kongzii, line 628: Created an issue for these https://github.com/gnosis/prediction-market-agent-tooling/issues/321

prediction_market_agent_tooling/tools/contract.py
```diff
@@ -240,3 +311,37 @@ class ContractERC20OnGnosisChain(ContractERC20BaseClass, ContractOnGnosisChain):
...
+    """
+
+
+def init_erc4626_or_erc20_contract(
+    address: ChecksumAddress,
+) -> ContractERC20OnGnosisChain | ContractERC4626OnGnosisChain:
+    """
+    Checks if the given contract is ERC-20 or ERC-4626 and returns the appropriate class instance.
+    Throws an error if the contract is neither of them.
+    TODO: Is there a better way to check if the address adheres to some standard, than trying to call some random function we believe aren't present on the other standard?
```
evangriffiths, line 328: 🫢

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -710,25 +737,54 @@ def omen_create_market_tx(
...
+                    asset_token_contract.deposit(api_keys, initial_funds_wei, web3=web3)
+
+                # Deposit the erc-20 token to the vault.
+                # We deposit erc-20 `initial_funds_wei`, but then we will have around `initial_funds_wei_in_shares` in the vault.
+                collateral_token_contract.deposit(
+                    api_keys, initial_funds_wei, from_address, web3=web3
+                )
+
+        else:
+            raise ValueError("Bug in our logic! :(")
```
evangriffiths, line 787: things are getting complicated now!

evangriffiths (approved):

prediction_market_agent_tooling/markets/omen/omen.py
kongzii: Hopefully no other ERC standard can be used as collateral 😄

prediction_market_agent_tooling/tools/contract.py
kongzii: Forgot about this one, @coderabbitasi please create an issue for this TODO comment.

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -694,10 +707,24 @@ def omen_create_market_tx(
 
     realitio_contract = OmenRealitioContract()
     conditional_token_contract = OmenConditionalTokenContract()
-    collateral_token_contract = OmenCollateralTokenContract()
+    collateral_token_contract = init_erc4626_or_erc20_contract(collateral_token_address)
```
gabrielfior, line 710: I am confused here and am probably missing something - let's assume we have xDAI (ERC20 - 0xe91D153E0b41518A2Ce8Dd3D7944Fa863463a97d) and sDAI (ERC4626 - 0xaf204776c7245bF4147c2612BF6e5972Ee483701). Since ERC4626 extends ERC20 and does not override methods we use (e.g. `approve`, `balanceOf`, etc), I would expect us to not have to differentiate between sDAI and xDAI for example.
If I understand the changes you did correctly, you are doing things on an `asset` (e.g. xDAI) level, which I imagined you don't want - this is the underlying xDAI that is accruing interest.

gabrielfior (changes requested):
Please see https://github.com/gnosis/prediction-market-agent-tooling/pull/320/files#r1688301684

prediction_market_agent_tooling/abis/erc20.abi.json
gabrielfior, line 1: wxDAI does not follow ERC20 abi, please rename.
wxdai -> https://gnosis.blockscout.com/token/0xe91D153E0b41518A2Ce8Dd3D7944Fa863463a97d?tab=contract_code
erc20 -> https://gist.github.com/veox/8800debbf56e24718f9f483e1e40c35c

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -684,6 +696,7 @@ def omen_create_market_tx(
     outcomes: list[str],
     auto_deposit: bool,
     fee: float = OMEN_DEFAULT_MARKET_FEE,
+    collateral_token_address: ChecksumAddress = WrappedxDaiContract().address,
```
gabrielfior, line 699: I think the entire logic (with a few if-else) can be simplified by using the adapter pattern ([link to one naive implementation of the pattern](https://refactoring.guru/design-patterns/adapter/python/example)).
By that, I mean that, instead of passing the `address` of the collateral, one can pass a class (e.g. `Collateral`), which is an ABC Metaclass. This should contain these methods
```
- [ ]  deposit
- [ ]  approve
- [ ]  balanceOf
- [ ]  convertToShares
```

Then, we need 2 classes (ERC4626, which can be sDAI) and wxDAI (not erc20 compatible, as discussed). For the ERC4626 case, methods need only to be delegated 1x1 to the underlying contract, whereas for wxDAI a bit more logic is needed (e.g. convertToAssets and convertToShares is 1x1, e.g. no fancy exchange rate). In other words, wxDAI is one example of ERC4626 with overriden signatures.

@kongzii please let me know if this implementation makes sense to you.

prediction_market_agent_tooling/markets/omen/omen.py
gabrielfior, line 710: This can be avoided by https://github.com/gnosis/prediction-market-agent-tooling/pull/320/files#r1688515591

prediction_market_agent_tooling/markets/omen/omen.py
gabrielfior: This complex logic can be avoided by https://github.com/gnosis/prediction-market-agent-tooling/pull/320/files#r1688515591

prediction_market_agent_tooling/markets/omen/omen.py
kongzii: I think that would just move the complexity to a different place, which would clean this function at least, however

>  instead of passing the address of the collateral, one can pass a class (e.g. Collateral)

We could do this here, but we can't do it in the other methods (buy, sell, fund), in them, we have only the address, so we need the logic to automatically convert address to a proper class:

https://github.com/gnosis/prediction-market-agent-tooling/blob/9e09ce38e35eb1b4e94af3feb09f1f0504724eb3/prediction_market_agent_tooling/markets/omen/omen.py#L626

and if we have the logic, we can use it here as well

> wxDAI (not erc20 compatible, as discussed)

wxDAI is indeed erc20 compatible, it's like a subclass of erc20 with additional deposit and withdraw methods. Seems like all WrappedX tokens are doing this.

> we can easily swap in other implementations, e.g. ERC20 (which most likely will also come)

ERC-20 should work now, but the complexity of if-else remains because we can not auto_deposit into ERC20.

>  ERC1155 (Circles' current implementation)

That I don't know how we will do. I think there will be many complications, so I'd think about that once it comes. (for example, what if better A and better B don't trust each other)

---

TLDR is that instead of simplifying, I added even more complexity 🤦  😄 🤯

gabrielfior: > I think that would just move the complexity to a different place, which would clean this function at least, however

I kindly disagree. With the adapter pattern, this whole `if-elif-else` logic disappears, because it does not matter which class you have, all methods are properly implemented (since they implement an interface/protocol).

> wxDAI is indeed erc20 compatible, it's like a subclass of erc20 with additional deposit and withdraw methods. Seems like all WrappedX tokens are doing this.

It's an extension of ERC20, just like ERC4626. The methods are indeed compatible (since it is an extension).

> circles
let's forget about this during this PR.

> ERC-20 should work now, but the complexity of if-else remains because we can not auto_deposit into ERC20.
The complexity goes away if adapter pattern is used.
I don't follow why `auto_deposit` is an issue - maybe we can have a function `auto_deposit` on each of the adapter classes?

kongzii: Agreed that `omen_create_market_tx` doesn't have to be bloated with this logic, so I cut it out. And did a small refactor again. But if you have another idea of how to implement this thing, maybe you could open a PR,  either against this one or after this is merged against the main branch?

prediction_market_agent_tooling/tools/contract.py
gabrielfior: Yes, there is - simply call
contract = web3_instance.eth.Contract(address, abi)
whatever_function = contract.get_function_by_name('deposit')
whatever_function.abi -> # dict{'inputs', 'outputs'} -> so you can check if this adheres to your interface

-> Still a bit clunky but better than calling the function and wrapping that in try-except

kongzii: > Still a bit clunky but better than calling the function and wrapping that in try-except

That was old code, doing it differently, check that out. Works without calling the function and also without the need for ABI.

prediction_market_agent_tooling/tools/contract.py
```diff
@@ -236,7 +346,190 @@ class ContractOnGnosisChain(ContractBaseClass):
...
+    return implements
+
+
+def init_erc4626_or_wrappererc20_or_erc20_contract(
+    address: ChecksumAddress,
+    web3: Web3,
+) -> (
+    ContractERC20BaseClass
+    | ContractERC4626BaseClass
+    | ContractDepositableWrapperERC20BaseClass
```
gabrielfior, line 412: Can we extract a common interface out of all these functions? This would avoid a lot of complexity downstream.

prediction_market_agent_tooling/tools/contract.py
kongzii: I will update the comment, forgot about that.

gabrielfior (approved):

PR #325, opened 2024-07-29
Refactoring ERC-4626 and support at other places

kongzii:


prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -549,13 +549,9 @@ def omen_buy_outcome_tx(
     Bets the given amount of xDai for the given outcome in the given market.
     """
     amount_wei = xdai_to_wei(amount)
-    from_address_checksummed = api_keys.bet_from_address
 
     market_contract: OmenFixedProductMarketMakerContract = market.get_contract()
     collateral_token_contract = market_contract.get_collateral_token_contract()
-    assert isinstance(
```
kongzii, line 556: Because FPMM uses IERC20 as the standard for collateral tokens, no changes required to support it here in the end. Just removing the assert and using the common auto deposit function.

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -669,10 +658,17 @@ def omen_sell_outcome_tx(
         max_outcome_tokens_to_sell,
         web3=web3,
     )
-    if auto_withdraw:
-        # Optionally, withdraw from the collateral token back to the `from_address` wallet.
-        collateral_token_contract.withdraw(
-            api_keys=api_keys, amount_wei=amount_wei, web3=web3
+    if auto_withdraw_erc4626 and isinstance(
+        collateral_token_contract, ContractERC4626BaseClass
```
kongzii, line 662: As talked on Slack, changed to withdrawing only 4626.

I tried to abstract away the `isinstance()` thing, unfortunately, mypy doesn't understand it then, so we can't have something like `if collateral_token_contract.is_erc4626`.

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -772,19 +767,15 @@ def omen_create_market_tx(
             web3=web3,
         )
 
-    # Use shares as the initial funds, if the collateral is erc-4626.
-    # We need to do this, because for example for 1 xDai in erc-20 token, we could receive <1 shares in the vault,
-    # and then, providing liquidity would fail, because we would not have enough shares.
-    initial_funds_in_asset_or_shares = wei_type(
-        asset_or_shares(collateral_token_contract, initial_funds_wei)
-        * 0.999  # Allow some slippage.
+    initial_funds_in_shares = collateral_token_contract.get_in_shares(
```
kongzii, line 770: Agree that `get_in_shares` can be implemented on ERC20BaseClass itself and then overridden in 4626 to keep things simple here.

prediction_market_agent_tooling/markets/omen/omen_contracts.py
```diff
@@ -390,16 +390,10 @@ def totalSupply(self, web3: Web3 | None = None) -> Wei:
 
     def get_collateral_token_contract(
         self, web3: Web3 | None = None
-    ) -> (
-        ContractERC20OnGnosisChain
-        | ContractERC4626OnGnosisChain
-        | ContractDepositableWrapperERC20OnGnosisChain
-    ):
+    ) -> ContractERC20OnGnosisChain:
```
kongzii, line 393: `Contract1 | Contract2 | Contract3` simplified to just `-> ContractERC20{BaseClass,OnGnosisChain}` everywhere.

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -338,9 +341,12 @@ def get_omen_binary_markets(
         sort_by_field: FieldPath | None = None,
         sort_direction: str | None = None,
         outcomes: list[str] = [OMEN_TRUE_OUTCOME, OMEN_FALSE_OUTCOME],
-        collateral_token_address: (
-            ChecksumAddress | None
-        ) = WrappedxDaiContract().address,  # TODO: Remove this default limitation once we fully support other than wxDai markets.
+        # TODO: Agents don't know how to convert value between other tokens, we assume 1 unit = 1xDai = $1 (for example if market would be in wETH, betting 1 unit of GNO would be crazy :D)
```
kongzii, line 344: We still need to have this default filter because of this, but otherwise, we support the manipulation of arbitrary collateral tokens.

prediction_market_agent_tooling/tools/contract.py
```diff
@@ -179,6 +181,14 @@ class ContractERC20BaseClass(ContractBaseClass):
         )
     )
 
+    def symbol(self, web3: Web3 | None = None) -> str:
+        symbol: str = self.call("symbol", web3=web3)
+        return symbol
+
+    @cache
+    def symbol_cached(self, web3: Web3 | None) -> str:
```
kongzii, line 189: When people don't care about typing or they are using common ERC20 methods, they can do just something like

```
if collateral_contract.symbol_cached() == "WXDAI":
    something something
```

prediction_market_agent_tooling/tools/contract.py
```diff
@@ -283,10 +297,11 @@ def deposit(
         self,
         api_keys: APIKeys,
         amount_wei: Wei,
-        receiver: ChecksumAddress,
+        receiver: ChecksumAddress | None = None,
         tx_params: t.Optional[TxParams] = None,
         web3: Web3 | None = None,
     ) -> TxReceipt:
+        receiver = receiver or api_keys.bet_from_address
```
kongzii, line 304: Did similar thing on `withdraw` as well. I think it's fine to default to the caller's public key, at least I'd expect that.

So now, if something supports both DepositableERC20 and ERC4626, the code can be just 

```
token.withdraw(api_keys, some_amount)
```

without any iffing.

prediction_market_agent_tooling/tools/contract.py
```diff
@@ -359,14 +395,16 @@ class ContractERC20OnGnosisChain(ContractERC20BaseClass, ContractOnGnosisChain):
 
 
 class ContractDepositableWrapperERC20OnGnosisChain(
-    ContractDepositableWrapperERC20BaseClass, ContractOnGnosisChain
+    ContractDepositableWrapperERC20BaseClass, ContractERC20OnGnosisChain
```
kongzii, line 398: Some changes were possible thanks to this shifting in class definition.

prediction_market_agent_tooling/tools/contract.py
```diff
@@ -435,95 +468,121 @@ def init_erc4626_or_wrappererc20_or_erc20_contract(
...
 
 
-def asset_or_shares(
-    collateral_token_contract: (
-        ContractERC20BaseClass
-        | ContractERC4626BaseClass
-        | ContractDepositableWrapperERC20BaseClass
-    ),
+def auto_deposit_depositable_wrapper_erc20(
+    collateral_token_contract: ContractDepositableWrapperERC20BaseClass,
```
kongzii, line 522: Hopefully this doesn't look so complex anymore, we have 

- auto_deposit_collateral_token
- auto_deposit_depositable_wrapper_erc20
- auto_deposit_depositable_wrapper_erc4626

if someone is working with one or the other contract, no iffing required, just use the appropriate function, if someone doesn't know what exactly he is working with (for example, because market can use any of them and so `market_contract.get_collateral_token_contract()` will return the right one), he can use the auto_deposit_collateral_token function that will do what's needed.

prediction_market_agent_tooling/tools/contract.py
```diff
@@ -435,95 +468,121 @@ def init_erc4626_or_wrappererc20_or_erc20_contract(
...
-        | ContractERC4626BaseClass
-        | ContractERC20BaseClass
-    ),
-) -> (
-    ContractDepositableWrapperERC20OnGnosisChain
-    | ContractERC4626OnGnosisChain
-    | ContractERC20OnGnosisChain
-):
+    contract: ContractERC20BaseClass,
+) -> ContractERC20OnGnosisChain:
```
kongzii, line 585: Simplified here as well.

prediction_market_agent_tooling/tools/contract.py
```diff
@@ -435,95 +470,121 @@ def init_erc4626_or_wrappererc20_or_erc20_contract(
 
     else:
         raise ValueError(
-            f"Contract at {address} on Gnosis Chain is neither WrapperERC-20, ERC-20 nor ERC-4626."
+            f"Contract at {address} is neither Depositable ERC-20, ERC-20 nor ERC-4626."
         )
 
 
-def auto_deposit_collateral_token(
+def auto_withdraw_collateral_token(
```
gabrielfior, line 477: The function `auto_withdraw_collateral_token` is called 1 time and simply calls `collateral_token.withdraw`. I think it can be removed.

prediction_market_agent_tooling/markets/omen/omen.py
gabrielfior: One question - in the case where the collateral_token is NOT erc4626 (e.g. wxDAI), is the `auto_withdraw` functionality still available?
I'm asking this because wxDAI also has a `withdraw` function albeit with a different signature.

prediction_market_agent_tooling/tools/contract.py
kongzii: Makes sense, removed.

prediction_market_agent_tooling/markets/omen/omen.py
kongzii: It has, and they can have the same signature thanks to the default arguments in the erc4626 method. Now removed auto_withdraw function was defined as 

```
def auto_withdraw_collateral_token(
    collateral_token_contract: (
        ContractERC4626BaseClass | ContractDepositableWrapperERC20BaseClass
    ),
    assets_wei: Wei,
    api_keys: APIKeys,
    web3: Web3 | None,
) -> None:
    collateral_token_contract.withdraw(
        api_keys,
        assets_wei,
        web3=web3,
    )
```

and `wxDai` is `ContractDepositableWrapperERC20BaseClass`

kongzii: However, we talked on slack about withdrawing only erc4626

kongzii: I added `auto_withdraw_depositable_erc20` flag

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -669,10 +661,22 @@ def omen_sell_outcome_tx(
...
-        # Optionally, withdraw from the collateral token back to the `from_address` wallet.
+    if (
+        auto_withdraw_erc4626
+        and isinstance(collateral_token_contract, ContractERC4626BaseClass)
+    ) or (
+        auto_withdraw_depositable_erc20
+        and isinstance(
+            collateral_token_contract, ContractDepositableWrapperERC20BaseClass
+        )
+    ):
```
gabrielfior, line 672: We already settled that these checks `isinstance(clz, ERC4626)` and `isistance(clz, ERC20Depositable)` will be the standard.
I would suggest, nevertheless, to pack these checks as methods, so that it becomes centralized in one place.
I see `isinstance` checks being called in multiple places in this class.

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -616,7 +610,8 @@ def omen_sell_outcome_tx(
     amount: xDai,  # The xDai value of shares to sell.
     market: OmenAgentMarket,
     outcome: str,
-    auto_withdraw: bool,
+    auto_withdraw_erc4626: bool,
+    auto_withdraw_depositable_erc20: bool,
```
gabrielfior, line 614: Can we revert this, and come back to the original option? I don't think it's reasonable to expect the user of the library to differentiate between both options.

prediction_market_agent_tooling/markets/omen/omen.py
gabrielfior: This `get_in_shares` is a great idea and simplifies things.

prediction_market_agent_tooling/markets/omen/omen.py
kongzii: I'm lost, that would go against what you proposed before (and it made sense to me).

Until now, we had auto_withdraw=False --> keeping funds in wxDai, as that's where agents need them anyway, no point in withdrawing to xDai and then depositing again to wxDai on the next run.

And you said erc 4626 (sdai) should be withdraw to erc 20 (wxdai), so we need `auto_withdraw_erc4626=True`, but `auto_withdraw_depositable_erc20=False`.

prediction_market_agent_tooling/markets/omen/omen.py
kongzii: Unfortunately I can't do that (even though I'd like it 😄 ), see https://github.com/gnosis/prediction-market-agent-tooling/pull/325#discussion_r1695143182

prediction_market_agent_tooling/markets/omen/omen_contracts.py
gabrielfior: Excellent

gabrielfior (approved):

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -669,10 +661,22 @@ def omen_sell_outcome_tx(
...
+        )
+    ):
         collateral_token_contract.withdraw(
-            api_keys=api_keys, amount_wei=amount_wei, web3=web3
+            api_keys,
+            remove_fraction(
+                amount_wei,
+                0.001,  # Allow 0.1% slippage.
+            ),
+            web3,
```
gabrielfior, line 679: Do we need the 2 `isinstance` above, even though the signature `collateral_token_contract.withdraw` is the same (which is great)?

In my head, we only need
```
if (
        auto_withdraw_erc4626
    ) or (
        auto_withdraw_depositable_erc20
```
hence only 1 auto_withdraw argument (at least for this function) would be needed.

image

prediction_market_agent_tooling/markets/omen/omen.py
gabrielfior: Ow, now I see - you mean that `auto_withdraw_depositable_erc20` means wxdai -> xdai, and `auto_withdraw_erc4626` means sdai -> wxdai?
If that's the case, my suggestion would be to have only 1 argument (e.g. `auto_withdraw`), and 1 round of unwrapping is made. In the cases above, we would have sdai -> wxdai OR wxdai -> xDAI, but not 2 rounds (e.g. sDAI -> wxDAI -> xDAI).
Does that make sense?

kongzii: There aren't two rounds. Possible scenarios are:

1. The collateral token is ERC20; no withdrawal is possible, and that's fine.
2. The collateral token is Wrapper ERC20; The withdraw method is available, and if called, it will withdraw wxDai to xDai.
3. Collateral token is ERC4626, withdraw method is available, but it withdraws into ERC20 (so sDai --> wxDai)

So if we have only a single `auto_withdraw` and we set it to True:

1. Nothing happens
2. wxDai will be withdrawn to xDai (that's not wanted)
3. sDai will be withdrawn to wxDai (I don't care whether we do this or not)

---

Initially, I proposed to have a single `auto_withdraw` argument and keep it to `False`, so I'm fine with that if you changed your opinion on it as well.

kongzii: For future reference, explained over Slack that such behaviour is not possible and agreed that single auto_withdraw, but set to `False` is the way to go for now

PR #330, opened 2024-07-30
Add simplified get_kelly_bet function, remove old

evangriffiths:


prediction_market_agent_tooling/tools/betting_strategies/kelly_criterion.py
```diff
@@ -11,6 +11,58 @@
...
+    confidence: float,
+    bet_outcome: BetOutcome,
+) -> float:
+    """
+    Calculate the optimal bet amount using the Kelly Criterion for a binary outcome market.
+
+    Based on https://en.wikipedia.org/wiki/Kelly_criterion
+
+    'Simplified' in that it is assumes that the bet does not change the market odds.
+    """
```
gabrielfior, line 40: Could you please paste the required formula here? Like `p - q/b` I believe

prediction_market_agent_tooling/tools/betting_strategies/kelly_criterion.py
```diff
@@ -11,6 +11,58 @@
...
+
+    # Handle the case where market_prob is 0
+    if market_prob == 0:
+        market_prob = 1e-10
+
+    odds = (1 / market_prob) - 1
+    kelly_fraction = edge / odds
+
+    # Ensure bet size is non-negative does not exceed the wallet balance
+    bet_size = max(min(kelly_fraction * max_bet, max_bet), 0)
```
gabrielfior, line 62: From my reading of the Wikipedia article,  I was under the impression it is possible that bet size is negative, hence we should bet on the opposite direction.
Is that also possible/desirable here?

tests/markets/test_betting_strategies.py
```diff
@@ -216,3 +218,42 @@ def test_stretch_bet_between(
...
+    assert (
+        get_kelly_bet_simplified(
+            market_p_yes=omen_market.current_p_yes,
+            estimated_p_yes=est_p_yes,
+            max_bet=10,
+            confidence=1.0,
+            bet_outcome=bet_outcome,
+        )
+        > 0
+    )
```
gabrielfior, line 248: I thought the formula presented a closed form, thus the test could simply assert the formula was calculated properly for a few conditions, or is it not the case here?

prediction_market_agent_tooling/tools/betting_strategies/kelly_criterion.py
evangriffiths: Yes, that's one option. I figured there are 3 ways of writing this:
1. user passes in p_yes, market_p_yes. Returns a tuple of [bet_direction, bet amount]
2. user passes in p_yes, market_p_yes. Returns signed 'yes' bet_amount (i.e. -ve amount means 'no' direction)
3. user passes in p_yes, market_p_yes, intended bet_direction. Returns bet_amount if > 0, else 0

I went with (3). Do you prefer one of the others (or something else)?

tests/markets/test_betting_strategies.py
evangriffiths: > assert the formula was calculated properly

How would that look like, without duplicating the code for the calculation in the test?

prediction_market_agent_tooling/tools/betting_strategies/kelly_criterion.py
gabrielfior: I like (1) better because you can simply pass the resulting `bet_direction` and `bet_amount` into whatever function is placing bets. Not a strong opinion on this though - if the ability to "switch" bet outcomes is covered here, then I'm happy.

prediction_market_agent_tooling/tools/betting_strategies/kelly_criterion.py
```diff
@@ -20,6 +80,7 @@ def _get_kelly_criterion_bet(
 
     Taken from https://github.com/valory-xyz/trader/blob/main/strategies/kelly_criterion/kelly_criterion.py
 
+    Takes into account how the bet changes the market odds.
```
gabrielfior, line 83: Can we remove `_get_kelly_criterion_bet`? It's not used anywhere and your approach is well integrated with the whole market classes we have in order to determine a bet size.

prediction_market_agent_tooling/tools/betting_strategies/kelly_criterion.py
```diff
@@ -11,6 +11,66 @@
...
+from enum import Enum
+
+
+class BetOutcome(str, Enum):
+    YES = "Yes"
+    NO = "No"
+
+
+def check_is_valid_probability(probability: float) -> None:
+    if probability < 0 or probability > 1:
```
kongzii, line 23:
```suggestion
    if not 0 <= probability <= 1:
```

little python magic showcase, no need to commit 😄

prediction_market_agent_tooling/tools/betting_strategies/kelly_criterion.py
```diff
@@ -11,6 +11,66 @@
...
+
+    f* = p - q / b
+
+    where:
+    - f* is the fraction of the current bankroll to wager
+    - p is the probability of a win
+    - q = 1-p is the probability of a loss
+    - b is the proportion of the bet gained with a win
+
+    'Simplified' in that it is assumes that the bet does not change the market odds.
```
kongzii, line 47: Are there any significant differences between using simplified and using the original? (in terms of final bet size)

kongzii (approved):

prediction_market_agent_tooling/tools/betting_strategies/kelly_criterion.py
evangriffiths: Screenshot 2024-07-31 at 11 15 58

They're close when the max bet value is <= to market volume, but they diverge after that.

evangriffiths: (https://github.com/gnosis/prediction-market-agent-tooling/pull/331 to reproduce)

prediction_market_agent_tooling/tools/betting_strategies/kelly_criterion.py
evangriffiths: Sure, done

evangriffiths:
# How does the bet size change with 'market p_yes'?

```
import numpy as np
from prediction_market_agent_tooling.tools.betting_strategies.kelly_criterion import (
    BetDirection,
    KellyBet,
    get_kelly_bet,
)

xdai_amounts_simplified = []
market_p_yess = np.linspace(0.01, 0.99, 100)
max_bet = 2
est_p_yes = 0.3
market_p_yes = 0.65
for market_p_yes in market_p_yess:
    kelly_bet: KellyBet = get_kelly_bet(
        market_p_yes=market_p_yes,
        estimated_p_yes=est_p_yes,
        max_bet=max_bet,
        confidence=1,
    )
    kelly_bet_signed = (
        kelly_bet.size if kelly_bet.direction == BetDirection.YES else -kelly_bet.size
    )
    # kelly_bet_signed = kelly_bet.size
    xdai_amounts_simplified.append(kelly_bet_signed)

# Plot the results
from matplotlib import pyplot as plt

plt.plot(
    market_p_yess,
    xdai_amounts_simplified,
)
plt.xlabel("Market p_yes")
plt.ylabel("Bet Amount")
plt.title(f"Kelly Bet ({max_bet=}, est_p_yes={est_p_yes:.3g})")
plt.legend()
plt.show()
```
Screenshot 2024-07-31 at 12 21 41

evangriffiths:
# How does the bet size change with 'estimated p_yes'?

```
import numpy as np
from prediction_market_agent_tooling.tools.betting_strategies.kelly_criterion import (
    BetDirection,
    KellyBet,
    get_kelly_bet,
)

bet_amounts = []
est_p_yess = np.linspace(0.01, 0.99, 100)
max_bet = 2
market_p_yes = 0.65
for est_p_yes in est_p_yess:
    kelly_bet: KellyBet = get_kelly_bet(
        market_p_yes=market_p_yes,
        estimated_p_yes=est_p_yes,
        max_bet=max_bet,
        confidence=1,
    )
    kelly_bet_signed = (
        kelly_bet.size if kelly_bet.direction == BetDirection.YES else -kelly_bet.size
    )
    bet_amounts.append(kelly_bet_signed)

# Plot the results
from matplotlib import pyplot as plt

plt.plot(est_p_yess, bet_amounts)
plt.xlabel("Estimated p_yes")
plt.ylabel("Bet Amount")
plt.title(f"Kelly Bet ({max_bet=}, {market_p_yes=})")
plt.show()
```

Screenshot 2024-07-31 at 12 28 16

prediction_market_agent_tooling/tools/betting_strategies/kelly_criterion.py
evangriffiths: Fair enough, have changed to (1)

evangriffiths:
pytest failure unrelated

PR #337, opened 2024-08-06
Store Tavily results

kongzii:


prediction_market_agent_tooling/tools/tavily_cached/tavily_cached.py
```diff
@@ -0,0 +1,97 @@
...
+        include_answer=include_answer,
+        include_raw_content=include_raw_content,
+        include_images=include_images,
+        use_cache=use_cache,
+        api_keys=api_keys,
+    )
+    response_parsed = TavilyResponse.model_validate(response)
+    if response_cache:
+        response_cache.save(
+            query=query,
```
gabrielfior, line 48: When I read `cached` I was thinking of having some sort of `cache-hit-or-miss` mechanism, but in this case I believe we are simply saving the Tavily results to a local DB, correct?
So instead of caching, we could use `TavilyMemory` or `TavilyDBStorage`, but not cache - what do you think?
Please correct me if my understanding of the use case is wrong.

prediction_market_agent_tooling/tools/tavily_cached/tavily_cached.py
```diff
@@ -0,0 +1,97 @@
...
+        query=query,
+        search_depth=search_depth,
+        topic=topic,
+        max_results=max_results,
+        include_domains=include_domains,
+        exclude_domains=exclude_domains,
+        include_answer=include_answer,
+        include_raw_content=include_raw_content,
+        include_images=include_images,
+        use_cache=use_cache,
```
evangriffiths, line 95: Tavily has its own cache!! Did you try it out and see if it saves credits? From the docs (which don't have much detail) this is default-True, so I guess we are using it already.

prediction_market_agent_tooling/tools/tavily_cached/tavily_cached.py
evangriffiths: Also a bit confused by this. I thought the original motivation was to save credits, but maybe I'm mis-remembering. Can you remind me what was the motivation if not that?

evangriffiths (commented):
Not sure how much of a spanner in the works this is... but when I was googling tavily's `use_cache` arg, this result came up https://tavily.com/policy, which under 'usage restrictions' says:

> you must not, and must not allow others to copy, store, archive, cache, or create a database of the Content, in part or in entirety;

😨

prediction_market_agent_tooling/tools/tavily_cached/tavily_cached.py
kongzii: Yeah I will rename it.

The motivation was to just store it for a few days and then check it out if some kind of cache makes sense to do.

prediction_market_agent_tooling/tools/tavily_cached/tavily_cached.py
kongzii: It's default to True, but it doesn't save credits (or at least I didn't notice it saving credits if I did multiple identical request in a row)

evangriffiths:
> Well, this sucks 😄 It doesn't make sense to me to use Tavily at all if we aren't allowed to save the results. We need them not just to save costs but also for later debugging and fine-tuning.

I was thinking for finetuning we would want to save:
- 'research report'
- prediction/'answer' from the agent
- market resolution

How were you thinking we would use the raw tavily results?

kongzii:
I'm thinking the same, but we need all the data that contribute to the prediction

- Returned URLs to check if the website scraping is working correctly (it happened before that scraped content was bugged, anything that follows afterward in our workflow is broken if it happens)
- Scraped content to check if the research report is generated correctly (LLM can hallucinate or miss something important)
- Scraped content with timestamp (as we have in TavilyStorage currently), to use the same data in the future

And the following points 

> Employ the Content in any machine learning or algorithmic activities;

Can we even use the raw content (without storing it) for our agents?

> Use the Content to create, refine, assess, or enhance any services offered by you or third parties; or

Sounds like LabsAPI (I mean, aiAPI!) needs to use something else anyway

---

But I don't understand this because: https://docs.tavily.com/docs/tavily-api/langchain. They are officially partnering with ML-related library and I would expect that LLM inputs and outputs are generally stored to be debugged later on 🤷

evangriffiths:
Good points. Makes sense. Yeah I agree it seems an odd thing to have in the terms of service

evangriffiths (approved):
Have made a follow-up task to look into loading the saved results instead of always doing a new search: https://github.com/gnosis/prediction-market-agent/issues/388

evangriffiths:
Will just mention here (otherwise after [this](https://github.com/gnosis/prediction-market-agent-tooling/pull/340#pullrequestreview-2226952585) comment it would be double-standards) that `TavilyStorage` doesn't really belong in PMAT - it's more of a PMA feature imo

kongzii:
For future reference, we contacted Tavily, and they are okay with it. They are also going to update their terms, as the current ones really don't fit LLM Agent use-cases.

PR #340, opened 2024-08-07
Gabriel/trade on cowswap

gabrielfior:
Added integration with CowSwap for future token swapping


evangriffiths (commented):
I'm not saying we should manage extra repos for the sake of it, but does `CowClient` really belong in PMAT?

Arguments against:

- PMAT seems quite clearly specialized in prediction markets. The fact that we support manifold/metaculus means it's not just a 'web3 tools/apis for things on gnosis chain that our agents use' repo.
- A simple python client is a nice stand-alone package that other people may find useful (and cowswap devs might help maintain!)

Arguments for:

- More repos == annoying.
- The repo is really for 'anything useful for our agents'. We should rename it to make that clear (e.g. gnosis-ai-agent-tooling)

Thoughts?

prediction_market_agent_tooling/tools/cowswap/cow_client.py
```diff
@@ -0,0 +1,140 @@
...
+    def post_quote(self, quote: QuoteInput) -> QuoteOutput:
+        quote_dict = quote.dict(by_alias=True, exclude_none=True)
+        r = requests.post(f"{self.api_url.value}/api/v1/quote", json=quote_dict)
+
+        self._if_error_log_and_raise(r)
+        return QuoteOutput.model_validate(r.json()["quote"])
+
+    @staticmethod
+    def build_order_with_fee_and_sell_amounts(quote: QuoteOutput) -> dict[str, Any]:
+        quote_dict = quote.dict(by_alias=True, exclude_none=True)
```
kongzii, line 71: `dict` is deprecated, we should get used to using the replacement, otherwise this will error out sooner or later: https://github.com/pydantic/pydantic/blob/61f027cab5004823d1e000a03e5c20eabcfba733/pydantic/main.py#L1074

prediction_market_agent_tooling/tools/cowswap/cow_client.py
```diff
@@ -0,0 +1,140 @@
...
+        new_sell_amount = int(quote.sell_amount) + int(quote.fee_amount)
+        order_data = {
```
kongzii, line 73: Couldn't these dynamic dictionaries be Pydantic models as well? Something like

```python
class OrderData(QuoteInput):
    sellAmount: str
    feeAount: str
```

prediction_market_agent_tooling/tools/cowswap/encoding.py
```diff
@@ -0,0 +1,41 @@
...
+        {"name": "sellTokenBalance", "type": "string"},
+        {"name": "buyTokenBalance", "type": "string"},
+    ]
+}
+
+# EIP-712 Domain
+DOMAIN = {
+    "name": "Gnosis Protocol",
+    "version": "v2",
+    "chainId": 100,  # Replace with actual chainId
```
kongzii, line 34: We have https://github.com/gnosis/prediction-market-agent-tooling/blob/6055ed2ef7f5e35378a16ba8ca167ddba7239f80/prediction_market_agent_tooling/tools/gnosis_rpc.py#L5, can be used here

```suggestion
    "chainId": GNOSIS_NETWORK_ID,  # Replace with actual chainId
```

kongzii (commented):
Just small suggestions, otherwise looks neat.

Agree with Evan, are Cow guys interested in having an official Python client? They could help us maintain it (or at least ping us if some changes are required).

We can motivate them by saying that if AI world is coming to web3, having Python client would be a huge advantage over others.

If they aren't interested, then fine by me to keep it in PMAT.

gabrielfior:
This will be replaced by an external library, see https://github.com/cowdao-grants/python-client/pull/3

gabrielfior:
This will be replaced by an external library, see https://github.com/cowdao-grants/python-client/pull/3.
Thanks for the review, the remarks will be addressed in the PR above.

prediction_market_agent_tooling/tools/cowswap/cow_client.py
gabrielfior: Refactored the code so that no dynamic dict creation is needed. But thanks.

PR #345, opened 2024-08-13
Updating langfuse integration

kongzii:


prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -229,42 +328,55 @@ def check_min_required_balance_to_operate(self, market_type: MarketType) -> None
...
             )
 
-    def pick_markets(
-        self, market_type: MarketType, markets: t.Sequence[AgentMarket]
-    ) -> t.Sequence[AgentMarket]:
+    @observe()
+    def have_bet_on_market_since_observed(
+        self, market: AgentMarket, since: timedelta
+    ) -> bool:
+        return self.have_bet_on_market_since(market, since)
```
kongzii, line 335: This is the pattern I was talking about last time, I don't like it so much but I also don't see any clearer way to proceed.

This pattern is used on all observed methods of DeployableAgent.

The reason is, that if some method is `@observed()` and the agent class needs to implement that method (`answer_binary_market` is a clear example, but I think it will be usable for others as well), then the overridden method would not be `@observed()` anymore.

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -229,42 +328,55 @@ def check_min_required_balance_to_operate(self, market_type: MarketType) -> None
...
+
+    def have_bet_on_market_since(self, market: AgentMarket, since: timedelta) -> bool:
+        return have_bet_on_market_since(keys=APIKeys(), market=market, since=since)
+
+    @observe()
+    def verify_market_observed(
```
kongzii, line 341: I had to refactor the logic of these methods, before it was:

1. Fetch markets
2. Filter markets
3. Predict+Bet on them one by one

Now it is

1. Fetch markets
2. One by one check if the market is ok to predict and if so, predict + bet on it.

The reason is that we need the whole process end-to-end saved in Langfuse, this is how it looks like:

Screenshot by Dropbox Capture

prediction_market_agent_tooling/deploy/agent_example.py
```diff
@@ -11,10 +11,8 @@
 
 
 class DeployableCoinFlipAgent(DeployableTraderAgent):
-    def pick_markets(
-        self, market_type: MarketType, markets: t.Sequence[AgentMarket]
-    ) -> t.Sequence[AgentMarket]:
-        return random.sample(markets, 1)
+    def verify_market(self, market_type: MarketType, market: AgentMarket) -> bool:
+        return True
```
kongzii, line 15: After refactoring, it's no longer possible to randomly pick one of the markets, but hopefully, that's not a problem.

prediction_market_agent_tooling/markets/categorize.py
```diff
@@ -17,6 +26,10 @@ def infer_category(
 """
     )
 
+    config: RunnableConfig = {}
+    if enable_langfuse:
+        config["callbacks"] = [langfuse_context.get_current_langchain_handler()]
```
kongzii, line 31: This is again a pattern that I don't like so much, but it's required.

One would probably expect a solution as:

```python
def infer_category(
    ...,
    callbacks: list[...] | None = None,
):
```

And then simply use the callbacks in the invoke method.

However, when `langfuse_context.get_current_langchain_handler()` is called, it will pick up the latest observed function (see in the previous screenshot that observation can be nested), and because of that, we need to initialize this callback right before calling the invoke method, because otherwise, it could be linked to a different observation.

prediction_market_agent_tooling/tools/langfuse_.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def observe(
+    name: Optional[str] = None,
+    as_type: Optional[Literal["generation"]] = None,
+    capture_input: bool = True,
+    capture_output: bool = True,
+    transform_to_string: Optional[Callable[[Iterable[Any]], str]] = None,
+) -> Callable[[Callable[P, R]], Callable[P, R]]:
+    casted: Callable[[Callable[P, R]], Callable[P, R]] = original_observe(
```
kongzii, line 19: This is described at https://github.com/langfuse/langfuse-python/pull/862

prediction_market_agent_tooling/markets/categorize.py
kongzii: But I will think some more about this, if you don't have some clear refactor idea for this, I'd merge it and change in another PR if I think of something 🙏

kongzii: Issued at https://github.com/gnosis/prediction-market-agent-tooling/issues/346

evangriffiths: Could you at least turn this into a util function, like:
```
def get_langchain_runnable_config(enable_langfuse: bool) -> RunnableConfig:
    config: RunnableConfig = {}
    if enable_langfuse:
        config["callbacks"] = [langfuse_context.get_current_langchain_handler()]
    return config
```
?

evangriffiths: I guess that's what you propose as (1) in https://github.com/gnosis/prediction-market-agent-tooling/issues/346, except `existing_config` is always None in PMAT.

prediction_market_agent_tooling/markets/categorize.py
```diff
@@ -1,13 +1,22 @@
...
 ) -> str:
+    try:
+        from langchain.prompts import ChatPromptTemplate
+        from langchain.schema.output_parser import StrOutputParser
+        from langchain_core.runnables.config import RunnableConfig
+        from langchain_openai import ChatOpenAI
+    except ImportError:
+        raise ImportError(
+            "openai not installed, please install extras `langchain` to use this function."
+        )
```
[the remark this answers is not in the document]
evangriffiths: @coderabbitai congrats on your Series A funding round!

prediction_market_agent_tooling/tools/langfuse_.py
evangriffiths: Could we also override their decorator to avoid having to have this pattern

```
@observe
def foo_observed()
    return foo(enable_langfuse=True)
```

by doing something like:

```
from prediction_market_agent.utils import APIKeys

class APIKeysTest(APIKeys):
    FOO: str | None = None
    @property
    def default_enable_langfuse(self):
        return self.FOO is not None

def original_observe(func):
    def wrapper(*args, **kwargs):
        result = func(*args, **kwargs)
        return result + "foo"
    return wrapper

def observe(func):
    def noop_decorator(func):
        return func

    enable = APIKeysTest().default_enable_langfuse
    if enable:
        return original_observe(func)
    else:
        return noop_decorator(func)

@observe
def hello_world():
    return "Hello, world! "

print(hello_world())
```
i.e. you get
```
%  python test.py
Hello, world! 
%  FOO=1 python test.py
Hello, world! foo
```

The drawback to this is that now you can't do:
```
agent0 = DeployableFooAgent(enable_langfuse=True)
agent1 = DeployableFooAgent(enable_langfuse=False)
```

prediction_market_agent_tooling/deploy/agent.py
gabrielfior: I don't love this pattern either, this means every time you add a new method, you need to add an observed one (to avoid people overriding the observed one and only overriding the "pure" one).

One solution I came across is adding a class decorator to `DeployableAgent`, which "observes" the a set of functions without annotating them. Hence you could avoid having to manually decorate them, instead only decorating the base class (see snippet below).
What this would imply:
-> Annotate `DeployableAgent` with decorator, define that methods `[have_bet_on_market, ...]` should be observed. Add logic to observe function (e.g. call `decorator(my_function)` or similar)
-> If subclass (e.g. CoinflipDeployableAgent) overwrites the method, it doesn't matter, it will keep getting observed (due to class decorator being in the base class).
-> This is nice because you avoid having duplicate methods everywhere.

Let me know what you think @kongzii - Gist [here](https://gist.github.com/gabrielfior/02325d1ea6a2d31440cc97bba195cd4a).

prediction_market_agent_tooling/config.py
```diff
@@ -134,6 +135,14 @@ def langfuse_host(self) -> str:
             self.LANGFUSE_HOST, "LANGFUSE_HOST missing in the environment."
         )
 
+    @property
+    def default_enable_langfuse(self) -> bool:
```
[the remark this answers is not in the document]
kongzii: That's not necessary for langfuse to work, it can be `None`

prediction_market_agent_tooling/deploy/agent.py
kongzii: That could work, but based on the gist, it would decorate all the methods, which is not needed (and maybe including methods as `__init__` and similar?)

So we would need something like

```
@prepend_foo_to_methods(allowed_methods=["have_bet_on_market", ...])
class MyClass:
```

which isn't type-safe and can be easily forgotten.

Another approach I was experimenting with was something like

```
def initialize_langfuse(self) -> None:
    ...
    self.have_bet_on_market = observe()(self.have_bet_on_market)  # type: ignore[method-assign]
```

It works similarly to your suggestion, but mypy should warn us if for example `have_bet_on_market` gets deleted or something (but I didn't test it), although it still needs `type: ignore` comment, because changing methods like this isn't allowed. 

I didn't use that approach because it needs `type: ignore`, but I don't feel strongly about it.

Do you have preferences?

gabrielfior: > @prepend_foo_to_methods(allowed_methods=["have_bet_on_market", ...])
class MyClass:

What I had in mind was:
-> Add the decorator to the base class (`DeployableTraderAgent`). Note that subclasses (e.g. `DeployableCoinFlipTraderAgent`) and similar do not get decorated. Hence we need to decorate only once (or maybe more, but only on base classes), so not very hard to remember.
-> The decorator does not modify all functions, only a subset (either hardcoded or even better, as you suggested it, provided as an argument).
-> Not sure how to satisfy mypy here, maybe using a FrozenSet but not sure

Regarding the other approach (overriding methods with `observe()(self.have_bet_on_market)  # type: ignore[method-assign]`, it's a similar approach to the decorator, but I like the decorator a bit better because I find it more elegant (no real reason).

The point I'm trying to make is: I don't love the duplicated-function pattern (observed + pure). If we can find any solution to this (be it decorator or non-decorator), I'm happy with any option.

prediction_market_agent_tooling/markets/categorize.py
evangriffiths: Any reason not to do that? Just to save lines I guess

kongzii: I don't think so, I can do it. I just also need to think about the suggestion below. But now I'm just responding in-between time when I don't know what to write into the paper 😄 

I'll look in more detail into all comments tomorrow 👌

prediction_market_agent_tooling/deploy/agent.py
kongzii: Personally, I don't care for duplicated `*_observer` method or doing some magic, but what I'm afraid of is that

```
@magic(["predict"])
class Agent:
    def predictt(self):
        print("Predicting")
```

will never trigger mypy as a mistake, but

```
from prediction_market_agent_tooling.tools.langfuse_ import observe

class Agent:
    def init_langfuse(self) -> None:
        self.predict = observe()(self.predict)

    def predictt(self) -> None:
        print("Predicting")
```

will. And it also works with the auto-refactoring features of VSCode.

In any case, we will get rid of _observed methods, I'll change it.

prediction_market_agent_tooling/tools/langfuse_.py
kongzii: I don't think it's a good idea to add additional logic into their decorator, because what can happen is that someone uses both PMAT and also observes some of his method manually (like we do in PMA), and then they could receive different behaviours on observed methods.

However, it seems like this is fine even without modifying the observer decorator:

Screenshot by Dropbox Capture

So we can just `@observe` everything by default. I thought it's a good idea to have both observed and non-observed functions, but now I don't feel strong about it.

And in combination with https://github.com/gnosis/prediction-market-agent-tooling/pull/345#discussion_r1719788051, code will be simpler to follow.

evangriffiths: > someone uses both PMAT and also observes some of his method manually (like we do in PMA), and then they could receive different behaviours on observed methods.

Sorry I don't understand this bit. 

> So we can just @observe everything by default.

But anyway, this sounds good

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -209,12 +281,50 @@ class DeployableTraderAgent(DeployableAgent):
...
+    ) -> None:
+        super().__init__(enable_langfuse=enable_langfuse)
         self.place_bet = place_bet
 
-    def have_bet_on_market_since(self, market: AgentMarket, since: timedelta) -> bool:
-        return have_bet_on_market_since(keys=APIKeys(), market=market, since=since)
+    def initialize_langfuse(self) -> None:
+        super().initialize_langfuse()
+        # Auto-observe all the methods where it makes sense, so that subclassses don't need to do it manually.
+        self.have_bet_on_market_since = observe()(self.have_bet_on_market_since)  # type: ignore[method-assign]
```
evangriffiths, line 295: Is there a reason you do `@observe()` in some places, but `observe()(self.some_class_method)` here?

kongzii: Yeah, if something is just a function:

```
@observe()
def research(...) -> ...
```

We can use the decorator, and no harm is done. Anyone who uses research will use the observed variant.

But in our agent:

```
class Agent:
   @observe()
   def xyz(...) -> <default strategy>

class ProphetAgent(Agent):
   def xyz(...) -> <some more fancy strategy> <<<--- This would remove the decorator, so all deployable agent would have to manually @observe their methods.
```

kongzii: This is related to https://github.com/gnosis/prediction-market-agent-tooling/pull/345#discussion_r1718263389

evangriffiths: Ah makes sense, thanks

evangriffiths (approved):

Issue #309, opened 2024-07-12
Fix local_chain test, timeout issues


gabrielfior:
Ref -> https://github.com/gnosis/prediction-market-agent-tooling/actions/runs/9905611350/job/27365498526?pr=302

Increasing timeout didn't help much - investigate issue, consider alternative solutions.

gabrielfior:
Errors arise due to RPC problems, see logs https://github.com/gnosis/prediction-market-agent-tooling/actions/runs/10469875908/job/28993818419?pr=350#step:8:40

Solution here is to use a different RPC (only for testing), not to overcomplicate things with `ape` which adds more dependencies/config files, etc.

gabrielfior:
Fixed by #357

Issue #338, opened 2024-08-06
Refactor DeployableTraderAgent to 1. predict a probability for a market then 2. place a bet based on its `BettingStrategy`


evangriffiths:
May need a refactor of `DeployableTraderAgent.process_bets` (see https://github.com/gnosis/prediction-market-agent-tooling/blob/69ec60e8206ac64bbabc8707d8b62d251d699e54/prediction_market_agent_tooling/deploy/agent.py#L281)

Currently it does:
```
result = self.answer_binary_market(market)
amount = self.calculate_bet_amount(result, market)
market.place_bet(amount=amount, outcome=result.decision)
```

The abstraction is a bit off here if we want to use `get_kelly_bet` [here](https://github.com/gnosis/prediction-market-agent-tooling/blob/69ec60e8206ac64bbabc8707d8b62d251d699e54/prediction_market_agent_tooling/tools/betting_strategies/kelly_criterion.py#L21), as we want to calculate the direction and amount together, so it would look like:

```
prediction = self.predict_binary_market(market)
bet = self.get_bet_from_prediction(market, prediction)
market.place_bet(amount=bet.amount, outcome=bet.direction)
```

evangriffiths:
Re-opening, as https://github.com/gnosis/prediction-market-agent-tooling/pull/352 didn't actually change the default

gabrielfior:
> Re-opening, as #352 didn't actually change the default

@evangriffiths do you mind closing this ticket in favor of https://github.com/gnosis/prediction-market-agent-tooling/issues/365?

PR #352, opened 2024-08-20
338 change default betting strategy from amount == tiny bet direction == yes if p yes 05 to use kelly bet

gabrielfior:


gabrielfior:
Closes https://github.com/gnosis/prediction-market-agent-tooling/issues/309

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -208,10 +183,15 @@ class DeployableTraderAgent(DeployableAgent):
...
+    strategy: BettingStrategy = FixedBetBettingStrategy()
 
-    def __init__(self, place_bet: bool = True) -> None:
+    def __init__(
+        self,
+        place_bet: bool = True,
+    ) -> None:
         super().__init__()
         self.place_bet = place_bet
+        check_not_none(self.strategy)
```
kongzii, line 194: Why would strategy be None?

```suggestion
```

prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -0,0 +1,61 @@
...
+    @abstractmethod
+    def calculate_bet_amount_and_direction(
+        self, answer: Answer, market: AgentMarket
+    ) -> TokenAmountAndDirection:
+        pass
+
+
+class FixedBetBettingStrategy(BettingStrategy, metaclass=ABCMeta):
+    def get_bet_amount_for_market(self, market: AgentMarket) -> float:
+        if isinstance(market, ManifoldAgentMarket):
```
kongzii, line 27: If we have market class here, why not just

```
return market.get_tiny_bet_amount()
```

as we had before?

prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -0,0 +1,61 @@
...
+
+class BettingStrategy(ABC):
+    @abstractmethod
+    def calculate_bet_amount_and_direction(
+        self, answer: Answer, market: AgentMarket
+    ) -> TokenAmountAndDirection:
+        pass
+
+
+class FixedBetBettingStrategy(BettingStrategy, metaclass=ABCMeta):
```
kongzii, line 25: This is no longer abstract class, or is it?
```suggestion
class FixedBetBettingStrategy(BettingStrategy):
```

prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -0,0 +1,61 @@
...
+    ) -> TokenAmountAndDirection:
+        bet_amount = self.get_bet_amount_for_market(market)
+        return TokenAmountAndDirection(
+            amount=bet_amount,
+            currency=market.currency,
+            direction=answer.decision,
+        )
+
+
+class KellyBettingStrategy(BettingStrategy, metaclass=ABCMeta):
```
kongzii, line 45: Same here
```suggestion
class KellyBettingStrategy(BettingStrategy):
```

prediction_market_agent_tooling/markets/data_models.py
```diff
@@ -57,6 +57,49 @@ def __str__(self) -> str:
         return f"Resolved bet for market {self.market_id} for question {self.market_question} created at {self.created_time}: {self.amount} on {self.outcome}. Bet was resolved at {self.resolved_time} and was {'correct' if self.is_correct else 'incorrect'}. Profit was {self.profit}"
 
 
+class KellyResult(BaseModel):
```
kongzii, line 60: Shouldn't KellyResult be defined in `prediction_market_agent_tooling/tools/betting_strategies/kelly_criterion.py`?

prediction_market_agent_tooling/tools/web3_utils.py
```diff
@@ -157,7 +157,7 @@ def _prepare_tx_params(
     # Retry only for the transaction errors that match the given patterns,
     # add other retrieable errors gradually to be safe.
     retry=tenacity.retry_if_exception_message(
-        match="(.*wrong transaction nonce.*)|(.*Invalid.*)|(.*OldNonce.*)"
+        match="(.*wrong transaction nonce.*)|(.*Invalid.*)|(.*OldNonce.*)|(AlreadyKnown.*)"
```
kongzii, line 160: When AlreadyKnown happens?

pyproject.toml
```diff
@@ -46,6 +46,8 @@ tavily-python = "^0.3.9"
 sqlmodel = "^0.0.21"
 psycopg2-binary = "^2.9.9"
 base58 = ">=1.0.2,<2.0"
+eth-ape = "^0.8.12"
```
kongzii, line 49: These aren't only dev dependencies anymore?

scripts/sell_all_omen.py
```diff
@@ -61,7 +62,10 @@ def sell_all(
         outcome = agent_market.outcomes[bet.outcomeIndex]
         current_token_balance = agent_market.get_token_balance(better_address, outcome)
 
-        if current_token_balance.amount <= OmenAgentMarket.get_tiny_bet_amount().amount:
+        minimum_token_amount_for_selling = BetAmount(
+            amount=0.00001, currency=OmenAgentMarket.currency
```
kongzii, line 66: Maybe we can keep the `.get_tiny_bet_amount` so that `0.0001` is defined on one place instead of two?

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -316,12 +302,17 @@ def process_bets(self, market_type: MarketType) -> None:
                 logger.info(f"Skipping market {market} as no answer was provided")
                 continue
             if self.place_bet:
-                amount = self.calculate_bet_amount(result, market)
+                amount_and_direction = self.calculate_bet_amount_and_direction(
```
evangriffiths, line 305: It feels to me odd that we're computing an `Answer` above with `result = self.answer_binary_market(market)` (which already contains a bet direction == `result.decision`) and then computing a direction again here.

Not sure how much of a can of worms that is to refactor, but I think that should come first (or as part of this PR)

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -316,12 +302,17 @@ def process_bets(self, market_type: MarketType) -> None:
...
+                    result, market
+                )
                 logger.info(
-                    f"Placing bet on {market} with result {result} and amount {amount}"
+                    f"Placing bet on {market} with direction {amount_and_direction.direction} and amount {amount_and_direction.amount}"
```
evangriffiths, line 309: Is `amount_and_direction.direction` guaranteed to match `result.decision`?

ape-config.yaml
```diff
@@ -8,7 +8,7 @@ default_ecosystem: gnosis
 node:
   gnosis:
     mainnet:
-      uri: https://gnosis-rpc.publicnode.com
+      uri: https://rpc.gnosischain.com
```
evangriffiths, line 11: Would be great for this (and deps changes) to be a separate PR, as it's unrelated to this betting strategy stuff

prediction_market_agent_tooling/deploy/agent.py
gabrielfior: In the default case (`FixedBetBettingStrategy`), yes.
In the `KellyBettingStrategy`, we overwrite the direction and they might not match.

ape-config.yaml
gabrielfior: Sure.
https://github.com/gnosis/prediction-market-agent-tooling/issues/355

prediction_market_agent_tooling/deploy/agent.py
gabrielfior: That's an important aspect.
>It feels to me odd that we're computing an Answer above with result = self.answer_binary_market(market) (which already contains a bet direction == result.decision) and then computing a direction again here.

We need the Answer because it contains `estimated_p_yes`, required by the Kelly criterion (which can then revert the direction). To compare that with an agent case, the TTA creates an Answer object with a "black-box" logic - if we provide the Kelly strategy, then that's how we want to calculate bet sizes | direction.

Another idea I see is having the agents simply output `p_yes`, `rationale`, etc but NOT directions. In this case it's clear how direction was defined.
Problem with that is that `FixedBettingStrategy` copies direction from Answer - so in this case, we would need a logic for deciding direction (e.g. `p_yes` > 0.5, go YES) or similar, but also observing the market `p_yes`.

scripts/sell_all_omen.py
gabrielfior: abstracted to a common variable and refactored both usages

prediction_market_agent_tooling/tools/web3_utils.py
gabrielfior: Reverted. This will get done on https://github.com/gnosis/prediction-market-agent-tooling/issues/356.
For context - `AlreadyKnown` happened when a tx with the same parameters (method, args, nonce) was already present in the RPC mempool (due to the fact that the agent was running at the same time on GCP), hence the RPC outputs "I have seen this transaction before and will process it".

prediction_market_agent_tooling/markets/data_models.py
gabrielfior: Moved.

scripts/sell_all_omen.py
kongzii: But why not just keep `get_tiny_bet_amount` as it is and use that? What was the motivation to remove it? (is there a need for that? from the code it doesn't look like it)

We have `AgentMarket` as a abstract class that our agents can work with, if something is subclass of `AgentMarket`, agent should just work with it (if all methods are implemented correctly).

But now, if we introduce a new `SomeNewMarket(AgentMarket)`, it's not clear that there are places in the code where we accept AgentMarket, yet it won't work.

prediction_market_agent_tooling/deploy/betting_strategy.py
gabrielfior: My approach here was to have a class `Strategy` handling bet sizing - it can use MarketType as part of its logic (represented by `isisntace`), however it does not have to (see Kelly case).

kongzii: The strategy can use `get_tiny_bet_amount`, so why not? And do market-checking-logic in strategies where it's needed.

scripts/sell_all_omen.py
gabrielfior: It was an implementation decision. Certainly `get_tiny_bet_amount` used to work, but in your example, if we have a new market (`Polymarket(AgentMarket)`, we would have to overwrite `get_tiny_bet_amount` implementation for that subclass - same here, we extend the logic on the FixedBettingStrategy. This is always needed because the bet_amount involves currency, which is different from market type to market type.

> But now, if we introduce a new SomeNewMarket(AgentMarket), it's not clear that there are places in the code where we accept AgentMarket, yet it won't work.
The same will happen, only the exception will be different - `NotImplementedError` vs `ValueError` (from the Strategy `get_max_bet_amount_for_market`).

kongzii: > This is always needed 

It was needed, but the difference is that it was forced. Because AgentMartet class has many such methods: 

Screenshot by Dropbox Capture

So, if we add new market support, it should be as hard as possible to forget about something. 

That was the idea behind AgentMarket. If we go with this other way of doing things, we will need to keep track of all the places around the code that need an update to make them work with basic agent stuff. And then AgentMarket class could be removed.

kongzii: > I can create a class without any warnings from the IDE

I assume one would scan through the `AgentMarket` class to check out what needs to be implemented, not just subclass it and hope for the best. But you are right that it would be great to have some kind of warning there.

And seems like that can be achieved easily:

1. Make it an abstract class: `class AgentMarket(BaseModel, ABC):`
2. Add @abstractmehod decorator
3. Profit:

Screenshot by Dropbox Capture

@coderabbitai please create an issue to do this everywhere where we use the NotImplemented pattern.

> I have an idea that uses mypy and generics, will try it out and ask for your input.

I mean, you can, but then will you change all the `Subclass needs to implement this` methods in AgentMarket? That's kind of a big refactor for something that's not actually required for this PR 🤷 And @evangriffiths put a lot of work into AgentMarket class back then; I think he experimented with many approaches before we stuck with this one.

prediction_market_agent_tooling/deploy/agent.py
evangriffiths: > Another idea I see is having the agents simply output p_yes, rationale, etc but NOT directions.

Yes, this one!! This is the refactor I meant in the description for the ticket (https://github.com/gnosis/prediction-market-agent-tooling/issues/338#issue-2451317112) 😁 (sorry if it wasn't very clear).

scripts/sell_all_omen.py
kongzii: Yeah, it seems like a misunderstanding.

`KellyBettingStrategy` is great; no change is needed there from my point of view.

`FixedBetBettingStrategy` could be renamed to `TinyBetStrategy` and use the `get_tiny_bet_amount` method of the market class. It makes sense to me that `TinyBetStrategy` would be the default strategy, because it's safe, and it will just work with arbitrary AgentMarket.

`FixedBetBettingStrategy` can exist as well, but it would take a `amount` argument in its `__init__` method, and again, it doesn't have to use any `get_tiny_bet_amount`-like implementation, just return the fixed value.

wdyt?

gabrielfior: I implemented a mix, see https://github.com/gnosis/prediction-market-agent-tooling/pull/352/commits/683b95ff0015ea770263ff7dfc1cd0fa7cb871f9#diff-ebe828b1c77960ecaea62858ada5f1ee0b662b19051aa2ffbf92cd6ea6287110R25

I like the idea of using `get_tiny_bet_amount`, because this will break (mypy will complain) if another AgentMarket is implemented which does not have the method `get_tiny_bet_amount`. Together with ABC from the other issue, this is robust enough for me.

prediction_market_agent_tooling/deploy/agent.py
gabrielfior: Makes sense. Removed `direction` from Answer and adapted the method, will change this in PMA when agents produce Answers.

gabrielfior: Implemented this, see `ProbabilisticAnswer` - also adapted PMA to restrain the agent from producing `Direction` - this will be done by the Strategy itself using `estimate_p_yes` and `market.current_p_yes`

prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+class BaseCalculator(ABC, Generic[T]):
+    choices: dict[Resolution, str]
+
+    @abstractmethod
+    def calculate_shares_for_outcome(
+        self, market: T, bet_amount: float, bet_choice: str
+    ) -> float:
+        pass
```
gabrielfior, line 40: Tests will still be added before merging

prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -0,0 +1,160 @@
...
+    The calculation is rather simple:
+        expected_value = (probability_yes * payoff_yes) - (probability_no * payoff_no)
+        return True if expected_value >=0 else False
+    """
+
+    choices: dict[Resolution, str]
+
+    @abstractmethod
+    def calculate_shares_for_outcome(
+        self, market: T, bet_amount: float, bet_choice: str
```
kongzii, line 45: Why not just accept `AgentMarket`?

prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -0,0 +1,160 @@
...
+            100  # as example, does not affect the calculation of Expected Value below.
+        )
+        num_tokens_yes = self.calculate_shares_for_outcome(
+            market, fixed_bet_amount, self.choices[Resolution.YES]
+        )
+        num_tokens_no = self.calculate_shares_for_outcome(
+            market, fixed_bet_amount, self.choices[Resolution.NO]
+        )
+        return (estimate_p_yes * num_tokens_yes) - (
+            (1 - estimate_p_yes) * num_tokens_no
```
kongzii, line 60: Shouldn't this be calculated using $ value instead of token (shares) value?

prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -0,0 +1,160 @@
...
+    def calculate_bet_amount_and_direction(
+        self, answer: ProbabilisticAnswer, market: AgentMarket
+    ) -> TokenAmountAndDirection:
+        pass
+
+
+class FixedBetBettingStrategy(BettingStrategy):
+    def calculate_direction(self, market: AgentMarket, estimate_p_yes: float) -> bool:
+        if isinstance(market, ManifoldAgentMarket):
+            calculator = ManifoldDirectionCalculator()
```
kongzii, line 116: I'd say `FixedBetBettingStrategy` shouldn't use this new direction strategy because that will change current behavior for all agents, and we didn't want to do that, or?

gabrielfior: ToDo
- Create FixedStrategy (P.0.5, bet yes, else bet no)
- ExpectedValue is an additional agent (like Kelly)

gabrielfior: @evangriffiths proved that expected_value considerations are equivalent to the condition
```
bet TRUE if estimate_p_yes > market.p_yes else bet FALSE
```
hence implemented this as default on `FixedBetBettingStrategy`, which is the default for all agents.

prediction_market_agent_tooling/deploy/betting_strategy.py
gabrielfior: This discussion was answered on a comment in another [thread](https://github.com/gnosis/prediction-market-agent-tooling/pull/352#discussion_r1725174278) - thus closing this

prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -0,0 +1,63 @@
...
+    ProbabilisticAnswer,
+    TokenAmountAndDirection,
+)
+from prediction_market_agent_tooling.tools.betting_strategies.kelly_criterion import (
+    get_kelly_bet,
+)
+
+T = TypeVar(
+    "T"
+)  # Not possible to use bound due to OmenMarket and ManifoldMarket not sharing a base class
```
evangriffiths, line 15: Looks like this is unused, so can remove
```suggestion
```

prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -0,0 +1,63 @@
...
+        return TokenAmountAndDirection(
+            amount=bet_amount,
+            currency=market.currency,
+            direction=direction,
+        )
+
+
+class KellyBettingStrategy(BettingStrategy):
+    @staticmethod
+    def get_max_bet_amount_for_market() -> float:
```
evangriffiths, line 48: I think it makes more sense for `max_bet_amount: float` to be an attribute of the class, settable in the `__init__` method:

```
class KellyBettingStrategy(BettingStrategy):
    def __init__(
            self,
            max_bet_amount: float = 10, # Currency-independent
        ):
        super().__init__()
        self.max_bet_amount = max_bet_amount

    ...
```
so agents can choose their own max bet size if they'd like (e.g. depending on their wallet size/risk level)

prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -0,0 +1,63 @@
...
+        # No difference between markets.
+        return 10  # Mana or xDAI
```
evangriffiths, line 50: Or poly-whatever-it's-called! The list might get bigger/smaller, so no need to mention specific currencies.
```suggestion
        return 10
```

pyproject.toml
```diff
@@ -57,8 +58,8 @@ google = ["google-api-python-client"]
 pytest = "*"
 mypy = "^1.11.1"
 black = "^23.12.1"
-ape-foundry = "^0.8.2"
-eth-ape = "^0.8.10"
+eth-ape = "^0.8.12"
+ape-foundry = "^0.8.4"
```
evangriffiths, line 62: These look unrelated to this PR

prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -0,0 +1,63 @@
...
+
+class BettingStrategy(ABC):
+    @abstractmethod
+    def calculate_bet_amount_and_direction(
+        self, answer: ProbabilisticAnswer, market: AgentMarket
+    ) -> TokenAmountAndDirection:
+        pass
+
+
+class FixedBetBettingStrategy(BettingStrategy):
```
evangriffiths, line 26: Are you set on 'FixedBet' as a name? If not, just suggesting 'MaxAccuracy', because this is the strategy that will give you the highest accuracy (% correct bets) possible.

prediction_market_agent_tooling/tools/betting_strategies/kelly_criterion.py
```diff
@@ -1,29 +1,22 @@
...
+    size: float
+
+
 def get_kelly_bet(
     max_bet: float,
     market_p_yes: float,
     estimated_p_yes: float,
     confidence: float,
-) -> KellyBet:
+) -> KellyResult:
```
evangriffiths, line 19: You don't like KellyBet? 😢  It's what they call it on wikipedia ("The kelly bet is...")

prediction_market_agent_tooling/deploy/betting_strategy.py
gabrielfior: Happy to rename this. MaxAccuracy approved.

prediction_market_agent_tooling/tools/betting_strategies/kelly_criterion.py
gabrielfior: KellyBet works, result is anyways too broad. Accepted.

pyproject.toml
gabrielfior: I just upgraded versions

prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -0,0 +1,58 @@
...
+            currency=market.currency,
+            direction=direction,
+        )
+
+
+class KellyBettingStrategy(BettingStrategy):
+    @staticmethod
+    def get_max_bet_amount_for_market() -> float:
+        # No difference between markets.
+        return 10
```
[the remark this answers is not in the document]
evangriffiths: dude you literally just copied my comment from earlier https://github.com/gnosis/prediction-market-agent-tooling/pull/352/files#r1729201133

pyproject.toml
evangriffiths: how come?

prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -0,0 +1,58 @@
...
+    def calculate_direction(market_p_yes: float, estimate_p_yes: float) -> bool:
+        # If estimate_p_yes >= market.current_p_yes, then bet TRUE, else bet FALSE.
+        # This is equivalent to saying EXPECTED_VALUE = (estimate_p_yes * num_tokens_obtained_by_betting_yes) -
+        # ((1 - estimate_p_yes) * num_tokens_obtained_by_betting_no) >= 0
+        return estimate_p_yes >= market_p_yes
+
+    def calculate_bet_amount_and_direction(
+        self, answer: ProbabilisticAnswer, market: AgentMarket
+    ) -> TokenAmountAndDirection:
+        bet_amount = market.get_tiny_bet_amount().amount
```
evangriffiths, line 32: oo, similar to https://github.com/gnosis/prediction-market-agent-tooling/pull/352/files#r1729201133, I think it makes more sense for the bet amount here to be configurable, with tiny as default (although this needs a bit of fiddling for it to work):
```
class MaxAccuracyBettingStrategy(BettingStrategy):
    def __init__(
        self,
        bet_amount: amount: Union[float, Literal["tiny"]] = "tiny",
    ):
        super().__init__()
        self.bet_amount = bet_amount

    def calculate_bet_amount_and_direction(
        self, answer: ProbabilisticAnswer, market: AgentMarket
    ) -> TokenAmountAndDirection:
        bet_amount = market.get_tiny_bet_amount().amount if self.bet_amount == "tiny" else self.bet_amount
        ...
```

prediction_market_agent_tooling/deploy/betting_strategy.py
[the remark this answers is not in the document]
evangriffiths: no worries, i'm sorry for being rude

prediction_market_agent_tooling/deploy/betting_strategy.py
gabrielfior: I thought None instead of "tiny" would be more practical - I assume it's also easier for the general agent to understand None or fixed value instead of tiny.
Let me know if you accept this variant.

prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -0,0 +1,65 @@
...
+    ProbabilisticAnswer,
+    TokenAmountAndDirection,
+)
+from prediction_market_agent_tooling.tools.betting_strategies.kelly_criterion import (
+    get_kelly_bet,
+)
+
+
+class BettingStrategy(ABC):
+    def __init__(self, bet_amount: float | None = None):
```
evangriffiths, line 14: no a big deal, but I'm not sure about bet_amount being on the base class. because in the KellyBettingStrategy case, you're specifying a 'max bet amount', which is different to how much you actually bet

prediction_market_agent_tooling/deploy/betting_strategy.py
evangriffiths: oo yes that's better!

evangriffiths (approved):
Nice one. Have left a couple things up to you. Oo also the title needs updating now!

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -360,13 +372,15 @@ def verify_market(self, market_type: MarketType, market: AgentMarket) -> bool:
...
 
-    def answer_binary_market(self, market: AgentMarket) -> Answer | None:
+    def answer_binary_market(self, market: AgentMarket) -> ProbabilisticAnswer | None:
         """
         Answer the binary market. This method must be implemented by the subclass.
         """
         raise NotImplementedError("This method must be implemented by the subclass")
 
-    def calculate_bet_amount(self, answer: Answer, market: AgentMarket) -> BetAmount:
+    def calculate_bet_amount(
```
kongzii, line 381: Is this older version still used? I can't find usage of it anywhere

kongzii (changes requested):

prediction_market_agent_tooling/deploy/agent.py
gabrielfior: thanks, removed

prediction_market_agent_tooling/tools/betting_strategies/kelly_criterion.py
```diff
@@ -47,10 +40,10 @@ def get_kelly_bet(
     check_is_valid_probability(confidence)
 
     if estimated_p_yes > market_p_yes:
-        bet_direction = BetDirection.YES
+        bet_direction = True
```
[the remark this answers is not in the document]
gabrielfior: This was discussed in another thread (can't remember which one from the 80 comments) - if you have specific concerns, happy to open a new PR to discuss having "YES" or True.

prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -0,0 +1,65 @@
...
+            currency=market.currency,
+            direction=direction,
+        )
+
+
+class KellyBettingStrategy(BettingStrategy):
+    bet_amount: float
+
+    def __init__(self, bet_amount: float = 10):
+        # We add a new default here because it represents an upper bound.
```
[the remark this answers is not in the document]
gabrielfior: Implemented, removed

prediction_market_agent_tooling/deploy/betting_strategy.py
gabrielfior: done

pyproject.toml
gabrielfior: reverted upgrade

prediction_market_agent_tooling/deploy/betting_strategy.py
gabrielfior: implemented

kongzii (approved):

PR #360, opened 2024-08-26
Retry always except for `reverted` messages

kongzii:


prediction_market_agent_tooling/tools/web3_utils.py
```diff
@@ -154,11 +154,8 @@ def _prepare_tx_params(
 
 
 @tenacity.retry(
-    # Retry only for the transaction errors that match the given patterns,
-    # add other retrieable errors gradually to be safe.
-    retry=tenacity.retry_if_exception_message(
-        match="(.*wrong transaction nonce.*)|(.*Invalid.*)|(.*OldNonce.*)"
-    ),
+    # Don't retry on `reverted` messages, as they would always fail again.
+    retry=tenacity.retry_if_exception_message(match="^(?!.*reverted.*)$"),
```
gabrielfior, line 158: Is `reverted` (lowercase) a proper match? I'm worried that doesn't always match properly - see issues below

https://stackoverflow.com/questions/57445305/how-to-get-solidity-revert-require-error-reason-in-python-and-web3-py
https://ethereum.stackexchange.com/questions/118394/why-does-the-call-method-from-web3-py-cause-a-revert-when-combined-with-transf

gabrielfior:
For testing purposes - can you add a test like below that asserts that retry was NOT triggered?
- Use local_web3 (so that we have the Gnosis chain state)
- Use anvil's test account[0] and [1]
- Use wxDAI
- Call transferFrom(sender=accounts[0], to=accounts[1]) - this will trigger a revert because of 
```
            require(allowance[src][msg.sender] >= wad);
```
https://gnosis.blockscout.com/token/0xe91D153E0b41518A2Ce8Dd3D7944Fa863463a97d?tab=contract

prediction_market_agent_tooling/tools/web3_utils.py
kongzii: Good point, I added case insensitivity there

gabrielfior (approved):

PR #364, opened 2024-08-27
Add comment on omen market life cycle

evangriffiths:


prediction_market_agent_tooling/markets/omen/data_models.py
```diff
@@ -164,7 +164,15 @@ def redeemable(self) -> bool:
 
 class OmenMarket(BaseModel):
     """
-    https://aiomen.eth.limo
+    https://presagio.pages.dev
+
+    An Omen market goes through the following stages:
+
+    1. creation - can add liquidty immediately, and trade immediately if there is liquidity
+    2. closing - market is closed, and a question is simultaneously created on Reality
```
kongzii, line 172:
```suggestion
    2. closing - market is closed, and a question is simultaneously opened for answers on Reality
```

question on reality is actually created even before the market is, but it opens for answers at the same time marker closes

kongzii (approved):

PR #366, opened 2024-08-28
Bugfix for is_predictable_without_description

kongzii:


prediction_market_agent_tooling/tools/is_predictable.py
```diff
@@ -143,7 +143,7 @@ def is_predictable_without_description(
         description=description,
     )
     completion = str(
-        llm(
+        llm.invoke(
```
kongzii, line 146: Without `invoke`, it doesn't support the `config` parameter. Sorry! 🙏

kongzii: These tests are behind a flag, so CI didn't catch it:

```
RUN_PAID_TESTS=1 pytest tests -k test_is_predictable_without_description 
```

kongzii: And `__call__` method is defined as 

```
def __call__(
        self,
        messages: List[BaseMessage],
        stop: Optional[List[str]] = None,
        callbacks: Callbacks = None,
        **kwargs: Any,
    ) -> BaseMessage:
```

so mypy doesn't complain about `config` and it's a proper call because of `**kwargs`. That's why I hate **kwargs btw 😄

evangriffiths (approved):

PR #368, opened 2024-08-29
Fix claiming of bonds if multiple answers are provided

kongzii:


prediction_market_agent_tooling/markets/omen/omen_resolving.py
```diff
@@ -77,50 +66,39 @@ def claim_bonds_on_realitio_question(
...
 
-    if not answers_objects:
-        raise ValueError(f"No answers found for {question.questionId=}")
+    # They need to be processed in order.
+    responses = sorted(responses, key=lambda x: x.timestamp)
 
-    if answers_objects[-1].question.historyHash == ZERO_BYTES:
-        raise ValueError(f"Already claimed {question.questionId=}.")
+    if not responses:
+        raise ValueError(f"No answers found for {question.questionId.hex()=}")
```
gabrielfior, line 75: Isn't it possible to have questions with no answers?

kongzii: Without answers, there's nothing to claim

tests_integration/markets/omen/test_reality.py
```diff
@@ -0,0 +1,69 @@
...
+from prediction_market_agent_tooling.markets.omen.omen_subgraph_handler import (
+    OmenSubgraphHandler,
+)
+from prediction_market_agent_tooling.tools.utils import utcnow
+from prediction_market_agent_tooling.tools.web3_utils import xdai_to_wei
+from tests.utils import RUN_PAID_TESTS
+
+
+@pytest.mark.skipif(not RUN_PAID_TESTS, reason="This test costs money to run.")
+def test_claim_bonds() -> None:
```
gabrielfior, line 25: Any reason why not to run this in the `local_chain`?

tests_integration/markets/omen/test_reality.py
```diff
@@ -0,0 +1,69 @@
...
+        )
+        bond = wei_type(bond * 2)
+        logger.info(f"Answered with: {answer}")
+        time.sleep(2)  # Give it a moment to settle on chain.
+
+    time.sleep(timeout.total_seconds() + 2)  # Wait for the question to be finalized.
+
+    # Try to claim bonds
+    question = OmenSubgraphHandler().get_questions(question_id_in=[question_id])[0]
+    logger.info(f"Claiming for {question.url}")
```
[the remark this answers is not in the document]
kongzii: It's not available in the contract. Data would have to be reconstructed somehow.

prediction_market_agent_tooling/markets/omen/omen_resolving.py
gabrielfior: Sure, but where is this being handled with `try-catch`?
Just worried the agent might error out although this is "expected" behavior.

tests_integration/markets/omen/test_reality.py
gabrielfior: How often do we run these tests? I believe not very often.
As this component is rather critical, I would push for this being either a. executed on every test run or b. having it executed on a local chain (see answer below for replacing graph api).

prediction_market_agent_tooling/markets/omen/omen_resolving.py
kongzii: If someone calls this on question without any answer, I'd consider it a bug that needs to be fixed.

tests_integration/markets/omen/test_reality.py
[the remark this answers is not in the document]
gabrielfior: You can either a. calculate the new history hash manually (following what the contract does) or b. listen for the events emitted in the RPC (Anvil chain running locally on port 8545) - since every transaction mines a new block, you can simply fetch the latest events using web3.py ([link](https://web3py.readthedocs.io/en/stable/filters.html#events-and-logs)).
Happy to assist further here.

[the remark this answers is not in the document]
kongzii: > b. listen for the events emitted in the RPC

That would mean testing something else, not the thing that's going on in the function ( where it's coming from the subgraph)

> a. calculate the new history hash manually

C'mon, man! 😄 `OmenSubgraphHandler` was used here before, plus I implemented the fix exactly as you proposed. Can we just make an issue out of it for whenever someone is bored?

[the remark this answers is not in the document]
gabrielfior: @coderabbit create an issue for refactoring this test to use the local_web3 fixture.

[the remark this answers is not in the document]
gabrielfior: Created https://github.com/gnosis/prediction-market-agent-tooling/issues/371

gabrielfior (approved):

PR #369, opened 2024-08-30
Fix polymarket data models

kongzii:


prediction_market_agent_tooling/markets/polymarket/data_models_web.py
```diff
@@ -38,7 +38,7 @@ class Event(BaseModel):
 
 
 class Event1(BaseModel):
-    startDate: datetime
+    startDate: datetime | None = None
```
kongzii, line 41: I'm tired of fixing it nonstop, so I propose just making optional anything that we don't need 😄

evangriffiths (approved):

PR #374, opened 2024-09-02
423 sell prev positions

gabrielfior:
-> Added check at the `DeployableTraderAgent` level, to sell previous, contradicting positions (for a given market) with respect to the direction of the current bet being placed.
-> Added local_chain test for Omen
-> Kept default implementation (do nothing) for Manifold since payouts there not top priority


prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -419,9 +423,18 @@ def process_market(
             logger.info(
                 f"Placing bet on {market} with direction {amount_and_direction.direction} and amount {amount_and_direction.amount}"
             )
+            extra_bet_amount_from_opposite_bets: float = 0
```
evangriffiths, line 426: Or in one line...
```
# If we have an existing position, sell it first if they bet on a different outcome.
extra_bet_amount_from_opposite_bets = 0 if allow_opposite_bets else market.liquidate_existing_positions(amount_and_direction.direction)
```

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -419,9 +423,18 @@ def process_market(
...
+                sold_amount = market.liquidate_existing_positions(
+                    amount_and_direction.direction
+                )
+                extra_bet_amount_from_opposite_bets += sold_amount
+
             market.place_bet(
                 amount=TokenAmount(
-                    amount=amount_and_direction.amount,
+                    amount=amount_and_direction.amount
+                    + extra_bet_amount_from_opposite_bets,
```
evangriffiths, line 437: I guess the other way to do this would be to call `liquidate_existing_positions` before `calculate_bet_amount_and_direction`, and then let the agent's strategy decide the amount in one go.

I can't think of a good reason for doing this over your way though (apart from simplifying the logic in the liquidate function, as you can return None, so don't need to call `calculate_sell_amount_in_collateral` any more). Any reason you picked this way, out of interest?

tests_integration_with_local_chain/markets/omen/test_omen.py
```diff
@@ -350,3 +356,69 @@ def test_place_bet_with_autodeposit(
...
+
+    conditional_token = OmenConditionalTokenContract()
+    conditional_tokens_contract = local_web3.eth.contract(
+        address=conditional_token.address, abi=conditional_token.abi
+    )
+    # We fetch the transfer single event emitted when outcome tokens were bought
+    ls = conditional_tokens_contract.events.TransferSingle().get_logs()  # type: ignore[attr-defined]
+    pos_id = ls[-1]["args"]["id"]
+    # check position
+    position_balance = get_position_balance_by_position_id(
```
evangriffiths, line 394: Ah is this a workaround that means you can use the local chain, because `market.get_positions` uses the subgraph?

prediction_market_agent_tooling/deploy/agent.py
kongzii: Yeah, whenever possible, it should be preferred over declaring new variables in `if` block 🙏

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -149,6 +149,51 @@ def get_liquidity(self) -> TokenAmount:
     def get_tiny_bet_amount(cls) -> BetAmount:
         return BetAmount(amount=0.00001, currency=cls.currency)
 
+    @staticmethod
+    def get_bet_outcome(binary_outcome: bool) -> str:
+        return OMEN_TRUE_OUTCOME if binary_outcome else OMEN_FALSE_OUTCOME
+
+    @staticmethod
+    def get_bet_outcome_bool(bet_outcome: OutcomeStr) -> bool:
```
kongzii, line 157: We already have https://github.com/gnosis/prediction-market-agent-tooling/blob/4840eb77720c6a6b2f71525e7f4de93391cc0d1c/prediction_market_agent_tooling/markets/omen/data_models.py#L39

kongzii: No preference over using this one and removing the other or removing this and keeping the other. But in any case, please keep the `raise ValueError(f"Outcome `{outcome_str}` is not a valid boolean outcome.")` logic there.

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -149,6 +149,51 @@ def get_liquidity(self) -> TokenAmount:
     def get_tiny_bet_amount(cls) -> BetAmount:
         return BetAmount(amount=0.00001, currency=cls.currency)
 
+    @staticmethod
+    def get_bet_outcome(binary_outcome: bool) -> str:
```
kongzii, line 153: If you decide to use `def get_boolean_outcome` and remove `get_bet_outcome_bool` in this class. Please also move this function next to the other one. 🙏

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -149,6 +149,51 @@ def get_liquidity(self) -> TokenAmount:
...
+        return True if bet_outcome == OMEN_TRUE_OUTCOME else False
+
+    def liquidate_existing_positions(
+        self, bet_outcome: bool, web3: Web3 | None = None
+    ) -> float:
+        """
+        Liquidates all previously existing positions.
+        Returns the amount in collateral obtained by selling the positions.
+        """
+        better_address = APIKeys().public_key
```
kongzii, line 167:
```suggestion
        better_address = APIKeys().bet_from_address
```

I guess this is preferred in this function? (in case that bets are from Safe)

tests_integration_with_local_chain/markets/omen/test_omen.py
```diff
@@ -350,3 +356,69 @@ def test_place_bet_with_autodeposit(
...
+) -> None:
+    # Fetch an open binary market.
+    sh = OmenSubgraphHandler()
+    market = sh.get_omen_binary_markets_simple(
+        limit=1, filter_by=FilterBy.OPEN, sort_by=SortBy.CLOSING_SOONEST
+    )[0]
+    omen_agent_market = OmenAgentMarket.from_data_model(market)
+
+    # Place a bet using a standard account (from .env)
+    bet_amount = BetAmount(amount=1, currency=Currency.xDai)
```
kongzii, line 383: Won't this break if Github CI's wallet happens to have less than one xDai?

tests_integration_with_local_chain/markets/omen/test_omen.py
```diff
@@ -350,3 +356,69 @@ def test_place_bet_with_autodeposit(
         web3=local_web3,
         api_keys=test_keys,
     )
+
+
+def get_position_balance_by_position_id(
+    from_address: ChecksumAddress, position_id: int, web3: Web3
+) -> int:
```
kongzii, line 363:
```suggestion
) -> Wei:
```

tests_integration_with_local_chain/markets/omen/test_omen.py
```diff
@@ -350,3 +356,69 @@ def test_place_bet_with_autodeposit(
...
+    ):
+        # We now want to sell the recently opened position.
+        omen_agent_market.liquidate_existing_positions(False, web3=local_web3)
+
+    position_balance_after_sell = get_position_balance_by_position_id(
+        from_address=test_keys.bet_from_address, position_id=pos_id, web3=local_web3
+    )
+
+    # Assert positions were liquidated
+    assert wei_to_xdai(Wei(position_balance_after_sell)) < 0.001  # xDAI
```
kongzii, line 424:
```suggestion
    assert wei_to_xdai(position_balance_after_sell) < 0.001  # xDAI
```

tests_integration_with_local_chain/markets/omen/test_omen.py
gabrielfior: Yes, this way you pick up the event "manually", since subgraph doesn't pick up events from local chains.
One can do a lot better (with structured events etc), but at least you get named properties (id, value, etc) as per event below:
```
event TransferSingle(address indexed operator, address indexed from, address indexed to, uint256 id, uint256 value);
```

tests_integration_with_local_chain/markets/omen/test_omen.py
gabrielfior: Thanks, yes it would - I refactored it to explicity accept `TestKeys`, which point to a test account from anvil with 100 ETH.

prediction_market_agent_tooling/deploy/agent.py
gabrielfior: Cool, implemented

prediction_market_agent_tooling/deploy/agent.py
gabrielfior: I don't love this arg passing (as you highlighted)
```
amount=amount_and_direction.amount
                    + extra_bet_amount_from_opposite_bets,
```
and I understand what you are saying, e.g. if we have a total amount of 1 xDAI for placing a bet and Kelly says, "place a bet of 50% of your money into market X", we should place a bet of 50% of (new_bet_amount + old_positions). This, however, implies adjusting the strategy`s `bet_amount` on the fly, so let's say we do the following
- process_market_1 with prev_positions -> edit strategy, now `bet_amount`=10 (it was 1 before)
- process_market_2 without prev_positions -> strategy was edited earlier, so bet_amount is 10 (but we want 1)

Curious if you have any suggestions here.

prediction_market_agent_tooling/markets/manifold/manifold.py
```diff
@@ -49,6 +49,10 @@ def get_minimum_bet_to_win(self, answer: bool, amount_to_win: float) -> Mana:
         # Manifold lowest bet is 1 Mana, so we need to ceil the result.
         return mana_type(ceil(minimum_bet_to_win(answer, amount_to_win, self)))
 
+    def liquidate_existing_positions(self, outcome: bool) -> float:
+        # Not implemented for Manifold.
```
kongzii, line 53: This sets a precedence that PMAT won't be a universal PM framework, but rather an Omen one.  Currently, we do have agents betting on Manifold, and from user's POV it looks like Manifold is supported, although it doesn't do anything in reality.

If I remember correctly, until now, everything is supported for both Omen and Manifold, except for Omen-specific stuff (such as redeeming winnings).

However, I do agree that it doesn't make much sense to spend time now implementing this for Manifold.

I see a few options here:

1. Because this is called behind a `allow_opposite_bets` flag, instead of `return 0`, this method can be removed, so a `NotImplemented` exception will be raised if anyone tries to use it. Someone can implement it if it will be required in the future. 
2. Remove other markets from the trading functionality. 
3. Implement it.

I'd currently vote for (1), because is less work 😄 (just removing this method, but keep in mind you will need to somehow set `allow_opposite_bets=True` for manifold-deployed agents).

prediction_market_agent_tooling/markets/omen/omen.py
gabrielfior: Thanks, removed the new method.

prediction_market_agent_tooling/deploy/agent.py
gabrielfior: As discussed with @evangriffiths  - proceeding as follows
- liquidation occurs -> wxDAI added to balance, NOT used for the bet happening immediately afterwards
- #377 will increase bet sizes since it gives Kelly access to the entire balance

prediction_market_agent_tooling/markets/manifold/manifold.py
gabrielfior: I considered implementing the logic (3) for Manifold, but it would involve adding many additional methods to the Manifold API, incl. retrieving positions, etc that are not directly related to the PR.

Regarding implementation:
-> Setting allow_opposite_bets = True for manifold and `False` for Omen would involve creating a dictionary e.g.
```
allow_opposite_bets = {'omen':True,'manifold':False}
```
and adding this to DeployableTraderAgent. This is in my opinion less practical than setting a default implementation for Manifold, i.e. 
```
# prediction_market_agent_tooling/markets/manifold/manifold.py
  def liquidate_existing_positions(self, outcome: bool) -> None:
      # Not implemented for Manifold.
      pass
```
I agree that it's important to know (either via Exception or someway else) that positions are not being liquidated in the Manifold case. I added a log entry to aid in this - if you think the selling positions is something important to Manifold and we should work on that now, I would say we should create a PR and implement (3).

kongzii: > allow_opposite_bets = {'omen':True,'manifold':False}

I meant that this should be configurable on the agent level. Something like

```
class DeployableTraderAgent(DeployableAgent):
   allow_opposite_bets: bool = True. # We would set this to False for the Manifold agents
```

That should be a few-line change and keep things in order, without fake methods that aren't doing what they should.

Keep in mind that we have also `MetaculusAgentMarket` and `PolymarketAgentMarket`, so currently we have three different behaviors in this PR:

1. Omen where it works as expected
2. Manifold where it pretends to work
3. Polymarket & Metaculus where it raises an exception

gabrielfior: > allow_opposite_bets: bool = False  # We would set this to True for the Manifold agents

The market_type is only defined at runtime, hence not possible to define `allow_opposite_bets` on a class level for Manifold market type.

> That should be a few-line change and keep things in order, without fake methods that aren't doing what they should.
Not sure how this can be implemented - could you share more?

kongzii: Currently `allow_opposite_bets` is a flag in `process_market` method of `DeployableTraderAgent`. Instead of that, if you define `allow_opposite_bets` as 

```
class DeployableTraderAgent(DeployableAgent):
   allow_opposite_bets: bool = False
```

On the class itself (and remove it from the method arguments), you can just do:

```
if not self.allow_opposite_bets:
```

so very similar to the current implementation, just using the `self`.

Thanks to it, we can set it to `allow_opposite_bets=True` for the Manifold agent. In such scenario, `liquidate_existing_positions` will never be called, so it doesn't have to be implemented on the ManifoldAgentarket.

kongzii: I see the problem now. That would mean we have to define new classes for Manifold agents -- but maybe that isn't so bad?

gabrielfior: > That would mean we have to define new classes for Manifold agents 

I like having market-specific classes, but the trade-off would be having a slightly more complicated `run_agent` [file](https://github.com/gnosis/prediction-market-agent/blob/main/prediction_market_agent/run_agent.py#L75) .
Happy to huddle tomorrow and agree on next steps.

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -149,6 +151,35 @@ def get_liquidity(self) -> TokenAmount:
...
+        )
+
+        for prev_position in prev_positions_for_market:
+            for position_outcome, token_amount in prev_position.amounts.items():
+                position_outcome_bool = get_boolean_outcome(position_outcome)
+                if position_outcome_bool != bet_outcome:
+                    # We keep it as collateral since we want to place a bet immediately after this function.
+                    self.sell_tokens(
+                        outcome=position_outcome_bool,
+                        amount=token_amount,
```
kongzii, line 178: Another evening thought before I forget 😄 

Here we are selling position based on its token amount obtained from `get_positions` function. A couple of questions:

1. After we sell it the first time, on the second run, what will happen? Will it be `0`? And if it's `0`, will the `sell_tokens` function work with `0`?
2. Is Hugo accounting for the possibility of sold tokens? I'd say that if the agent sells his previous bets, they shouldn't be calculated into its accuracy. (our monitoring app in PMAT probably needs to changed as well)

prediction_market_agent_tooling/markets/manifold/manifold.py
kongzii: Yeah. I'm also fine with shutting down Manifold deployments, personally I don't remember last time I checked on them and we have Replicator for having metrics against Manifold markets on omen.

So if you would remove this method, to make it throw an exception if used, make `allow_opposite_bets ` class-level variable on DeployableTraderAgent, and then just remove the Manifold deployments in the infra repository, that's it.

But here let's check with @evangriffiths if he agrees as well.

evangriffiths: Yes this makes sense. Happy with retiring manifold deployments. I also like the idea of `allow_opposite_bets` being a class property. It feels similar to betting strategy - like a `ReevaluatePositionsStrategy`.

Thinking about it, this particular strategy of `allow_opposite_bets = False` seems like something that makes sense in combination with the `MaxAccuracyBettingStrategy`, but not necessarily in the general case.

e.g. if using `KellyBettingStrategy`, you are more interested in whether the distance between estimated p_yes and market p_yes has changed.

gabrielfior: As discussed on Slack, next steps here are
- Remove ManifoldAgentMarket `liquidate_positions`
- IN the infra_repo, remove `manifold` deployments
- Close this comment

kongzii: plus move `allow_opposite_bets` to class level

gabrielfior: Infra PR
https://github.com/gnosis/gnosisai-infrastructure/pull/94/files

gabrielfior: Removed ManifoldAgentMarket liquidate_positions in latest commit

prediction_market_agent_tooling/markets/omen/omen.py
gabrielfior: > After we sell it the first time, on the second run, what will happen? Will it be 0? And if it's 0, will the sell_tokens function work with 0?

Second run (see [link](https://github.com/gnosis/prediction-market-agent-tooling/pull/374/files#diff-a13882b8a216695b763e6c4f08eedd589ab372b59b4d795548584b0e5a21715dR171-R174)) - it won't be 0 since there are some leftovers (I reference this on the tests I wrote - for ex, I had approx 8 outcome tokens, after liquidating I still had approx 1.1e-5 tokens), hence position will be retrieved again and sold again.
So to answer your question -> yes, it works for multiple runs.

> Is Hugo accounting for the possibility of sold tokens?
That's a good question - let's track this in an extra ticket.

kongzii: > hence position will be retrieved again and sold again.

Won't that break at some point? I guess we can not sell to infinity?

prediction_market_agent_tooling/markets/manifold/manifold.py
gabrielfior: Added `allow_opposite_bets` as class variable

prediction_market_agent_tooling/markets/omen/omen.py
gabrielfior: I tried selling twice, it works (balance doesn't change much).
The logic is here (https://github.com/gnosis/conditional-tokens-market-makers/blob/master/contracts/FixedProductMarketMaker.sol#L321-L334), my understanding is that there is nothing stopping you from selling 0 (or a tiny amount) multiple times.

One could add a check (e.g. `sell if outcome_token_amount >= 1e-3`) or similar to avoid this behavior.

evangriffiths: `AgentMarket.get_positions` has `larger_than` arg already, for this exact reason!

evangriffiths: > Is Hugo accounting for the possibility of sold tokens? I'd say that if the agent sells his previous bets, they shouldn't be calculated into its accuracy. (our monitoring app in PMAT probably needs to changed as well)

This is a good point! But if you think of the "betting accuracy" metric as a proxy for "prediction accuracy" I think it's actually okay to just carry on measuring it how we are, and ignore the fact that we are selling positions. Selling positions is part of the trading strategy, and the performance of that is measured by ROI/profitability, and so is measured separately.

gabrielfior: > AgentMarket.get_positions has larger_than arg already, for this exact reason!

That's what I had in mind when I said `check` - adding this here and setting the value of 1e-2 xDAI as filter. Let me know if that should be changed.

kongzii: Isn't that too big? 0.01 xDai, TT was normally betting 0.001, for example 😄 tiny bet is defined as `0.00001`. 

So maybe like 1e-8 xDai?

evangriffiths: once we store predictions separately, so have proper separation between "prediction accuracy" and "profitability", this will make more sense. i.e. hugo will be taking accuracy stats from the stored predictions, and there won't be any knowledge of how the agent has used that prediction to trade (and that's okay, I think! 😅)

gabrielfior: @evangriffiths see latest commit, added a filter using `larger_than` as discussed
https://github.com/gnosis/prediction-market-agent-tooling/pull/374/commits/f9b9f348e51d499238f063b9ef823661ba0544c7

kongzii: > hugo will be taking accuracy stats from the stored predictions

But from the stored prediction, he can't say if the bet was sold or not.

Let's say today we predict that GPT-5 won't be released in 3 days, because there is no information indicating it.
But tomorrow, Sam will announce the GPT-5 release the next day. So, the agent picks it up and changes the answer.

Should the first (now sold) bet be taken as incorrect?

gabrielfior: > So maybe like 1e-8 xDai?

Realistically speaking, we simply want to exclude positions that were previously liquidated (note that the `redeemPositions` will anyways extract any leftover).
I did more tests and I find that the leftovers are approx equal to `bet_amount * 1e-5` (true for `bet_amount` [1000,1,1e-5] xDAI).
So this could be handled in the DeployableTraderAgent like below:
```
# agent.py
if not self.allow_opposite_bets:
#[...]
# If we have an existing position, sell it first if they bet on a different outcome.
bet_amount = self.strategy.bet_amount  # This requires modification of BettingStrategy class, adding bet_amount as property both to MaxAccuracy and Kelly.
market.liquidate_existing_positions(amount_and_direction.direction, larger_than = bet_amount * 1e-4)
```

evangriffiths: > But from the stored prediction, he can't say if the bet was sold or not.
> ...
> Should the first (now sold) bet be taken as incorrect?

Hmm I would say the question is the same regardless of whether we sell the position that came from an earlier prediction or not. At the moment, the accuracy metric is a function of a few things:

1. the quality of the agent's research, and prediction based on the research
2. when the agent bets (i.e. more likely to have better accuracy if it bets closer to market close time)

Ideally we'd just want it to be a measure of (1). But I don't think we should add a third variable - whether the agent decides to sell the position, which IMO is based on betting strategy - into the mix.

kongzii: That's a valid point 🤔 🤔 Then let's keep it as it is, and we can change it if needed in the future.

evangriffiths (approved):

PR #375, opened 2024-09-03
Chec if agent can pay for gas

kongzii:


prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -337,18 +341,32 @@ def update_langfuse_trace_by_processed_market(
...
-                f"for agent with address {api_keys.public_key} is not met."
-            )
+        if market_type == MarketType.OMEN:
+            if check_for_gas and not is_minimum_required_balance(
+                api_keys.public_key,
+                min_required_balance=xdai_type(0.001),
+                sum_wxdai=False,
+            ):
+                raise CantPayForGasError(
+                    f"{api_keys.public_key=} doesn't have enough xDai to pay for gas."
```
kongzii, line 360: If `public_key` (agent's wallet, not safe) doesn't have at least something for paying for gas, it can not redeem winnings and is stuck forever. So we need to differentiate such errors to ping us to send him at least something.

evangriffiths (approved):

PR #376, opened 2024-09-03
More filters for get_responses

kongzii:


prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -608,21 +609,26 @@ def get_resolved_bets_with_valid_answer(
         )
         return [b for b in bets if b.fpmm.is_resolved_with_valid_answer]
 
-    def get_questions(
-        self,
+    @staticmethod
+    def apply_question_filters(
+        where_stms: dict[str, t.Any] | None = None,
```
evangriffiths, line 614: doesn't look like the `where_stms` arg is used anywhere when calling `apply_question_filters`

kongzii: True! Removed.

PR #379, opened 2024-09-04
Fix Safe integration via AccessList

kongzii:


prediction_market_agent_tooling/tools/web3_utils.py
```diff
@@ -211,20 +216,46 @@ def send_function_on_contract_tx_using_safe(
     tx_params: Optional[TxParams] = None,
     timeout: int = 180,
 ) -> TxReceipt:
+    if not web3.provider.endpoint_uri:  # type: ignore
+        raise EnvironmentError(f"RPC_URL not available in web3 object.")
+    ethereum_client = EthereumClient(ethereum_node_url=URI(web3.provider.endpoint_uri))  # type: ignore
+    s = Safe(safe_address, ethereum_client)  # type: ignore
+    safe_master_copy_address = s.retrieve_master_copy_address()
+    eoa_public_key = private_key_to_public_key(from_private_key)
+    # See https://ethereum.stackexchange.com/questions/123750/how-to-implement-eip-2930-access-list for details,
```
kongzii, line 225: Poor Challenger wasn't withdrawing from Reality contract because of this

prediction_market_agent_tooling/tools/web3_utils.py
```diff
@@ -211,20 +216,46 @@ def send_function_on_contract_tx_using_safe(
...
+        ]
+    )
     tx_params = prepare_tx(
         web3=web3,
         contract_address=contract_address,
         contract_abi=contract_abi,
         from_address=safe_address,
         function_name=function_name,
         function_params=function_params,
+        access_list=access_list,
```
gabrielfior, line 256: If the safe wasn't working for a particular type of contract interaction, I think we should have a chain test for that, to avoid having that problem in the feature ([see existing tests](https://github.com/gnosis/prediction-market-agent-tooling/blob/2d63918eb9756223329383bf311235de06f6b4f5/tests_integration_with_local_chain/safe/test_safe.py#L46)).
So I would suggest either
a. add a test in this PR or
b. create a new ticket for writing a test for this

Any option above is fine by me

prediction_market_agent_tooling/tools/web3_utils.py
gabrielfior: Great. Added an issue to their repo (https://github.com/safe-global/safe-eth-py/issues/1320).

gabrielfior (approved):

prediction_market_agent_tooling/tools/web3_utils.py
kongzii: Good point, I extended the erc20 test.

With accessList it works, if I comment out the accessList, it fails.

PR #386, opened 2024-09-06
Add outcome_token_pool to AgentMarket

evangriffiths:
Motivation: we wan't to place large bet sizes (multiples of market liquidity size), and therefore need to know the token pool sizes in order to calculate a 'market moving' bet, or an accurate kelly bet


prediction_market_agent_tooling/markets/polymarket/polymarket.py
```diff
@@ -39,6 +39,7 @@ def from_data_model(model: PolymarketMarketWithPrices) -> "PolymarketAgentMarket
             close_time=model.end_date_iso,
             url=model.url,
             volume=None,
+            outcome_token_pool=None,
```
evangriffiths, line 42: @gabrielfior unfortunately polymarket don't give market liquidity in the API (https://docs.polymarket.com/#clob-api) either 😢

gabrielfior: They don't have one, they use an orderbook behind the scenes 😢

kongzii (approved):

PR #387, opened 2024-09-06
383 move rebalancing logic involving previous positions to strategy

gabrielfior:
-> Exposing method `calculate_trades` from `BettingStrategy`, so that agent gets a list of trades and executes these.
-> Implemented rebalance with lower gas fees (don't necessarily sell, buy only diff amount, etc)


prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -422,6 +408,19 @@ def before_process_market(
     ) -> None:
         self.update_langfuse_trace_by_market(market_type, market)
 
+    @staticmethod
+    def get_existing_position_for_market(
+        market: AgentMarket, user_id: str
+    ) -> Position | None:
+        minimum_liquidatable_amount = market.get_tiny_bet_amount().amount / 10
```
kongzii, line 415: I think this `tiny_bet / 10` logic was used somewhere already. What about refactoring it to a common property on `AgentMarket` ?

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -442,33 +441,23 @@ def process_market(
             self.update_langfuse_trace_by_processed_market(market_type, None)
             return None
 
-        amount_and_direction = self.calculate_bet_amount_and_direction(answer, market)
+        existing_position = self.get_existing_position_for_market(
+            market=market, user_id=APIKeys().public_key
```
kongzii, line 445: Shouldn't it be `bet_from_address`?

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -442,33 +441,23 @@ def process_market(
...
-                    amount=amount_and_direction.amount,
-                    currency=amount_and_direction.currency,
-                ),
-                outcome=amount_and_direction.direction,
-            )
+            for trade in trades:
+                logger.info(f"Executing trade {trade}")
+                if trade.trade_type == TradeType.SELL:
+                    market.sell_tokens(outcome=trade.outcome, amount=trade.amount)
+                else:
```
kongzii, line 454: Can this be `elif trade.trade_type == TradeType.BUY` and then `else` with raise ValueError?

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -442,33 +441,23 @@ def process_market(
...
+                    market.buy_tokens(outcome=trade.outcome, amount=trade.amount)
 
         self.after_process_market(market_type, market)
 
         processed_market = ProcessedMarket(
             answer=answer,
-            amount=amount_and_direction,
```
kongzii, line 471: Can you add `trades` to this model instead? To keep Langfuse practical.

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -422,6 +408,19 @@ def before_process_market(
     ) -> None:
         self.update_langfuse_trace_by_market(market_type, market)
 
+    @staticmethod
+    def get_existing_position_for_market(
```
kongzii, line 412: Wdyt about moving this to AgentMarket? Feels odd that "get all positions" is defined on Market, but "get single position" is defined on Agent.

prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -2,61 +2,153 @@
...
+                amount=TokenAmount(amount=diff_amount, currency=market.currency),
+                outcome=outcome_bool,
+                trade_type=trade_type,
+            )
+
+            trades.append(trade)
+
+        # Sort inplace with SELL last
+        trades.sort(key=lambda t: t.trade_type == TradeType.SELL)
+        BettingStrategy._assert_trades_currency_match_markets(market, trades)
```
evangriffiths, line 84: I think this should be moved to just after `calculate_trades` is called in `process_market`, so that it runs even if the betting strategy doesn't call `_build_rebalance_trades_from_positions`

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -442,33 +441,23 @@ def process_market(
...
-                market.liquidate_existing_positions(amount_and_direction.direction)
-
-            market.place_bet(
-                amount=TokenAmount(
-                    amount=amount_and_direction.amount,
-                    currency=amount_and_direction.currency,
-                ),
-                outcome=amount_and_direction.direction,
-            )
+            for trade in trades:
```
evangriffiths, line 450: Shall we have some logic to handle the case where a trade is a nop (e.g. `trade = Trade(outcome=0, amount=10, trade_type=TradeType.BUY)`)

evangriffiths (commented):
Currently if we do

```
market0 = OmenAgentMarket(id=...)
agent.process_market(market=market0) # once
agent.process_market(market=market0) # twice
```

current behaviour (assuming the prediction made in each case is the same):

- `MaxAccuracyBettingStrategy` - place `bet_amount` both times
- `KellyBettingStrategy` - place x <= `max_bet` both times

new behaviour:

- `MaxAccuracyBettingStrategy`
  - place `bet_amount` first time
  - place no bet the second time
- `KellyBettingStrategy`
  - place x <= `max_bet` first time
  - place no bet the second time

Is this what you intended? I would have thought for Kelly at least it would be:
  - place x <= `max_bet` first time
  - place x <= y <= `max_bet` second time

for `MaxAccuracyBettingStrategy` I'm not sure.

prediction_market_agent_tooling/deploy/agent.py
gabrielfior: I wonder if TradeType can be None (mypy would catch that, right?). If yes, then happy to raise ValueError.
I imagined that, since property is Enum, only enum values would be "allowed".

prediction_market_agent_tooling/deploy/agent.py
gabrielfior: > case where a trade is a nop

Sorry, which case is that? If amount = 0, no trade is generated.

prediction_market_agent_tooling/deploy/agent.py
kongzii: It's just about future-proofing, imagine you have

```
class E(Enum):
 a = a
 b = b

if my_variable == E.a:
  ...
else:
 ...
```

And then, for whatever reason, someone extends the `E` to

```
class E(Enum):
 a = a
 b = b
 c = c
```

Now, you have places in the code where default behavior is happening unexpectedly.

gabrielfior: Implemented with `match` as discussed

prediction_market_agent_tooling/deploy/agent.py
gabrielfior: Added to the DeployableTraderAgent as you suggested

prediction_market_agent_tooling/deploy/betting_strategy.py
gabrielfior: @evangriffiths can you please check if the latest commit (https://github.com/gnosis/prediction-market-agent-tooling/pull/387/commits/0b8a0040fb67eb0bbe40a9be4eaa1fd2cf3082d6#diff-ebe828b1c77960ecaea62858ada5f1ee0b662b19051aa2ffbf92cd6ea6287110R100) solves this?

prediction_market_agent_tooling/markets/agent_market.py
```diff
@@ -230,6 +239,9 @@ def get_outcome_index(self, outcome: str) -> int:
     def get_token_balance(self, user_id: str, outcome: str) -> TokenAmount:
         raise NotImplementedError("Subclasses must implement this method")
 
+    def get_existing_position_for_market(self, api_keys: APIKeys) -> Position | None:
```
evangriffiths, line 242: Looks like this only needs `user_id: str`, like `get_positions`. And could rename to just `get_position`.

prediction_market_agent_tooling/deploy/betting_strategy.py
evangriffiths: Yes looks good 👍

prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -2,61 +2,193 @@
...
+
 
-    def calculate_bet_amount_and_direction(
-        self, answer: ProbabilisticAnswer, market: AgentMarket
-    ) -> TokenAmountAndDirection:
+"""aaa
+Option 1 - add Mixin with adjust_bet_amount or similar
+Option 2 - rename max_bet_amount and bet_amount -> bet_amount
+Option 3 - whatever, do everything I can in the base class
+"""
```
evangriffiths, line 97: Old comment, can be removed

```suggestion
```

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -152,22 +152,30 @@ def get_liquidity(self) -> TokenAmount:
     def get_tiny_bet_amount(cls) -> BetAmount:
         return BetAmount(amount=OMEN_TINY_BET_AMOUNT, currency=cls.currency)
 
+    @classmethod
+    def get_liquidatable_amount(cls) -> BetAmount:
+        tiny_amount = cls.get_tiny_bet_amount()
+        tiny_amount.amount = tiny_amount.amount / 10
+        return tiny_amount
+
```
evangriffiths, line 160: This is the same as the method on the base class, so not needed?

```suggestion
```

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -432,6 +445,19 @@ def get_token_balance(
             currency=self.currency,
         )
 
+    def get_existing_position_for_market(self, api_keys: APIKeys) -> Position | None:
```
evangriffiths, line 448: Nothing in this looks omen-specific, so it could be moved to the base class

prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -2,61 +2,193 @@
...
-            amount=bet_amount,
-            currency=market.currency,
-            direction=direction,
+
+        amounts = {
+            market.get_outcome_str_from_bool(direction): TokenAmount(
+                amount=adjusted_bet_amount,
+                currency=market.currency,
+            ),
+            market.get_outcome_str_from_bool(not direction): TokenAmount(
```
evangriffiths, line 132: Do you need to specify these 0 TokenAmounts when instantiating the target position (here, and in  KellyBettingStrategy.calculate_trades)

prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -2,61 +2,193 @@
...
+        market: AgentMarket,
+    ) -> list[Trade]:
+        """
+        This helper method builds trades by rebalancing token allocations to each outcome.
+        For example, if we have an existing position with 10 tokens in outcome 0 and 5 in outcome 1,
+        and our target position is 20 tokens in outcome 0 and 0 in outcome 1, we would return these trades:
+        trades = [
+            Trade(outcome=0, amount=10, trade_type=TradeType.BUY),
+            Trade(outcome=1, amount=5, trade_type=TradeType.SELL)
+            ]
```
evangriffiths, line 56:
```suggestion
        trades = [
            Trade(outcome=0, amount=10, trade_type=TradeType.BUY),
            Trade(outcome=1, amount=5, trade_type=TradeType.SELL)
        ]
```

prediction_market_agent_tooling/markets/omen/omen.py
gabrielfior: this is not present in the updated version

prediction_market_agent_tooling/deploy/betting_strategy.py
evangriffiths: sorry, i mean why not just:
```
        amounts = {
            market.get_outcome_str_from_bool(direction): TokenAmount(
                amount=adjusted_bet_amount,
                currency=market.currency,
            ),
        }
```

gabrielfior: Good point, we can simply not create an amount for the other direction. This will get caught on `_build_rebalance_trades_from_positions`.

evangriffiths (approved):

PR #391, opened 2024-09-10
Fix getting of asset token on Gnosis chain

kongzii:


prediction_market_agent_tooling/tools/contract.py
```diff
@@ -417,6 +417,11 @@ class ContractERC4626OnGnosisChain(
     ERC-4626 standard base class with Gnosis Chain configuration.
     """
 
+    def get_asset_token_contract(
+        self, web3: Web3 | None = None
+    ) -> ContractERC20OnGnosisChain | ContractDepositableWrapperERC20OnGnosisChain:
+        return to_gnosis_chain_contract(super().get_asset_token_contract(web3=web3))
```
kongzii, line 423: Without this, the returned contract doesn't have `CHAIN_RPC_URL` property, and then it fails if the web3 argument isn't provided on its methods.

kongzii: Which is interesting that it was working during the local test, as web3 was provided by local chain

prediction_market_agent_tooling/tools/contract.py
```diff
@@ -351,7 +351,7 @@ def get_asset_token_contract(
         web3 = web3 or self.get_web3()
         contract = init_collateral_token_contract(self.asset(), web3=web3)
         assert not isinstance(
-            contract, ContractERC4626OnGnosisChain
+            contract, ContractERC4626BaseClass
```
gabrielfior, line 354: Why don't we assert that the type match the returned types?
e.g. isinstance(contract, ContractERC20BaseClass) or isistance(type, ContractDepositableWrapperERC20BaseClass)?

gabrielfior:
Would it be possible for us to simplify all the contract classes (currently 10) that we have in `tools/contract.py`? I think that adds complexity that we could avoid.

gabrielfior (approved):

PR #393, opened 2024-09-10
Fix bet amount calculation in sDai markets

kongzii:


tests_integration_with_local_chain/markets/omen/test_omen.py
```diff
@@ -336,20 +355,8 @@ def test_place_bet_with_autodeposit(
     assert initial_balances.wxdai == xdai_type(0)
     assert initial_balances.xdai > xdai_type(0)
 
-    # Convert half of the xDai to wxDai
-    collateral_token_contract.deposit(
```
kongzii, line 340: Below, `place_bet` is called with `omen_auto_deposit=True`, so this depositing here isn't required.

PR #396, opened 2024-09-10
hotfix - trade amounts now correctly calculated

gabrielfior:
-> Fixes the way trade amount is calculated (prev. wrong sign)
-> Added unit test


prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -78,7 +78,7 @@ def _build_rebalance_trades_from_positions(
 
             if prev_amount.currency != new_amount.currency:
                 raise ValueError("Cannot handle positions with different currencies")
-            diff_amount = prev_amount.amount - new_amount.amount
+            diff_amount = new_amount.amount - prev_amount.amount
```
gabrielfior, line 81: This fixes the error we saw in production of agents selling positions they didn't own.

pyproject.toml
```diff
@@ -20,7 +20,7 @@ cron-validator = "^1.0.8"
 pydantic = "^2.6.1"
 web3 = "^6.15.1"
 eth-typing = "^3.0.0"
-pydantic-settings = "^2.1.0"
+pydantic-settings = ">=2.4.0,<2.5" #eth-ape limit
```
gabrielfior, line 23: See https://github.com/ApeWorX/ape/blob/203e63d9383ffb9260d91fa4aa20828497b38b7a/setup.py#L113

evangriffiths (approved):

pyproject.toml
kongzii: `>=2.4.0,<2.5` is equal to `^2.4.0` (no big deal to keep it like this, it just makes updates in the future a bit less wordy 😄 )

tests/test_betting_strategy.py
```diff
@@ -18,3 +29,32 @@ def test_answer_decision(
...
+    mock_answer = ProbabilisticAnswer(p_yes=Probability(0.9), confidence=0.5)
+    mock_market = Mock(OmenAgentMarket, wraps=OmenAgentMarket)
+    mock_market.get_tiny_bet_amount.return_value = tiny_amount
+    mock_market.current_p_yes = 0.5
+    mock_market.currency = Currency.xDai
+    mock_market.id = "0x123"
+
+    trades = strategy.calculate_trades(mock_existing_position, mock_answer, mock_market)
+    # assert 1 buy trade and 1 sell trade
+    buy_trade = next(iter([t for t in trades if t.trade_type == TradeType.BUY]))
```
kongzii, line 55: This doesn't assert there is 1 buy trade and 1 sell trade. There could be potentially many of them, which would be a bug.

```
>>> a = ["buy", "buy", "buy", "sell"]
>>> next(iter(x for x in a if x == "buy"))
'buy'
>>>
```

Buy and Sell trades are ordered, right? So you can do just

```
assert len(trades) == 2
buy_trade = trades[0]
sell_trade = trades[1]
```

(or flipped indexes, I don't remember the sorting there)

tests/test_betting_strategy.py
```diff
@@ -18,3 +29,32 @@ def test_answer_decision(
...
+    assert buy_trade
```
kongzii, line 56: There would be an exception, if no buy trade is found, so this assert is unnecessary:

```
>>> next(iter([]))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
StopIteration
```

```suggestion
    assert buy_trade.trade_type == TradeType.BUY
```

tests/test_betting_strategy.py
```diff
@@ -18,3 +29,32 @@ def test_answer_decision(
...
+    assert buy_trade.amount.amount == (mock_amount.amount + tiny_amount.amount)
+    sell_trade = next(iter([t for t in trades if t.trade_type == TradeType.SELL]))
+    assert sell_trade
```
kongzii, line 59:
```suggestion
    assert sell_trade.trade_type == TradeType.SELL
```

Issue #401, opened 2024-09-12
Use sensible default to prevent size-zero bet on balanced markets using kelly


evangriffiths:
see this comment: https://github.com/gnosis/prediction-market-agent-tooling/pull/389/#discussion_r1754889940

PR #408, opened 2024-09-17
Fix kelly bet if token pool is exactly balanced

evangriffiths:
Fixes https://github.com/gnosis/prediction-market-agent-tooling/issues/401

### DIFF
--- a/prediction_market_agent_tooling/tools/betting_strategies/kelly_criterion.py
+++ b/prediction_market_agent_tooling/tools/betting_strategies/kelly_criterion.py
@@ -136,4 +136,4 @@ def get_kelly_bet_full(
     denominator = 2 * (x**2 * f - y**2 * f)
     kelly_bet_amount = numerator / denominator
 
-    return SimpleBet(direction=kelly_bet_amount > 0, size=abs(kelly_bet_amount))
+    return SimpleBet(direction=kelly_bet_amount > 0, size=min(max_bet, abs(kelly_bet_amount)))
--- a/prediction_market_agent_tooling/tools/betting_strategies/kelly_criterion.py
+++ b/prediction_market_agent_tooling/tools/betting_strategies/kelly_criterion.py
@@ -100,6 +100,10 @@ def get_kelly_bet_full(
     b = max_bet
     f = 1 - fee
 
+    if x == y:
+        # Add a delta to prevent division by zero
+        y += 1e-10
+
     numerator = (
         -4 * x**2 * y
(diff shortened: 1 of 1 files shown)


kongzii (approved):

prediction_market_agent_tooling/tools/betting_strategies/kelly_criterion.py
```diff
@@ -130,8 +134,6 @@ def get_kelly_bet_full(
         ** (1 / 2)
     )
     denominator = 2 * (x**2 * f - y**2 * f)
-    if denominator == 0:
-        return SimpleBet(direction=True, size=0)
     kelly_bet_amount = numerator / denominator
 
     return SimpleBet(direction=kelly_bet_amount > 0, size=abs(kelly_bet_amount))
```
kongzii, line 139:
```suggestion
    return SimpleBet(direction=kelly_bet_amount > 0, size=min(max_bet, abs(kelly_bet_amount)))
```

Should we have this here? A returned bet can be slightly higher than the maximum requested one, and it could also prevent some bugs in the future.

Screenshot by Dropbox Capture

evangriffiths: What were the inputs that give `kelly_bet_amount > max_bet`? I would say we should have an error in this case (and if we see it happen find+fix the bug)!

kongzii:
```
print(
    get_kelly_bet_full(
        **{
            "yes_outcome_pool_size": 7.0,
            "no_outcome_pool_size": 7.0,
            "estimated_p_yes": 1.0,
            "confidence": 1.0,
            "max_bet": 10,
            "fee": 0.0,
        }
    )
)
```

gives `size=10.000030451751737`. I thought such a small difference could happen given all the multiplications/divisions there are.

evangriffiths: ah okay, yeah fair point

ooo but I think it's because of the delta. e.g.

```
print(
    get_kelly_bet_full(
        **{
            "yes_outcome_pool_size": 8.0,
            "no_outcome_pool_size": 7.0,
            "estimated_p_yes": 1.0,
            "confidence": 1.0,
            "max_bet": 10,
            "fee": 0.0,
        }
    )
)
```
gives `10.0` (you'd it expect it to be more!)

PR #409, opened 2024-09-17
Getting jobs

kongzii:


prediction_market_agent_tooling/jobs/jobs.py
```diff
@@ -0,0 +1,19 @@
+from prediction_market_agent_tooling.gtypes import xDai
+from prediction_market_agent_tooling.jobs.jobs_models import OmenJob
+from prediction_market_agent_tooling.markets.omen.omen_subgraph_handler import (
+    FilterBy,
+    OmenSubgraphHandler,
+    SortBy,
+)
+
+JOBS_CATEGORY = "jobs"
```
kongzii, line 9: Suggestions for other names are welcome

prediction_market_agent_tooling/jobs/jobs_models.py
```diff
@@ -0,0 +1,72 @@
...
+    assert trade.outcome, "Should buy only YES on job markets."
+    assert (
+        trade.amount.currency == Currency.xDai
+    ), "Should work only on real-money markets."
+
+    bet_amount = xdai_type(trade.amount.amount)
+
+    outcome_tokens = market_contract.calcBuyAmount(
+        investment_amount=xdai_to_wei(bet_amount),
+        outcome_index=get_outcome_index(get_bet_outcome(trade.outcome)),
```
evangriffiths, line 66: Could use the existing helper methods on OmenMarket. Then you don't need to rely on the order of outcomes being the same as in the hardcoded `OMEN_BINARY_MARKET_OUTCOMES`

```suggestion
        outcome_index=market.yes_index() if trade.outcome else market.no_index()
```

prediction_market_agent_tooling/jobs/jobs_models.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def compute_job_reward(
+    market: OmenMarket, max_bond: xDai, web3: Web3 | None = None
+) -> xDai:
+    agent_market = OmenAgentMarket.from_data_model(market)
+    market_contract = agent_market.get_contract()
+
+    # Because jobs are powered by prediction markets, potentional reward depends on job's liquidity and our will to bond (bet) our xDai into our job completion.
+    required_trades = KellyBettingStrategy(max_bet_amount=max_bond).calculate_trades(
```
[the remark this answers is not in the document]
evangriffiths: It does seem like a flaw of using prediction markets for this job request/fulfilment pipeline that the reward isn't fixed (and that a poorer agent gets a smaller reward for doing the same job). But I can't think of a better way of doing it. And maybe something with a more economics-y brain will say that this is actually optimal!

prediction_market_agent_tooling/jobs/jobs_models.py
kongzii: Nice!

prediction_market_agent_tooling/jobs/jobs_models.py
```diff
@@ -0,0 +1,68 @@
...
+)
+from prediction_market_agent_tooling.gtypes import Probability, xDai, xdai_type
+from prediction_market_agent_tooling.markets.omen.omen import (
+    OmenAgentMarket,
+    OmenMarket,
+)
+from prediction_market_agent_tooling.tools.web3_utils import wei_to_xdai, xdai_to_wei
+
+
+class OmenJob(BaseModel):
```
kongzii, line 20: I'm thinking if this shouldn't be just a subclass of `OmenMarket` or `OmenAgentMarket` 🤔 As job basically is market and has all of it properties

prediction_market_agent_tooling/markets/omen/data_models.py
```diff
@@ -48,6 +51,12 @@ def get_bet_outcome(binary_outcome: bool) -> str:
     return OMEN_TRUE_OUTCOME if binary_outcome else OMEN_FALSE_OUTCOME
 
 
+def get_outcome_index(outcome_str: str) -> OutcomeIndex:
```
evangriffiths, line 56: unused now, can delete

prediction_market_agent_tooling/jobs/jobs_models.py
evangriffiths: So far it doesn't seem like it needs to be market-type specific. i.e. `get_jobs` could take a `market_type` arg. It seems like we are heading in the direction of being more omen-specific, but is it worth keeping the option open for the job-related code?

prediction_market_agent_tooling/jobs/jobs_models.py
```diff
@@ -0,0 +1,68 @@
...
+    trade = required_trades[0]
+    assert trade.trade_type == TradeType.BUY, "Should only buy on job markets."
+    assert trade.outcome, "Should buy only YES on job markets."
+    assert (
+        trade.amount.currency == Currency.xDai
+    ), "Should work only on real-money markets."
+
+    bet_amount = xdai_type(trade.amount.amount)
+
+    outcome_tokens = market_contract.calcBuyAmount(
```
evangriffiths, line 104: Just a heads-up. After https://github.com/gnosis/prediction-market-agent-tooling/pull/415 you won't need the smart contract

prediction_market_agent_tooling/jobs/jobs_models.py
kongzii: Done, I guess we will see how it plays out in practice, some further modifications might be required. But this makes sense to me atm.

prediction_market_agent_tooling/jobs/jobs_models.py
```diff
@@ -0,0 +1,112 @@
...
+
+class Job(BaseModel):
+    id: str
+    job: str
+    reward: float
+    currency: str
+    deadline: datetime
+
+
+class OmenJobMarket(OmenAgentMarket):
```
kongzii, line 28: OmenJobMarket is subclass of OmenAgentMarket, because it is a market. But it's another class with extra properties, so we can distinguish it from "question markets"

prediction_market_agent_tooling/jobs/jobs_models.py
evangriffiths: but w.r.t. whether it should be a subclass, I don't have a strong opinion. It's nice for testing to be able to just construct a more minimal `OmenJob`, but can't think of much else :)

prediction_market_agent_tooling/jobs/jobs_models.py
```diff
@@ -0,0 +1,112 @@
...
+)
+from prediction_market_agent_tooling.gtypes import Probability, xDai, xdai_type
+from prediction_market_agent_tooling.markets.omen.omen import (
+    OmenAgentMarket,
+    OmenMarket,
+)
+from prediction_market_agent_tooling.tools.web3_utils import wei_to_xdai, xdai_to_wei
+
+
+class Job(BaseModel):
```
kongzii, line 20: `Job` is just a simple pydantic model that can be dumped into JSON and feeded into agent's prompt

prediction_market_agent_tooling/jobs/jobs_models.py
kongzii: > So far it doesn't seem like it needs to be market-type specific. i.e. get_jobs could take a market_type arg. It seems like we are heading in the direction of being more omen-specific, but is it worth keeping the option open for the job-related code?

Good point. Refactored, check it out please 🙏

evangriffiths (approved):

PR #414, opened 2024-09-18
Add some utils for getting traces from the langfuse client

evangriffiths:
Example usage:
```
api_keys = APIKeys()
assert api_keys.bet_from_address == Web3.to_checksum_address(
    "0xe7aa88a1d044e5c987ecce55ae8d2b562a41b72d"  # prophetgpt4
)
start_time = datetime(2024, 9, 13)
langfuse = Langfuse(
    secret_key=api_keys.langfuse_secret_key.get_secret_value(),
    public_key=api_keys.langfuse_public_key,
    host=api_keys.langfuse_host,
)

traces = get_traces_for_agent(
    agent_name="DeployablePredictionProphetGPT4TurboFinalAgent",
    from_timestamp=start_time,
    has_output=True,
    client=langfuse,
)

bets: list[ResolvedBet] = OmenAgentMarket.get_resolved_bets_made_since(
    better_address=api_keys.bet_from_address,
    start_time=start_time,
    end_time=None,
)


for bet in bets:
    trace = get_trace_for_bet(bet, traces)
    market = trace_to_omen_agent_market(trace)
    answer = trace_to_answer(trace)
    trades = trace_to_trades(trace)
```

### DIFF
--- a/prediction_market_agent_tooling/tools/langfuse_client_utils.py
+++ b/prediction_market_agent_tooling/tools/langfuse_client_utils.py
@@ -3,6 +3,7 @@ from datetime import datetime
 from langfuse import Langfuse
 from langfuse.client import TraceWithDetails
 
+from prediction_market_agent_tooling.loggers import logger
 from prediction_market_agent_tooling.markets.data_models import (
     ProbabilisticAnswer,
     ResolvedBet,
@@ -14,6 +15,7 @@ from prediction_market_agent_tooling.tools.utils import add_utc_timezone_validat
 
 def get_traces_for_agent(
     agent_name: str,
+    trace_name: str,
     from_timestamp: datetime,
     has_output: bool,
     client: Langfuse,
@@ -25,6 +27,7 @@ def get_traces_for_agent(
     all_agent_traces = []
     while True:
         traces = client.fetch_traces(
+            name=trace_name,
             limit=100,
             page=page,
             from_timestamp=from_timestamp,
@@ -76,7 +79,7 @@ def get_closest_datetime_from_list(
 
 def get_trace_for_bet(
     bet: ResolvedBet, traces: list[TraceWithDetails]
-) -> TraceWithDetails:
+) -> TraceWithDetails | None:
     # Get traces with the same market id
     traces_for_bet = [
         t for t in traces if trace_to_omen_agent_market(t).id == bet.market_id
@@ -88,6 +91,12 @@ def get_trace_for_bet(
         [t.timestamp for t in traces_for_bet],
     )
     # Sanity check - the trace should be after the bet
-    assert traces_for_bet[closest_trace_index].timestamp > bet.created_time
+    if traces_for_bet[closest_trace_index].timestamp < add_utc_timezone_validator(
+        bet.created_time
+    ):
+        logger.warning(
+            f"No trace for bet on market {bet.market_id} at time {bet.created_time} found"
+        )
+        return None
 
     return traces_for_bet[closest_trace_index]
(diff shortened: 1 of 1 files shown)


prediction_market_agent_tooling/tools/langfuse_client_utils.py
```diff
@@ -0,0 +1,93 @@
...
+            from_timestamp=from_timestamp,
+        )
+        if not traces.data:
+            break
+        page += 1
+
+        agent_traces = [
+            t
+            for t in traces.data
+            if t.session_id is not None and agent_name in t.session_id
```
kongzii, line 39: Hmm, shall we change session_id to just the agent's name? Or we could add agent's name as a tag to filter it by that.

Screenshot by Dropbox Capture

(not as part of this PR, just thinking that fetching all the traces won't be very efficient in future)

prediction_market_agent_tooling/tools/langfuse_client_utils.py
```diff
@@ -0,0 +1,93 @@
...
+from prediction_market_agent_tooling.markets.data_models import (
+    ProbabilisticAnswer,
+    ResolvedBet,
+    Trade,
+)
+from prediction_market_agent_tooling.markets.omen.omen import OmenAgentMarket
+from prediction_market_agent_tooling.tools.utils import add_utc_timezone_validator
+
+
+def get_traces_for_agent(
```
kongzii, line 15: Instead of getting traces, it might be more future-proof to get observations with name `process_market`, right now, it's the same thing (trace of trader agent === observation of process_market). But it could be changed in the future and one trace could contain many `process_market` observations: https://github.com/gnosis/prediction-market-agent/issues/444#issuecomment-2346131672.

However it works now, it's already implement and it's just "maybe" so good by me, just FYI.

prediction_market_agent_tooling/tools/langfuse_client_utils.py
kongzii: In UI, they allow to filter by session id startswith and similar fancy stuff, it didn't cross my mind to check if they allow that in Python SDK as well 😮‍💨

kongzii (approved):

prediction_market_agent_tooling/tools/langfuse_client_utils.py
evangriffiths: Yeah I was surprised by that too. I would have guessed that the sdk they expose would be the same as what they're using for the web backend, but I guess not!

> Or we could add agent's name as a tag to filter it by that.

I was going to suggest this! Will make a ticket

evangriffiths: Ticket for us here: https://github.com/gnosis/prediction-market-agent-tooling/issues/417

Also made an issue on the langfuse github https://github.com/orgs/langfuse/discussions/3395

prediction_market_agent_tooling/tools/langfuse_client_utils.py
evangriffiths: Just tried this (`client.get_observations(name="process_market", ...)`) but it returns an empty list - observations names match those of the spans (red), whereas `process_market` is the name of the trace.

So I've added the `trace_name="process_market"` arg. But atm the only traces containing agent names (e.g. `DeployablePredictionProphetGPT4TurboFinalAgent`) are already the "process_market" traces. But I guess as you say it's more future proof.

Screenshot 2024-09-18 at 15 26 13

kongzii: But both `trace` and `span` are created using `@observe()`! That behaviour doesn't make sense! 😄 

Thanks for trying and sorry for killing your time then 😢 

image

evangriffiths: No worries - it might be something I'm doing wrong. But I'll leave it for now if that's okay :)

PR #415, opened 2024-09-18
Calculate omen buy amount locally

evangriffiths:


prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -580,6 +580,23 @@ def get_user_url(cls, keys: APIKeys) -> str:
...
+            no_outcome_pool_size=outcome_token_pool[OMEN_FALSE_OUTCOME],
+            fee=self.fee,
+        )
+        return TokenAmount(amount=amount, currency=self.currency)
+
+    def _get_buy_token_amount_from_smart_contract(
+        self, bet_amount: BetAmount, direction: bool
+    ) -> TokenAmount:
+        """
+        Note: this is only valid if the market instance's token pool is
```
kongzii, line 597: Shouldn't this comment be on the other method?

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -1208,3 +1225,36 @@ def withdraw_wxdai_to_xdai_to_keep_balance(
...
+    yes_outcome_pool_size: float,
+    no_outcome_pool_size: float,
+    fee: float,
+) -> float:
+    """
+    Calculates the amount of outcome tokens received for a given investment
+
+    Taken from https://github.com/gnosis/conditional-tokens-market-makers/blob/6814c0247c745680bb13298d4f0dd7f5b574d0db/contracts/FixedProductMarketMaker.sol#L264
+    """
+    outcome_index = float(buy_direction)  # Convert boolean to index (0 or 1)
```
kongzii, line 1242: It's sad but markets have `index 0 == yes/true`, `index 1 == no/false`.

Is this logic here correct?

tests/markets/omen/test_omen.py
```diff
@@ -243,3 +243,28 @@ def test_get_new_p_yes() -> None:
...
+
+
+def test_get_buy_token_amount() -> None:
+    """
+    Test that the two methods of calculating buy amount are equivalent for a
+    'live' market (i.e. where the token pool matches that of the current smart
+    contract state)
+    """
+    market = OmenAgentMarket.get_binary_markets(
+        limit=1,
```
kongzii, line 255: Instead of 1, could this be at least `10` and also iterate over buy_direction True or False? Just feels like this calculation could have some edge cases

prediction_market_agent_tooling/markets/omen/omen.py
evangriffiths: Oops, these variables were unused - removed

prediction_market_agent_tooling/markets/omen/omen.py
evangriffiths: Yes, switched!

tests/markets/omen/test_omen.py
evangriffiths: fair point. I tested with a larger number locally, but thought it was quite slow, but actually there's a larger fixed overhead and it's not much slower to make the extra smart contract calls (5.2s for 1 vs 7.3s for 10)

kongzii (approved):

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -1208,3 +1225,33 @@ def withdraw_wxdai_to_xdai_to_keep_balance(
     logger.info(
         f"Withdrew {need_to_withdraw} wxDai to keep the balance above the minimum required balance {min_required_balance}."
     )
+
+
+def get_buy_token_amount(
+    investment_amount: float,
+    buy_direction: bool,
```
gabrielfior, line 1232: Tiny ask - can we rename this to `outcome_to_be_bought` or similar? Anything with the word `outcome` would suffice for me.

PR #421, opened 2024-09-19
Remove timestamp check when matching bets with traces

evangriffiths:


prediction_market_agent_tooling/tools/langfuse_client_utils.py
```diff
@@ -150,10 +150,5 @@ def get_trace_for_bet(
             add_utc_timezone_validator(bet.created_time),
             [t.timestamp for t in traces_for_bet],
         )
-        # Sanity check - the trace should be after the bet
```
evangriffiths, line 153: This was an incorrect assertion. Removing for now - a more robust way of matching will come from https://github.com/gnosis/prediction-market-agent-tooling/issues/422

kongzii (approved):

PR #425, opened 2024-09-23
Add ability to specify bet amount dynamically

evangriffiths:


prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -282,7 +282,7 @@ class DeployableTraderAgent(DeployableAgent):
     bet_on_n_markets_per_run: int = 1
     min_required_balance_to_operate: xDai | None = xdai_type(1)
     min_balance_to_keep_in_native_currency: xDai | None = xdai_type(0.1)
-    strategy: BettingStrategy = MaxAccuracyBettingStrategy()
+    strategy: BettingStrategy
```
kongzii, line 285: Do we need `strategy` as class' an attribute, if `get_betting_strategy` is used in the end?

prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -108,13 +112,10 @@ def adjust_bet_amount(
         bet_amount = (
             market.get_tiny_bet_amount().amount
             if self.bet_amount is None
-            else self.bet_amount
+            else self.bet_amount.get()
```
gabrielfior, line 115: What's this `get()`?

prediction_market_agent_tooling/tools/amount.py
```diff
@@ -0,0 +1,27 @@
...
+        return self.amount
+
+
+class DynamicAmount(Amount):
+    def __init__(self, get_amount_fn: Callable[[], float], proportion: float = 1.0):
+        self.get_wallet_balance_fn = get_amount_fn
+        self.proportion = proportion
+
+    def get(self) -> float:
+        return self.get_wallet_balance_fn() * self.proportion
```
gabrielfior, line 27: In my head, this feels like an over-kill - it's simply a `float` container. Surely it's nice that the `get_wallet_balance_fn` can be defined here, but still having an extra class for this instead of a simple class property (`bet_amount`) feels a bit too much.
@evangriffiths can you provide more context into why this is needed?

prediction_market_agent_tooling/tools/amount.py
```diff
@@ -0,0 +1,27 @@
...
+        return self.amount
+
+
+class DynamicAmount(Amount):
+    def __init__(self, get_amount_fn: Callable[[], float], proportion: float = 1.0):
+        self.get_amount_fn = get_amount_fn
+        self.proportion = proportion
+
+    def get(self) -> float:
+        return self.get_amount_fn() * self.proportion
```
gabrielfior, line 27: I feel this is a bit of over-kill -> in theory, the amount should simply be a 'float', that can either be updated or not.
This update logic is part of the agent mechanism itself, for example, as part of the 'DeployableTraderAgent'. I don´t see a reason for storing this logic here - @evangriffiths can you provide more context as to why this here is needed?

prediction_market_agent_tooling/tools/amount.py
evangriffiths: In main, the amount is a fixed property of the `BettingStrategy`. imo you should also be able to specify at the BettingStrategy level whether you want the 'amount' to be static vs dynamic. 

Are you suggesting keeping it the way it is on `main`, and having some logic in a DeployableTraderAgent method to update this amount based on the agent's balance?

Something like changing `DeployableTraderAgent.build_trades` to:

```python
    def build_trades(
        self,
        market_type: MarketType,
        market: AgentMarket,
        answer: ProbabilisticAnswer,
        existing_position: Position | None,
    ) -> list[Trade]:
        self.betting_strategy.update(wallet_balance=get_wallet_balance(market_type=market_type)) # new logic!!
        trades = self.betting_strategy.calculate_trades(existing_position, answer, market)
        BettingStrategy.assert_trades_currency_match_markets(market, trades)
        return trades
```

?

That would come with its own messiness. e.g. for the KnownOutcomeAgent case you'd have to subclass the KellyBettingStrategy to change the `update` to use the `min(balance * 0.5, 25)` logic.

prediction_market_agent_tooling/deploy/agent.py
gabrielfior: > Yes, I agree. But if the strategy is going to be able to dynamically get the wallet balance, it requires market_type, which means it can't be set as a class variable.

I can see a way where market_type is not required inside the strategy.

The problem here is `how can we update the bet amount dynamically?`For that reason, there is the abstract function `adjust_bet_amount` -> https://github.com/gnosis/prediction-market-agent-tooling/blob/d5766e5f3709b8fdcf65a93c522824e7946be9aa/prediction_market_agent_tooling/deploy/betting_strategy.py#L34-L37

This way, you could do e.g.
```
# DeployableTraderAgent
# [....]
    def build_trades(
        self,
        market: AgentMarket,
        answer: ProbabilisticAnswer,
        existing_position: Position | None,
    ) -> list[Trade]:
       # get balance from wallet
       balance = get_balance_from_wallet(market)
       self.strategy.adjust_bet_amount(balance)
       # proceed as before
        trades = self.strategy.calculate_trades(existing_position, answer, market)
        BettingStrategy.assert_trades_currency_match_markets(market, trades)
        return trades
```

@evangriffiths Do you think the construction above makes sense?

prediction_market_agent_tooling/tools/amount.py
gabrielfior: > BettingStrategy level whether you want the 'amount' to be static vs dynamic.

I would dispute that. The Kelly betting strategy (as an example) should simply get a `max_bet_amount` as arg and return a `kelly_bet_amount` as output.

Your code example is roughly what I wroteup in the other thread. I think this is preferred since we have all the required variables (market) and functions (get_wallet_balance) in-place, vs creating more stuff inside strategy. That way, Strategy remains small and modular and (relatively) not complex.

But if you have a different approach, happy to see some code and discuss it.

prediction_market_agent_tooling/deploy/agent.py
evangriffiths: 😄 funny timing... continuing in the thread [below](https://github.com/gnosis/prediction-market-agent-tooling/pull/425#discussion_r1771353295)

prediction_market_agent_tooling/tools/amount.py
evangriffiths: I get what you're saying about putting the dynamic stuff in `adjust_bet_amount` to be consistent. But I'm not sure how well it fits...

What do you have in mind for the method's implementation in this case? e.g. for kelly, something like:

```python
class KellyBettingStrategy(BettingStrategy):
    def adjust_bet_amount(
        self,
        existing_position: Position | None,
        market: AgentMarket,
        agent_balance: float, # new arg
    ) -> float:
        existing_position_total_amount = (
            existing_position.total_amount.amount if existing_position else 0
        )
        if self.update_max_bet_from_wallet_balance: # New boolean property on KellyBettingStrategy
            self.max_bet_amount = agent_balance
        return self.max_bet_amount + existing_position_total_amount
```

?

With this approach, as mentioned above, for the desired behaviour you'd have to make a custom betting strategy for the KnownOutcomeAgent, instead of using existing KellyBettingStrategy.

Another question. How would initialising a strategy look with this approach? e.g.
```python
s = KellyBettingStrategy(max_bet_amount=10000.0, update_max_bet_from_wallet_balance=True)`
```

It feels a bit weird having these as separate fields, since `max_bet_amount` won't ever be used if `update_max_bet_from_wallet_balance == True`

gabrielfior: Agree that it feels weird to have separate fields - in my head, `KellyBettingStrategy` has no idea about wallet balances, only about `max_bet_amount` - that's why I'm pushing for this logic to be handled elsewhere.
Also,
> if self.update_max_bet_from_wallet_balance: # New boolean property on KellyBettingStrategy
            self.max_bet_amount = agent_balance

To me, what's where the complexity originates from - this update should happen outside of the betting strategy.

Let's huddle and discuss about this.

evangriffiths: > in my head, KellyBettingStrategy has no idea about wallet balances, only about max_bet_amount

That's the case in this PR currently! The DynamicAmount takes some arbitrary getter function, which is defined at the agent level.

evangriffiths:
closing in favour of https://github.com/gnosis/prediction-market-agent-tooling/pull/430

PR #428, opened 2024-09-23
Add tool to compute user's accuracy on Reality.eth

kongzii:


prediction_market_agent_tooling/tools/omen/reality_accuracy.py
```diff
@@ -0,0 +1,74 @@
...
+    )
+    unique_question_ids = set(r.question.questionId for r in user_responses)
+
+    # Get all responses for these questions (including not ours)
+    question_to_responses = {
+        question_id: OmenSubgraphHandler().get_responses(
+            limit=None, question_id=question_id
+        )
+        for question_id in unique_question_ids
+    }
```
gabrielfior, line 41: As discussed in other thread - move `OmenSubgraphHandler()` outside of the loop for now.

prediction_market_agent_tooling/tools/omen/reality_accuracy.py
```diff
@@ -0,0 +1,74 @@
...
+
+    total = 0
+    correct = 0
+
+    for question_id, responses in question_to_responses.items():
+        is_correct = user_was_correct(user, responses)
+        assert (
+            is_correct is not None
+        ), f"All these questions should be challenged by provded user: {responses[0].question.url}"
```
gabrielfior, line 50:
```suggestion
        ), f"All these questions should be challenged by provided user: {responses[0].question.url}"
```

prediction_market_agent_tooling/tools/omen/reality_accuracy.py
kongzii: No need for that here. This runs in a single thread and that looks fine. I wonder if that behavior in https://github.com/gnosis/prediction-market-agent/issues/485 also isn't an expected behavior? Like every process will create its own singleton, I'd say that's fine. But if it can be fixed, then why not.

Screenshot by Dropbox Capture

gabrielfior (approved):

Issue #422, opened 2024-09-19
Add a unique id to `Bet` class to make matching bets with langfuse traces easier


evangriffiths:

gabrielfior:
@evangriffiths correct me if I'm wrong, but I understood `transaction_hash` would be sufficient for uniquely identifying a bet?

evangriffiths:
Yeah sure. There's already an ID associated with a bet from the subgraph (see https://github.com/gnosis/prediction-market-agent-tooling/blob/5fe7dfd40ecb9cf088c4a9c42d5f78fbfc3a3605/prediction_market_agent_tooling/markets/omen/data_models.py#L385). Hopefully this is the same as the transaction hash so that we can:

1. save the transaction hash to the `Trade` object that is 'observed' by langfuse
2. get the id from the `ResolvedBet`s from `OmenAgentMarket.get_resolved_bets_made_since`

and match these up. I shall check!

PR #434, opened 2024-09-25
Trades return ID, matches with ID returned from AgentMarket.get_bets_made_since

evangriffiths:
Fixes https://github.com/gnosis/prediction-market-agent-tooling/issues/422


prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -703,14 +703,16 @@ def omen_buy_outcome_tx(
...
-    market_contract.buy(
+    tx_receipt = market_contract.buy(
         api_keys=api_keys,
         amount_wei=amount_wei_to_buy,
         outcome_index=outcome_index,
         min_outcome_tokens_to_buy=expected_shares,
         web3=web3,
     )
 
+    return tx_receipt["transactionHash"].hex()
```
evangriffiths, line 714: If I try to access the property like `tx_receipt.transactionHash` mypy complains, so accessing it as a dict key

evangriffiths, line 714: Could do
```
    trader_address = api_keys.bet_from_address
    nonce = ContractOnGnosisChain.get_transaction_count(trader_address)
    return str(tx_receipt.to) + str(trader_address.lower()) + str(nonce)
```
instead (to match `OmenBet.id` instead of `OmenBet.transactionHash`, but chose this for simplicity

prediction_market_agent_tooling/markets/data_models.py
```diff
@@ -126,3 +127,16 @@ class Trade(BaseModel):
     trade_type: TradeType
     outcome: bool
     amount: TokenAmount
+
+
+class PlacedTrade(Trade):
+    id: str
```
evangriffiths, line 133: Q: should we have an 'interim period with `id: str | None = None` to allow for re-use of old langfuse traces?

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -110,7 +111,7 @@ class OutOfFundsError(ValueError):
 
 class ProcessedMarket(BaseModel):
     answer: ProbabilisticAnswer
-    trades: list[Trade]
+    trades: list[PlacedTrade]
```
gabrielfior, line 114: Can we rename this to `LangfuseTrade` or similar, to make it clear it should be used just for tracking purposes?

prediction_market_agent_tooling/markets/omen/data_models.py
```diff
@@ -382,7 +382,7 @@ class OmenBetCreator(BaseModel):
 
 
 class OmenBet(BaseModel):
-    id: HexAddress
+    id: HexAddress  # A concatenation of: FPMM contract ID, trader ID, and trade transaction nonce. See https://github.com/protofire/omen-subgraph/blob/f92bbfb6fa31ed9cd5985c416a26a2f640837d8b/src/FixedProductMarketMakerMapping.ts#L109
```
gabrielfior, line 385: Tiny correction - nonce is a property of the account completing the transaction, not the transaction itself (basically a counter).
https://blog.thirdweb.com/nonce-ethereum/

tests_integration/markets/omen/test_get_bets.py
```diff
@@ -0,0 +1,34 @@
...
+import pytest
+
+from prediction_market_agent_tooling.config import APIKeys
+from prediction_market_agent_tooling.markets.agent_market import FilterBy, SortBy
+from prediction_market_agent_tooling.markets.omen.omen import OmenAgentMarket
+from prediction_market_agent_tooling.tools.utils import utcnow
+from tests.utils import RUN_PAID_TESTS
+
+
+@pytest.mark.skipif(not RUN_PAID_TESTS, reason="This test costs money to run.")
```
gabrielfior, line 12: Fine with this approach since you tested it manually, however please create a ticket to convert this to a `local_web3` test (should not be too hard, we just need to listen for FPMMBuy and FPMMSell and act accordingly). 
https://github.com/protofire/omen-subgraph/pull/65/files#diff-2395bda6bbff3cddb38de906d28e99e8752531a0019f6fed6541ac4af70f91beR288

prediction_market_agent_tooling/markets/data_models.py
gabrielfior: Can we also rename id to `transaction_hash` or at least add a comment for explicitness?

tests_integration/markets/omen/test_get_bets.py
evangriffiths: Hmm, but the point of this test is to check that the subgraph-returned `OmenBet` matches with the id from the transaction. If we convert it to a local web3 test, we'd also not use the subgraph, right?

prediction_market_agent_tooling/markets/data_models.py
evangriffiths: I've added the comment in omen/data_models.py. It don't like the idea of naming this transaction_hash, because it applied to all market types

prediction_market_agent_tooling/deploy/agent.py
evangriffiths: It's not just for tracking purposes - it has generic use (e.g. you could write a single script, like the test I added, that places trades, and gets them from the subgraph and does some processing on them)

tests_integration/markets/omen/test_get_bets.py
gabrielfior: If the point is testing the id-logic of the subgraph `(concat(txHash, etc...)`, you don't need to execute a transaction, rather you can simply fetch one bet and, from its transaction, fetch the relevant info and reconstruct the ID. So, in this case, I would argue for NOT having the @pytest.skip_if_paid flag.

I believe the above accomplishes the same as what your current test is trying to test, what do you think?

kongzii (approved):

tests_integration/markets/omen/test_get_bets.py
evangriffiths: > If the point is testing the id-logic of the subgraph (concat(txHash, etc...)
No, we're not using that concated id anywhere. The point is testing the subgraph in combination with the return values of buy/sell functions.

The test you mention doesn't have the same coverage - you could change something in `OmenAgentMarket.get_bets_made_since` that would break this test, but not that one.

I've added some asserts to tests_integration_with_local_chain/markets/omen/test_omen.py::test_omen_buy_and_sell_outcome so there's some CI coverage now, but I don't think we can get the new test to run without doing it on the real chain.

prediction_market_agent_tooling/deploy/agent.py
gabrielfior: Fine by me

tests_integration/markets/omen/test_get_bets.py
gabrielfior: > you could change something in OmenAgentMarket.get_bets_made_since that would break this test, but not that one.

I imagined this would call the `ID-generation logic`, thus breaking the test if something changed in the graph or here. But if the logic is not called anymore, then sure, it's not tracked in the test.

> I've added some asserts to

Those solve my concerns. Thanks.

gabrielfior (approved):

PR #436, opened 2024-09-26
Integrate Tenderly to local_web3

kongzii:


evangriffiths (commented):
Noob question - why do we need to support tenderly here in addition to foundry? As I understand it, tenderly is for running solidity tests on a mock chain. What's the use of that in this repo?

kongzii:
> Noob question - why do we need to support tenderly here in addition to foundry? As I understand it, tenderly is for running solidity tests on a mock chain. What's the use of that in this repo?

Use case is if you are creating a test against local chain (foundry), but it's failing without obvious reasons. You can switch to Tenderly by providing env variable and then, you can debug your test on Tenderly.

evangriffiths (approved):

gabrielfior (approved):

PR #437, opened 2024-09-26
Add a new contract used for debugging

kongzii:


prediction_market_agent_tooling/tools/contract.py
```diff
@@ -423,6 +429,48 @@ def get_asset_token_contract(
...
+        )
+        return now
+
+    def get_now(
+        self,
+        web3: Web3 | None = None,
+    ) -> DatetimeWithTimezone:
+        return add_utc_timezone_validator(
+            datetime.fromtimestamp(self.getNow(web3), tz=pytz.UTC)
+        )
```
gabrielfior, line 460: By the way, `getNow` can be misleading - what you are actually fetching from the contract is `block.timestamp` - see https://docs.soliditylang.org/en/latest/units-and-global-variables.html#block-and-transaction-properties

kongzii: I saw this somewhere as well, but Reality's contract is old and using `now`, so I wanted to debug with exactly the same thing.

tests_integration_with_local_chain/markets/omen/test_local_chain.py
```diff
@@ -93,3 +98,37 @@ def test_fresh_account_has_less_than_minimum_required_balance(
...
+def test_now_datetime(local_web3: Web3, test_keys: APIKeys) -> None:
+    # we need to mint a new block to update timestamp
+    DebuggingContract().inc(test_keys, local_web3)
+    allowed_difference = 10  # seconds
+    chain_datetime = DebuggingContract().get_now(local_web3)
+    utc_datetime = utcnow()
+    actual_difference = abs(chain_datetime - utc_datetime)
+    assert actual_difference <= timedelta(
+        seconds=allowed_difference
+    ), f"chain_datetime and utc_datetime differ by more than {allowed_difference} seconds: {chain_datetime=} {utc_datetime=} {actual_difference=}"
```
gabrielfior, line 134: Those tests to me don't make too much sense 🤔 

By default, Anvil creates a new block every time a new transaction is completed.
https://book.getfoundry.sh/reference/anvil/#mining-modes

Additionally, `DebuggingContract` simply outputs `block.timestamp`, which can be more easily fetched (without a smart contract) via `local_web3.eth.block_number`.
So I'm not sure what exactly is the goal of this test - could you please clarify?

kongzii: When debugging something, it's good to be as close to the real scenario as possible. Reality's contract isn't using `local_web3.eth.block_number`, it's using `now` in its functions.

Compare `test_now` and `test_now_failed`.

These tests verify my assumption that if you want `now` in the contract to be up-to-date, you have to mint a new block (using the `DebuggingContract().inc(test_keys, local_web3)`.

gabrielfior: > These tests verify my assumption that if you want now in the contract to be up-to-date, you have to mint a new block

We are talking about different things - `now` reflects the block timestamp when the transaction was executed. On Anvil, since each block has one transaction, the block is quickly confirmed hence timestamps don't differ much. On Gnosis mainnet, block confirmation takes a little longer (approx 6s), hence now would return an outdated value.

> you have to mint a new block (using the DebuggingContract().inc(test_keys, local_web3).

That's not entirely true. You can also achieve the exact same thing using the web3 provider.
```
previous_block_timestamp = local_web3.eth.get_block('latest').timestamp
local_web3.manager.request_blocking(RPC.evm_mine, [previous_block_timestamp + 1]) # next block timestamp >= prev block timestamp
```

kongzii: This is what I was talking about on standup. Imagine this:

1. Anvil creates local_chain, it's timestamp (`now`) is updated
2. You create a new question, there's a new transaction because of it, so timestamp (`now`) is updated again
3. You do `time.sleep(X)` where `X` is the time of seconds when the question should open
4. You want to answer this question, the first thing that contracts does is checking if `opening_timestamp >= now`, however, because there was no transaction between (3) and (4), `now` in your local chain is still old, and so the answering fails with revert error.

So between (3) and (4), you need to do a dummy transaction that will cause the local chain to be updated. Then (4) works because `now` in the contract is updated.

> That's not entirely true. You can also achieve the exact same thing using the web3 provider.

That's interesting. What is `RPC` in your code? But executing a dummy transaction `DebuggingContract().inc(test_keys, local_web3)` still feels more real-world-like.

gabrielfior: I agree with your logic - the trade-off here is, to implement your test, you can either deploy a new contract OR use standard web3py methods. The less complex option is probably the latter.

> What is RPC in your code?
from web3._utils.rpc_abi import RPC

kongzii: It's already deployed, so not a big deal. Do you see any potential issues arising from the current implementation?

gabrielfior: Yes, please see comment -> https://github.com/gnosis/labs-contracts/pull/9#issuecomment-2377812890
As far as I understood the situation (please correct me if I'm wrong), the contract is not needed in favor of web3py methods. Surely the contract has already been deployed, but if it's not needed then we should remove it.
Removing it implies on re-writing 2 tests on this PR with the method I mentioned earlier `RPC.evm_mine`, so it's also not a big deal.

kongzii: > Yes, please see comment -> https://github.com/gnosis/labs-contracts/pull/9#issuecomment-2377812890

You don't mention any potential issues in that comment. What problems do you reckon?

I'm arguing for using the current way because it resembles the real production scenario more.

I'm going to move `DebuggingContract().inc(test_keys, local_web3)` to a function to make it more modular, but otherwise, what problem do you see here?

gabrielfior: From my comment
```
Everything that is achieved calling the getNow function can be achieved calling web3.eth.get_block('latest').timestamp
```

From the Realitio contract (I assume this is what you mean by real production scenaio?)(https://gnosisscan.io/address/0x79e32aE03fb27B07C89c0c568F80287C01ca2E57#code#L222)
```
require(finalize_ts == UNANSWERED || finalize_ts > uint32(now), "finalization deadline must not have passed");
```

I don't see any potential issues with the contract implementation. What I'm saying is, you can either
a. use the current PR and new contract to mine a new block on a local chain and also test a native EVM function (block.timestamp) OR
b. Use standard functions from web3py and RPC method `auto_mine` for accomplishing the exact same thing, without an extra contract. You can also test Realitio contract using their ABI (it's verified).

kongzii: > What I'm saying is, you can either

Yeah, and I said a few comments above that I preferred the `a.`, so the question from my side was if you see any issues with that implementation.

Because you don't, is that still, for some reason, a real blocker for you, or can you approve so I can go ahead and merge all the chained PRs?

If it bugs you too much I'll change it 😄 If not it makes more sense like this to me.

gabrielfior (approved):

PR #441, opened 2024-09-26
Create OmenMarket out of blockchain events

kongzii:


tests_integration/markets/omen/test_markets.py
```diff
@@ -0,0 +1,86 @@
...
+    # Save because otherwise it's hard to debug
+    with open("tests_files/created_market.json", "w") as f:
+        json.dump(created_market.model_dump(), f, indent=2)
+    with open("tests_files/from_created_market.json", "w") as f:
+        json.dump(from_created_market.model_dump(), f, indent=2)
+    with open("tests_files/from_subgraph_market.json", "w") as f:
+        json.dump(from_subgraph_market.model_dump(), f, indent=2)
+
+    # Compare!
+    assert from_created_market == from_subgraph_market
```
kongzii, line 86: It's 100% 1:1

Screenshot by Dropbox Capture

prediction_market_agent_tooling/markets/omen/data_models.py
```diff
@@ -376,6 +376,75 @@ def get_resolution_enum(self) -> t.Optional[Resolution]:
...
+        if len(model.market_event.conditionIds) != 1:
+            raise ValueError(
+                f"Unexpected number of conditions: {len(model.market_event.conditionIds)}"
+            )
+        outcome_token_amounts = model.funding_event.outcome_token_amounts
+        total_amount = sum(outcome_token_amounts)
+        outcome_token_marginal_prices = [
+            xDai((total_amount - amount) / total_amount)
+            for amount in outcome_token_amounts
+        ]
```
gabrielfior, line 394: I'm unsure if the order here matters or not.
If a market has outcomes [Yes,No] or [No,Yes], would this still work?

kongzii: This is just an ordered list of outcome tokens as retrieved from the event. It doesn't care about what outcomes are at these indices. If you switch Yes,No order, it needs to be handled at different places.

gabrielfior: What I guess I'm suggesting here is
-> Fetch the 2 outcomes (Yes, No) from the question.
-> Find the indices of Yes, No in the outcomes.
-> Having the indices, you can create a correctly-ordered list of `outcomeTokenMarginalPrices` with Yes, No in a known order.

It's also possible to find out downstream the order of outcomes, I just find that doing it here is a good spot.

kongzii: >  I just find that doing it here is a good spot

I disagree here. This here is doing the same thing that subgraph is doing, it's meant to be 1:1 OmenMarket that you'd get from subgraph otherwise (which it is). Subgraph also won't re-order these outcome tokens according to your specific needs. 

That's why we have methods like [get_outcome_index](https://github.com/gnosis/prediction-market-agent-tooling/blob/53a85c9bbd7e0facd94dfe6aa5f42502e141ba74/prediction_market_agent_tooling/markets/agent_market.py#L234), so the right way to go about it, if you need outcome tokens for a specific outcome, is something like

```
x.outcome_token_marginal_prices[x.get_outcome_index("Yes")]
```

and then you don't need to care about the order of [Yes, No].

kongzii: Which we are actually alredy doing for example here https://github.com/gnosis/prediction-market-agent-tooling/blob/53a85c9bbd7e0facd94dfe6aa5f42502e141ba74/prediction_market_agent_tooling/markets/omen/data_models.py#L345

or here https://github.com/gnosis/prediction-market-agent-tooling/blob/45601b68c827206a85dcf194732c1f3990fd7509/tests/markets/test_betting_strategies.py#L180

evangriffiths (approved):

evangriffiths:
Looks like a couple of CI tests need fixing

Issue #444, opened 2024-09-29
Kleros Arbitrator is outdated and creates disputes with no possibility for appeal


greenlucid:
The current Kleros Contract, https://github.com/gnosis/prediction-market-agent-tooling/blob/247e7b25b255ca2a614286b4fe1b1ce8613312af/prediction_market_agent_tooling/markets/omen/omen_contracts.py#L489

`0xe40DD83a262da3f56976038F1554Fe541Fa75ecd`, which was probably found here: https://docs.kleros.io/developer/deployment-addresses

is in Gnosis Chain, links to this proxy in Mainnet:

[0x79d0464Ec27F67663DADf761432fC8DD0AeA3D49](https://etherscan.io/address/0x79d0464Ec27F67663DADf761432fC8DD0AeA3D49)

it has NO appeals. This is a problem if a market escalates.

Historically, this was the one used: [0xd47f72a2d1d0e91b0ec5e5f5d02b2dc26d00a14d](https://etherscan.io/address/0xd47f72a2d1d0e91b0ec5e5f5d02b2dc26d00a14d) but I don't know why it's not the one in there. I assume due to some MetaEvidence bugs that forced it to be updated.

---

Anyhow, no appeals is a problem. If you check [0x79d0464Ec27F67663DADf761432fC8DD0AeA3D49](https://etherscan.io/address/0x79d0464Ec27F67663DADf761432fC8DD0AeA3D49) and read the `arbitratorExtraData`, after some digging you'll find how it will intend to create a dispute with 511 jurors, and provides no interface whatsoever to appeal. Even if it did, it cannot be appealed anyway since that's the max amount of jurors that can be summoned.

For context, which is not shown anywhere and might be important for the prompting in some of the agents, this is the policy that would currently be put in place if a question ever escalated: https://cdn.kleros.link/ipfs/QmZM12kkguXFk2C94ykrKpambt4iUVKsVsxGxDEdLS68ws/omen-rules.pdf

Now, if you want disputes with appeals (and I'm pretty sure you do, consult @clesaege on this), you need to setup a different proxy, it's also found on Kleros Deployments.

This will be the Mainnet Proxy: [0x2F0895732bfacdCF2fdB19962fE609D0dA695F21](https://etherscan.io/address/0x2F0895732bfacdCF2fdB19962fE609D0dA695F21)

[...]

PR #445, opened 2024-09-29
feat: create kleros disputes that can be appealed

greenlucid:
Just read #444 (nice number) for an explanation of what's going on


kongzii (approved):
Thanks again, I tested this today and looks good. However now it sounds like this should be configurable instead of forced, I will merge this no, but feel free to take a look at https://github.com/gnosis/prediction-market-agent-tooling/pull/451 if you want

Issue #363, opened 2024-08-27
Universal place to store information about the bet


kongzii:
We need to store somewhere:

- predicted probability 
- reasoning about the prediction
- ? maybe something else

Omen contracts are public, and Presagio is as well, so the solution should be usable by everyone, such that anyone's agent can contribute it.

@evangriffiths says let's host a public Postgres

@kongzii Complains that Postgres is too web2ish! Let's create a new contract for it.

@gabrielfior Says some fancy web3storage solution exists.

Goals:

- Presagio can show it as bet's details
- We can use it on Dune to track accuracy and other fancy metrics based on probability

kongzii:
>? maybe something else

There could be also `is_agent: bool` flag that Presagio could use to flag the bets as AI on the UI. Right now, they are querying it from Dune.

gabrielfior:
I'm thinking about a way to use IPFS to solve this - maybe leveraging this project (https://storacha.network/).
We would have a central file containing a dictionary (market_id -> prediction_str_IPFS_hash (and other metadata)), so we could fetch the information on a series of IPFS addresses.
The unsolved question is: how do we add more market_ids to the initial file and keep it addressable, since changing the file contents will change its CID?

kongzii:
For IPFS we have already Pinata from devops.

But I’m afraid that we would have the same unsolvable problems as with image mapping (which ended up being a contract)

kongzii:
Note: Here it would make sense to me if the solution is upgradable in a backward-compatible way. If we want to add some new fields or something in the future.

berteotti:
On Presagio side it would be good to query this info by agent_id and market_id. So somewhere on the data or some kind of index with this info would be appreciated. On our side we see this data has relevant:

- Reasoning about the prediction (this could look like a comment about the decision)
- Predicted probability
- Model used (already present in mechs data)
- Data sources for the reasoning (news, articles, etc.), if possible

> There could be also `is_agent: bool` flag that Presagio could use to flag the bets as AI on the UI. Right now, they are querying it from Dune.

This means that the "database" will have agent and non-agent information?

evangriffiths:
> Note: Here it would make sense to me if the solution is upgradable in a backward-compatible way. If we want to add some new fields or something in the future.

[...]

PR #446, opened 2024-09-29
404 add methods to pmat for storing prediction into ipfs and adding that to contract mapping

gabrielfior:


prediction_market_agent_tooling/tools/ipfs/ipfs_handler.py
```diff
@@ -0,0 +1,22 @@
...
+from prediction_market_agent_tooling.config import APIKeys
+from prediction_market_agent_tooling.gtypes import IPFSCIDVersion0
+
+
+class IPFSHandler:
+    def __init__(self, api_keys: APIKeys):
+        self.pinata = PinataPy(
+            api_keys.pinata_api_key.get_secret_value(),
+            api_keys.pinata_api_secret.get_secret_value(),
+        )
```
gabrielfior, line 12: See https://github.com/gnosis/prediction-market-agent/issues/489

tests/markets/omen/test_omen_subgraph_handler.py
```diff
@@ -285,44 +285,47 @@ def build_incomplete_user_position_from_condition_ids(
     )
 
 
-def test_get_markets_id_in() -> None:
+def test_get_markets_id_in(omen_subgraph_handler: OmenSubgraphHandler) -> None:
     market_id = "0x934b9f379dd9d8850e468df707d58711da2966cd"
-    sgh = OmenSubgraphHandler()
-    markets = sgh.get_omen_binary_markets(
+    markets = omen_subgraph_handler.get_omen_binary_markets(
```
gabrielfior, line 290: Using pytest.fixture here instead of instantiating a new instance every time. Small tweak hence doing it here also.

gabrielfior:
Closes https://github.com/gnosis/prediction-market-agent-tooling/issues/363

tests/markets/omen/test_omen_subgraph_handler.py
gabrielfior: Regardless if it's a singleton or not, we want to instantiate (or get a reference to the created instance) in just one place, and that place is the `conftest`.

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -476,18 +488,57 @@ def process_market(
...
+    def store_prediction(
+        self, market_id: str, processed_market: ProcessedMarket, keys: APIKeys
+    ) -> None:
+        reasoning = processed_market.answer.reasoning if processed_market.answer.reasoning else ""
+        ipfs_hash = self.ipfs_handler.store_agent_result(
+            IPFSAgentResult(reasoning=reasoning)
+        )
+
+        # We store only the first trade, since it's only here for tracking purposes - also,
+        # we want only 1 entry per prediction, not 1 entry per trade.
```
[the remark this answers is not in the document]
gabrielfior: I think this should be fixed now (https://github.com/gnosis/prediction-market-agent-tooling/pull/446/files#diff-6aaf85d952cb99bbd129bef760f5e94cd4477d698103727006d150de8ff9780fR543) that we store multiple `txHashes` per prediction.

tests/markets/omen/test_omen_subgraph_handler.py
kongzii: > we want to instantiate (or get a reference to the created instance) in just one place

Why? Wasn't your point back then about making it a singleton that it's initialized only once? And no matter how many times we do `OmenSubgraphHandler()` it's always returning the already initialised one? So, is not initialising it "again" just a micro-optimization at best? (because it just returns existing, already initialised, instance)

(Just to be clear: I don't mind refactoring it to a fixture that's passed as an argument as you are doing it here, just trying to understand the motivation as it feels like we are discussing "do not initialize OmenSubgraphHandler" multiple times recently and I thought it was solved nicely by you when it became a singleton. If nothing else, it's nice not to do `OmenSubgraphHandler()` all over the tests just for esthetic reasons 😄 )

prediction_market_agent_tooling/deploy/agent.py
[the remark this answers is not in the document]
gabrielfior: test

tests/markets/omen/test_omen_subgraph_handler.py
gabrielfior: The point here is not `is it a singleton?` (although for the record it is) - the point is, `fixtures` (https://docs.pytest.org/en/stable/explanation/fixtures.html#about-fixtures) is the pytest way to abstract away the `arrange` phase of the tests. Instantiating the `OmenSubgraphHandler` is necessary for every test and should be done in the same way (e.g. with the same args) on every test, hence it makes sense to abstract it away, and in pytest one does that with fixtures.

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -96,6 +99,13 @@ def __init__(self) -> None:
             )
         )
 
+        # ToDo - Update URL once curations are live
```
evangriffiths, line 102: Can this TODO be removed now?

tests/markets/test_betting_strategies.py
```diff
@@ -281,7 +281,7 @@ def test_zero_bets() -> None:
         target_p_yes=market.current_p_yes,
         fee=market.fee,
     )
-    assert np.isclose(market_moving_bet.size, 0.0, atol=1e-4)
+    assert np.isclose(market_moving_bet.size, 0.0, atol=1e-3)
```
evangriffiths, line 284: eek! flaky test?

tests_integration/tools/ipfs/test_ipfs_handler.py
```diff
@@ -0,0 +1,33 @@
...
+    # We add the current datetime to avoid uploading an existing file (CID is content-based)
+    temp_string = f"Hello World {datetime.datetime.utcnow()}"
+    with NamedTemporaryFile() as temp_file:
+        temp_file.write(temp_string.encode("utf-8"))
+        temp_file.flush()
+        ipfs_hash = test_ipfs_handler.upload_file(temp_file.name)
+
+    # assert uploaded
+    # can take a while to be available for download
+    # r = requests.get(f"https://ipfs.io/ipfs/{ipfs_hash}", timeout=120)
```
evangriffiths, line 28:
```suggestion
```

tests_integration_with_local_chain/conftest.py
```diff
@@ -73,3 +76,16 @@ def fund_account_on_tenderly(
     }
     response = requests.post(fork_rpc, json=payload)
     response.raise_for_status()
+
+
+def create_and_fund_random_account(
```
evangriffiths, line 81: This looks unrelated to the rest of the PR. I see you've added a test for it, but it's not used anywhere in the code. No biggie but would make sense in a separate PR

evangriffiths (approved):

tests_integration_with_local_chain/conftest.py
gabrielfior: That's correct, but this test was used to test a part of the function that fails sometimes during CI runs - so I added it here as well for improving test coverage.
Happy to refactor if this becomes too big to maintain or unreliable.

tests/markets/test_betting_strategies.py
gabrielfior: I had to increase the absolute tolerance - for some markets, it fails, for others, it works. Should be still a valuable test though.

PR #451, opened 2024-10-01
Configurable arbitrator

kongzii:


evangriffiths (approved):

greenlucid:
Yeah that looks good, would you like some more options? Like, deploying the dispute in Gnosis Chain directly (for low value use cases, e.g. these markets which are about ~7$ each)
For the 3 jurors in Mainnet (which would likely be the sanest option) I'm not sure how I would proceed. I'd ask @unknownunknown1 since he's the one who deployed these RealityProxies I believe. [0x2F0895732bfacdCF2fdB19962fE609D0dA695F21](https://etherscan.io/address/0x2F0895732bfacdCF2fdB19962fE609D0dA695F21#code) has no option to change the `arbitratorExtraData` so deploying a new one would be needed.

unknownunknown1:
>I believe. [0x2F0895732bfacdCF2fdB19962fE609D0dA695F21](https://etherscan.io/address/0x2F0895732bfacdCF2fdB19962fE609D0dA695F21#code) has no option to change the `arbitratorExtraData` so deploying a new one would be needed.

Yeah this version of the proxy doesn't have governance functionality, so if you need different extradata you'll have to to redeploy

greenlucid:
@unknownunknown1
Can you discuss internally if it would be a good idea to deploy a RealityProxy in mainnet to support that new `arbitratorExtraData`? I would do it but I don't want to spend my own ETH in gas to deploy it.
Also maybe General Court is not the best Mainnet court for answering Omen questions, unsure.
Also you might want to get Facundo to review in case there's an issue in the policy.
Anyway, letting you know in case you might want to mention this in a call or something, since there are many many markets created by these agents and it would be good to give good guarantees

unknownunknown1:
I'll mention it tomorrow but can you describe this use case fully. I'm not aware. Also do you have a link to policy?
General court probably not the best for such low number of jurors indeed. 
@greenlucid

greenlucid:
Context:

Gnosis made this repo, and https://github.com/gnosis/prediction-market-agent repo, that has AI agents creating prediction markets in Gnosis Chain, adding liquidity, and resolving them automatically. Whenever these markets are created with Omen, they setup Kleros as Arbitrator.

I made an issue here: https://github.com/gnosis/prediction-market-agent-tooling/issues/444 pinpointing how the arbitrator they setup for Omen all this time, was connected to the RealityProxy in mainnet that created disputes for 511 jurors in General Court. In this issue you'll see links to the both policies the two RealityProxies have: [the 511-juror one (previous)](https://cdn.kleros.link/ipfs/QmZM12kkguXFk2C94ykrKpambt4iUVKsVsxGxDEdLS68ws/omen-rules.pdf) and the [31-juror one (current default)](https://ipfs.kleros.io/ipfs/QmaUr6hnSVxYD899xdcn2GUVtXVjXoSXKZbce3zFtGWw4H/Question_Resolution_Policy.pdf). This new current default is more of a general purpose policy.

So, **on the policies**, the idea would be to make sure these policies are up to date, if they're good enough, etc. I know a [new Policy was made for Seer](https://cdn.kleros.link/ipfs/QmW6npv4J3jvF87WVCEaQQ6f2PK1T9ytX2eYg86XkT9ScK/Seer%20-%20Markets%20Policy.pdf), although Omen markets do not properly handle `ANSWERED_TOO_SOON` so they must use invalid instead.

And **on the proxies**, I wrote the following in Gnosis Discord, #gnosis-ai which I'll quote:

>  options are:
> - make another pair of RealityProxies, but with arbitratorExtraData to make a dispute with 3 jurors in an specialized court. that'd prob be 0.05ETH to start. you could even create a dispute with 1 juror (but, you'll likely receive advice for why 1 juror might be a bad idea)
> - instead of using Mainnet Kleros Court, you can use the Court in Gnosis Chain. So, for example, you could draw 3 jurors from Gnosis Curation Court, that'd be 19.8 xdai (0.01ETH), and cheaper in gas since the whole operation is contained in Gnosis Chain. https://klerosboard.com/100/courts/1 And yeah you could even go down to 1 juror, 6.6 xdai for arbitration.

I think any of those two options would be preferable to the status quo. Maybe the second option I propose, setting up a RealityProxy and arbitrate in Gnosis Chain directly would be way cheaper, especially given these are agents putting in between 5$ to 9$ in liquidity per market, so they are strictly low value use cases.

PR #456, opened 2024-10-02
Force DatetimeUTC everywhere

kongzii:
There are two patterns, both are valid:

(1)

```
class X(BaseModel):
   start_time: DatetimeUTC
```

or

(2)

```
class X(BaseModel):
  start_time_timestamp: int

  def start_time_datetime(self) -> DatetimeUTC:
    return to_utc_datetime(self.start_time_timestamp)
```

I chose (2) when the name of the field is like `x_ts,` `x_timestamp,` `xTimestamp,` etc. Because if the field is named timestamp, I guess it should be a timestamp.

And (1) if the name is like `start_time`, `creation`, etc. Because then there is no assumption about the type and it can be datetime right away.


prediction_market_agent_tooling/markets/manifold/data_models.py
```diff
@@ -100,15 +99,6 @@ def is_resolved_non_cancelled(self) -> bool:
     def __repr__(self) -> str:
         return f"Manifold's market: {self.question}"
 
-    @field_validator("closeTime", mode="before")
```
evangriffiths, line 103: Do we still not want to keep this? I remember there were one or two rogue markets that had closeTimes that weren't valid datetimes, and that caused `get_binary_markets` to throw an exception

kongzii: DatetimeUTC is now taking care of it by itself: https://github.com/gnosis/prediction-market-agent-tooling/pull/456/files#diff-f527635d08816722a0a224ff68f723862951187fcb07d64937a2b14e6a97551dR69

evangriffiths: Oh I see you've moved the logic to `DatetimeUTC.to_datetime_utc`. But that wouldn't run here, so you're still vulnerable to the exception, right?

evangriffiths: Oh scrap that, it should run in `DatetimeUTC._validate`. Nice!

kongzii: It will! 👍 `DatetimeUTC` implements `_validate` which is used by Pydantic

Screenshot by Dropbox Capture

So any Pydantic model that is typed as `xyz: DatetimeUTC` will use this `to_datetime_utc` method to validate input

prediction_market_agent_tooling/tools/datetime_utc.py
```diff
@@ -0,0 +1,74 @@
...
+        return core_schema.no_info_after_validator_function(cls._validate, dt_schema)
+
+    @staticmethod
+    def from_datetime(dt: datetime) -> "DatetimeUTC":
+        """
+        Converts a datetime object to DatetimeUTC, ensuring it is timezone-aware in UTC.
+        """
+        if dt.tzinfo is None:
+            logger.warning(
+                f"tzinfo not provided, assuming the timezone of {dt=} is UTC."
```
evangriffiths, line 47: Do we get spammed with warnings now?

evangriffiths (approved):
Nice one. This looked like a big undertaking 😅

prediction_market_agent_tooling/tools/datetime_utc.py
kongzii: I hope not 😄 I will check it out in the logs

prediction_market_agent_tooling/markets/omen/omen_contracts.py
```diff
@@ -574,7 +575,7 @@ def askQuestion(
         outcomes: list[str],
         language: str,
         arbitrator: Arbitrator,
-        opening: datetime,
+        opening: DatetimeUTC,
```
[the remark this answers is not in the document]
kongzii: Not true! Added a test to test it out.

PR #458, opened 2024-10-02
Fix metaculus limit when getting markets

kongzii:


prediction_market_agent_tooling/markets/metaculus/metaculus.py
```diff
@@ -100,7 +100,7 @@ def get_binary_markets(
 
             if len(all_questions) >= limit:
                 break
-        return [MetaculusAgentMarket.from_data_model(q) for q in all_questions]
+        return [MetaculusAgentMarket.from_data_model(q) for q in all_questions[:limit]]
```
kongzii, line 103: I'm writing test for the datetime stuff and noticed metaculus returned 100 markets when I asked for 10

PR #466, opened 2024-10-04
Default 3 days reality timeout

kongzii:


prediction_market_agent_tooling/markets/omen/omen_contracts.py
```diff
@@ -431,6 +431,7 @@ class sDaiContract(ContractERC4626OnGnosisChain):
 
 
 OMEN_DEFAULT_MARKET_FEE_PERC = 0.02  # 2% fee from the buying shares amount.
+REALITY_DEFAULT_FINALIZATION_TIMEOUT = timedelta(days=3)
```
kongzii, line 434: Clement recommended 3.5 days, but in the Olas market-creator, the code is typed as integer and I didn't want to change their code too much. So using just 3 days here as well to keep it consistent.

evangriffiths (approved):

PR #470, opened 2024-10-04
Added boolean flag to control IPFS upload

gabrielfior:


prediction_market_agent_tooling/config.py
```diff
@@ -151,6 +152,12 @@ def default_enable_langfuse(self) -> bool:
             and self.LANGFUSE_HOST is not None
         )
 
+    @property
+    def enable_ipfs_upload(self) -> bool:
+        return check_not_none(
+            self.ENABLE_IPFS_UPLOAD, "ENABLE_IPFS_UPLOAD missing in the environment."
```
kongzii, line 158: I'm thinking if we should force user to enable or disable it like this (because if they don't provide the key, it will error out). Or if we should default to False? 

No strong opinion about either choice on my side.

HandyPicklesStudioz (commented):
YOLO

prediction_market_agent_tooling/config.py
gabrielfior: Default False, please see latest commit

kongzii: In that case it can be just `ENABLE_IPFS_UPLOAD: bool = False`, no need for optionality anymore

PR #474, opened 2024-10-07
Fix market creation script

kongzii:


scripts/create_market_omen.py
```diff
@@ -18,7 +20,7 @@
 
 def main(
     question: str = typer.Option(),
-    closing_time: DatetimeUTC = typer.Option(),
+    closing_time: datetime = typer.Option(),
```
kongzii, line 23: Typer errors out on custom type

evangriffiths (approved):

PR #478, opened 2024-10-08
Remove misleading filter

kongzii:


prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -273,12 +272,6 @@ def _build_where_statements(
             else:
                 where_stms["resolutionTimestamp"] = None
 
-        if finalized is not None:
```
kongzii, line 276: This filter was misleading because `answerFinalizedTimestamp` can be in the future, see an example at https://github.com/SwaprHQ/presagio/pull/110#issuecomment-2393335169. So before this PR, if we filtered for `finalized`, it didn't mean that the question is actually finalized, it can be finalized any time in the future. It means only that some answered is posted there.

It's safer and more straightforward to use the filter `finalized_before`, and this one can be removed.

kongzii: Good question is what about the `resolved` flag we have here as well. That one should works as expected, because only finalized question can be resolved and so the resolution time can not be in the future.

evangriffiths (approved):

kongzii:
Failed tests are unrelated to this PR, will check them out in a separate one 

Screenshot by Dropbox Capture

PR #488, opened 2024-10-10
Auto-verify that buy trades wouldn't cause guaranted loss

kongzii:


prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -112,9 +112,11 @@ def calculate_trades(
     ) -> list[Trade]:
         direction = self.calculate_direction(market.current_p_yes, answer.p_yes)
 
+        # Don't bet `bet_amount` if it's less than liquidity, because that means just losing money instantly.
```
evangriffiths, line 115: Can you spell out the argument for this? I think this is a counter-example:
```
- Initial market: 2 yes, 2 no. Constant: 2 * 2 = 4
- I bet on Yes with 5 xDai.
- Market now has 7 yes, 7 no.
- 7 * 0.5714 = 4, 7 - 0.5714 = 6.4286, so...
- I get back 6.4286 tokens to restore the constant.
- If I win, I make a profit of 1.4286 xDai (even though my bet was larger than the market liquidity).
```

kongzii: Aaah, you are right, thanks!

I over-simplified it too much.

The motivation here is that for example if agent is going to bet on this market (assuming it would be valid question) with this strategy and fixed bet size, it will lose money right away because of low liquidity.

But it needs to be calculated better than just this min(x,y) check, I'll take a look tomorrow.

Screenshot by Dropbox Capture

gabrielfior: Expanding on @evangriffiths 's example for a market with k=4 (y=2, n=2) and bet_amount=100
```
k = 4
received_y_outcome_tokens = 101.96
```
So it's still profitable if you place a large bet, but `(profit/bet_amount) -> 0` as bet_amount increases.

kongzii: > So it's still profitable if you place a large bet, but (profit/bet_amount) -> 0 as bet_amount increases.

Forgetting fees here, the default fee is 2%, so you could win $1.96, but you will for sure lose $2, and then you are immediately in loss no matter the outcome.

I over-simplified the solution in this PR, but it's a real problem that needs a fix.

evangriffiths: Sure. But fwiw the bet size you need to make for this to happen on newly-created Olas markets (with 7 xDai liquidity) is ~341 xDai, so not too much of a concern for the common case!

Screenshot 2024-10-11 at 10 57 25

evangriffiths: ooo, interesting. you'd make a great red hat hacker!

prediction_market_agent_tooling/markets/agent_market.py
```diff
@@ -166,6 +166,11 @@ def place_bet(self, outcome: bool, amount: BetAmount) -> str:
     def buy_tokens(self, outcome: bool, amount: TokenAmount) -> str:
         return self.place_bet(outcome=outcome, amount=amount)
 
+    def get_buy_token_amount(
+        self, bet_amount: BetAmount, direction: bool
+    ) -> TokenAmount:
+        raise NotImplementedError("Subclasses must implement this method")
```
evangriffiths, line 172: Was thinking if this would be possible for other market types to implement (not for this PR, just thinking...)

and for manifold the v0/bet [api ](https://docs.manifold.markets/api#post-v0bet) has a `dryRun` param that returns a simulated result - so that could work 👍

evangriffiths (approved):

prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -48,6 +48,26 @@ def assert_trades_currency_match_markets(
...
+            if trade.trade_type == TradeType.BUY:
+                outcome_tokens_to_get = market.get_buy_token_amount(
+                    trade.amount, trade.outcome
+                )
+
+                if outcome_tokens_to_get.amount < trade.amount.amount:
+                    raise RuntimeError(
+                        f"Trade {trade=} would result in guaranteed loss by getting only {outcome_tokens_to_get=}."
+                    )
+
```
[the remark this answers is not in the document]
kongzii: > However, due to the mechanics of automated market makers, it's possible for the amount of outcome tokens received to be less than the trade amount while still expecting a profit upon a favorable outcome.

@coderabbitai How's that possible? Isn't 1 outcome token equal to 1 investment value? So if I receive less of them, I'm in loss?

prediction_market_agent_tooling/markets/agent_market.py
kongzii: Cool! I was actually trying to use `get_buy_token_amount` from Omen logic but it doesn't match.

@coderabbitai Please create an issue for Evan's comment above, in case we decide to bet on Manifold again.

PR #493, opened 2024-10-14
Refactor omen-specific logic out of DeployableTraderAgent + fix minimum required balance to operate

kongzii:


prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -294,7 +294,6 @@ def get_gcloud_fname(self, market_type: MarketType) -> str:
 
 class DeployableTraderAgent(DeployableAgent):
     bet_on_n_markets_per_run: int = 1
-    min_required_balance_to_operate: xDai | None = xdai_type(1)
```
kongzii, line 297: This was hard-coded to 1, but now agents can place high bets thanks to different strategies.

Instead of hard-coding, this PR get the strategy and requires that minimal required balance is maximal possible bet from the strategy.

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -352,37 +351,35 @@ def update_langfuse_trace_by_processed_market(
             ]
         )
 
-    def check_min_required_balance_to_operate(
-        self,
-        market_type: MarketType,
-        check_for_gas: bool = True,
-        check_for_trades: bool = True,
-    ) -> None:
+    def check_min_required_balance_to_operate(self, market_type: MarketType) -> None:
```
kongzii, line 354: I split the original function into 2, because the second one (because of the betting strategy) needs to accept AgentMarket as well, but we don't have that in `before_process_markets`

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -352,37 +351,35 @@ def update_langfuse_trace_by_processed_market(
...
+        self, market_type: MarketType, market: AgentMarket
+    ) -> None:
+        api_keys = APIKeys()
+
+        # Get the strategy to know how much it will bet.
+        strategy = self.get_betting_strategy(market)
+        # Have a little bandwich after the bet.
+        min_required_balance_to_trade = strategy.maximum_possible_bet_amount * 1.01
+
+        if market_type == MarketType.OMEN and not is_minimum_required_balance(
```
evangriffiths, line 377: I see that `is_minimum_required_balance` is an omen-specific function, hence the market_type check. Could you either:
- add a comment/TODO to say how this is partial coverage, and other market types will just fail depending on how the polymarket/manifold/... APIs handle it.
- or maybe a nicer thing to do would be add `pass` function on AgentMarket, and then override it for OmenAgentMarket
?

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -352,37 +351,35 @@ def update_langfuse_trace_by_processed_market(
...
-            )
+
+    def check_min_required_balance_to_trade(
+        self, market_type: MarketType, market: AgentMarket
+    ) -> None:
+        api_keys = APIKeys()
+
+        # Get the strategy to know how much it will bet.
+        strategy = self.get_betting_strategy(market)
+        # Have a little bandwich after the bet.
```
evangriffiths, line 374: sandwich 🥪 , or bandwidth?

prediction_market_agent_tooling/deploy/agent.py
kongzii: Good point, DeployableTraderAgent was becoming a "bit" Omen-specific.

Wdyt now? (but it bloated the PR, sorry 😄 )

prediction_market_agent_tooling/markets/agent_market.py
```diff
@@ -198,6 +205,36 @@ def get_binary_markets(
...
+    def get_trade_balance(api_keys: APIKeys) -> float:
+        """
+        Return balance that can be used to trade on the given market.
+        """
+        raise NotImplementedError("Subclasses must implement this method")
+
+    @staticmethod
+    def verify_operational_balance(api_keys: APIKeys) -> bool:
+        """
+        Return if the user has enough of operational balance, if not needed, just implement with `pass`.
```
evangriffiths, line 225: Function return type is a boolean
```suggestion
        Return `True` if the user has enough of operational balance. If not needed, just return `True`.
```

prediction_market_agent_tooling/deploy/agent.py
evangriffiths: Yes, but it's great!

evangriffiths (approved):

prediction_market_agent_tooling/markets/agent_market.py
kongzii: Oops True! 😄

PR #507, opened 2024-10-17
Pass fees to Kelly

kongzii:
I executed the betting strategy benchmark with no fees vs. with fees, comparison is here: https://www.diffchecker.com/bOvQec3S/

TLDR: With fees included in Kelly, profits are lower (which they also are in the reality), but the best strategies remain the same.


prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -153,25 +152,17 @@ def calculate_trades(
...
-                confidence=answer.confidence,
-            )
-            if market.has_token_pool()
-            else get_kelly_bet_simplified(
-                self.max_bet_amount,
-                market.current_p_yes,
-                answer.p_yes,
-                answer.confidence,
-            )
+        kelly_bet = get_kelly_bet_full(
```
kongzii, line 155: Open to opinions here 😄 

A few lines above, `outcome_token_pool = check_not_none(market.outcome_token_pool)` is used, which means that `get_kelly_bet_simplified` was never going to be used. 

Also `get_kelly_bet_simplified` doesn't accept fees which goes against this PR, but then if someones is going to create markets with huge fees, this other PR should solve that https://github.com/gnosis/prediction-market-agent-tooling/pull/488

evangriffiths: Makes sense, no objections your honour!

evangriffiths (approved):
LGTM!

PR #517, opened 2024-10-21
Simulate invalid markets

kongzii:


tests_integration_with_local_chain/markets/omen/test_redeem_invalid_markets.py
```diff
@@ -0,0 +1,195 @@
...
+
+    assert (
+        -bet_size < account_B_difference < 0
+    ), "Assumption was that B will get most of the money back but would incur a loss because he bought tokens at a higher price than 0.5."
+    assert (
+        account_C_difference > 0
+    ), "Assumption was that C will be profitable because he was buying the cheaper tokens."
+    assert (
+        account_C_difference > account_B_difference
+    ), "Assumption was that C will profit more than B, because he bought the cheaper tokens."
```
evangriffiths, line 195: This will necessarily be true if the previous 2 are true, so could remove

evangriffiths (approved):
Nice test for the invalid market agent!

PR #521, opened 2024-10-21
469 missing information in ipfsagentresult

gabrielfior:
-> Adding agent_name to structure being uploaded to IPFS


prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -492,7 +497,15 @@ def after_process_market(
         processed_market: ProcessedMarket,
     ) -> None:
         keys = APIKeys()
-        market.store_prediction(processed_market=processed_market, keys=keys)
+        reasoning = (
+            processed_market.answer.reasoning
+            if processed_market.answer.reasoning
+            else ""
+        )
+        agent_result = IPFSAgentResult(reasoning=reasoning, agent_name=self.agent_name)
```
gabrielfior, line 505: I moved this logic from OmenAgentMarket -> DeployableTraderAgent because the property `agent_name` belongs to `DeployableTraderAgent`.
One could simply pass the property as arg, but I would argue that the responsibility to build a IPFSAgentResult should be handled by the agent itself.

prediction_market_agent_tooling/markets/omen/data_models.py
```diff
@@ -798,7 +798,7 @@ def from_tuple(values: tuple[t.Any]) -> "ContractPrediction":
 
 class IPFSAgentResult(BaseModel):
     reasoning: str
-
+    agent_name: str | None
```
kongzii, line 801: Why would it be None?

prediction_market_agent_tooling/deploy/agent.py
kongzii: `IPFSAgentResult` is specific to Omen; for example, Metaculus wants only a comment without any IPFS upload.

That refactoring was made based on Evan's comment because there were a lot of Omen-uniquess in the DeployableTraderAgent.

Maybe it could be named `IPFSOmenAgentResult` or something to make it clearer in the code. 🤔 

Wdyt just passing `agent_name` into `store_prediction`?

gabrielfior: > Maybe it could be named IPFSOmenAgentResult or something to make it clearer in the code. 🤔

I would prefer to keep it as `IPFSAgentResult` - we can rename it once we have other agents uploading results as well.

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -441,10 +445,13 @@ def after_process_market(
         market_type: MarketType,
         market: AgentMarket,
         processed_market: ProcessedMarket | None,
+        agent_name: str,
```
kongzii, line 448: Not used (can just use self.agent_name as its part of this class)

```suggestion
```

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -611,12 +620,15 @@ def after_process_market(
...
```
kongzii, line 623: Also won't be neccessary

```suggestion
```

prediction_market_agent_tooling/deploy/agent.py
gabrielfior: Done

prediction_market_agent_tooling/deploy/agent.py
gabrielfior: Done

kongzii (approved):

PR #523, opened 2024-10-22
Run test_local_deployment in local chain

kongzii:


tests_integration_with_local_chain/deploy/test_deploy.py
```diff
@@ -0,0 +1,16 @@
...
+from web3 import Web3
+
+from prediction_market_agent_tooling.deploy.agent_example import DeployableCoinFlipAgent
+from prediction_market_agent_tooling.markets.markets import MarketType
+from prediction_market_agent_tooling.tools.contract import ContractBaseClass
+
+
+def test_local_deployment(local_web3: Web3) -> None:
+    with patch.object(ContractBaseClass, "get_web3", return_value=local_web3):
+        DeployableCoinFlipAgent(enable_langfuse=False).deploy_local(
```
kongzii, line 12: Now we can place bets, but I disabled the language; no need to spam it with this test.

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -458,10 +458,10 @@ def process_market(
         )
 
         placed_trades = []
-        if self.place_bet:
-            for trade in trades:
-                logger.info(f"Executing trade {trade} on market {market.id}")
+        for trade in trades:
+            logger.info(f"Executing trade {trade} on market {market.id} ({market.url})")
```
kongzii, line 462: Moved the `if self.place_bet` check just to see what bets would be placed during testing.

evangriffiths (approved):
🙌

PR #524, opened 2024-10-22
Fix tavily search for when days=None

evangriffiths:
This test:
```python
from prediction_market_agent_tooling.tools.tavily.tavily_search import tavily_search

days_topics = [
    ("news", 1),
    ("news", None),
    ("general", 1),
    ("general", None),
]
for topic, days in days_topics:
    try:
        news = tavily_search(
            query="what is the capital of paris?",
            days=days,
            max_results=1,
            topic=topic,
        )
    except Exception as exc:
        if exc.__class__.__name__ == "ValueError":
            continue  # our errors are expected
        print(f"Error for topic={topic}, days={days}")
```

previously would error if days=None:
```
Error for topic=news, days=None
Error for topic=general, days=None
```

Now it passes.


kongzii (approved):
Internally, they will default to `days=3` if we don't provide it https://github.com/tavily-ai/tavily-python/blob/master/tavily/tavily.py#L83

But I think this is fine until (if) they fix it in they Python client/API.

Issue #519, opened 2024-10-21
Add `supported_markets` field to `DeployableTraderAgent`


kongzii:
We are doing stuff like

```
if market_type != MarketType.OMEN:
    raise RuntimeError("This agent works only on Omen markets.")
```

in PMA. I think it'd be great to have it as a property directly on `DeployableTraderAgent`.

And then refactor PMA to use it.

kongzii:
Was needed in https://github.com/gnosis/prediction-market-agent-tooling/pull/525, so added it as part of that PR

PR #525, opened 2024-10-23
Split `DeployableTraderAgent` into `DeployablePredictionAgent`

kongzii:


prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -288,23 +291,13 @@ def __init__(
         super().__init__(enable_langfuse=enable_langfuse)
         self.place_bet = place_bet
 
-    def get_betting_strategy(self, market: AgentMarket) -> BettingStrategy:
```
kongzii, line 291: Everything around trading was moved to `DeployableTraderAgent`. `DeployablePredictionAgent` only does the prediction.

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -436,62 +403,28 @@ def process_market(
     ) -> ProcessedMarket | None:
         logger.info(f"Processing market {market.question=} from {market.url=}.")
 
-        self.before_process_market(market_type, market)
```
kongzii, line 439: Moved to `process_markets` function, so that it will be called correctly in the DeployableTraderAgent subclass.

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -436,62 +403,28 @@ def process_market(
...
-
         if verify_market and not self.verify_market(market_type, market):
             logger.info(f"Market '{market.question}' doesn't meet the criteria.")
-            self.update_langfuse_trace_by_processed_market(market_type, None)
```
kongzii, line 443: Moved to `after_process_market` hook.

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -528,6 +463,115 @@ def after_process_markets(self, market_type: MarketType) -> None:
         "Executes actions that occur after bets are placed."
 
     def run(self, market_type: MarketType) -> None:
+        if market_type not in self.supported_markets:
+            raise ValueError(
+                f"Only {self.supported_markets} are supported by this agent."
```
kongzii, line 468: I already had this in mind here https://github.com/gnosis/prediction-market-agent-tooling/issues/519

But now it's required here as well, because only METACULUS can be used with DeployablePredictionAgent, and rest of the markets is for DeployableTraderAgent.

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -528,6 +463,115 @@ def after_process_markets(self, market_type: MarketType) -> None:
...
+
+    def after_process_market(
+        self,
+        market_type: MarketType,
+        market: AgentMarket,
+        processed_market: ProcessedMarket | ProcessedTradedMarket | None,
+    ) -> None:
+        api_keys = APIKeys()
+        super().after_process_market(market_type, market, processed_market)
+        if isinstance(processed_market, ProcessedTradedMarket):
```
kongzii, line 576: Not nice, but idk how to do it better. `after_process_market` hook in this subclass will always receive `ProcessedTradedMarket,` but we need to accept `processed_market: ProcessedMarket | ProcessedTradedMarket | None` in the arguments because otherwise, mypy will it complain about subclass not being compatible with parent classs.

prediction_market_agent_tooling/markets/agent_market.py
```diff
@@ -228,13 +231,21 @@ def verify_operational_balance(api_keys: APIKeys) -> bool:
...
     def store_prediction(
-        self, processed_market: ProcessedMarket, keys: APIKeys
+        self, processed_market: ProcessedMarket | None, keys: APIKeys
     ) -> None:
         """
         If market allows to upload predictions somewhere, implement it in this method.
         """
         raise NotImplementedError("Subclasses must implement this method")
 
+    def store_trades(
```
kongzii, line 241: Now `store_prediction` is implemented on MetaculusAgentMarket and `store_trades` on the others.

tests/markets/test_markets.py
```diff
@@ -86,3 +90,11 @@ def test_get_markets(market_type: MarketType) -> None:
         limit=limit, sort_by=SortBy.NONE, filter_by=FilterBy.OPEN
     )
     assert len(markets) <= limit
+
+
+@pytest.mark.parametrize("market_type", list(MarketType))
+def test_market_is_covered(market_type: MarketType) -> None:
```
evangriffiths, line 96: I was concerned that the `supported_markets` check would get out of date if we added a new market type - but this test prevents that - nice!

prediction_market_agent_tooling/deploy/agent.py
evangriffiths: Could you do:
`processed_market: ProcessedMarket | None,`
and then do:
`assert isinstance(processed_market, ProcessedTradedMarket)`
?

prediction_market_agent_tooling/markets/agent_market.py
```diff
@@ -29,6 +29,9 @@
 
 class ProcessedMarket(BaseModel):
     answer: ProbabilisticAnswer
+
```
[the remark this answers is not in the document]
kongzii: Yes, but then we have `ProcessedTradedMarket` which adds `trades` field. 

For all to come nicely together, we need this `ProcessedMarket`, because:

`DeployablePredictionAgent`'s `process_market` returns `ProcessedMarket`
`DeployableTraderAgent`'s `process_market` returns `ProcessedTradedMarket`

If `ProcessedTradedMarket` is subclass of `ProcessedMarket`, then `process_market` methods are compatible.

If `DeployablePredictionAgent`'s `process_market` starts to return only `ProbabilisticAnswer`, then we'd need to add `trades` into `ProbabilisticAnswer`.

But not sure if I explain this properly.. 😄 If you have some other idea how to structure it, show me please. 🙏

prediction_market_agent_tooling/deploy/agent.py
kongzii: Sure, that should work. ` | ProcessedTradedMarket` in the argument typing is just to be explicit. Removed.

prediction_market_agent_tooling/markets/agent_market.py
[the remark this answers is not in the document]
evangriffiths: Hmm I had a look, so you could alias:
```
ProcessedMarket = ProbabilisticAnswer
```

but mypy 

or subclass
```
class ProcessedMarket(ProbabilisticAnswer):
    pass
```

but then constructing a `ProcessedTradedMarket` instance from a `ProcessedMarket` becomes more messy if you don't want to do `**processed_market.dict()`.

soooo okay, i retract my objection!

evangriffiths (approved):

PR #531, opened 2024-10-28
Allow for different configurations of DeployablePredictionAgent.same_market_trade_interval

evangriffiths:
`DeployablePredictionAgent.same_market_trade_interval` is now an instance of a subclass of `TradeInterval`. Can define fixed intervals (default behaviour unchanged from before), or dynamic intervals based on market length.

This is another feature that is required for us to be able to scale the number of markets an agent can look at in a single run - in this case so that it doesn't bet too frequently on long-running markets.


kongzii (approved):
Cool!

PR #532, opened 2024-10-28
Added limit as a configurable variable to DeployablePredictionAgent

gabrielfior:


prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -370,12 +371,13 @@ def answer_binary_market(self, market: AgentMarket) -> ProbabilisticAnswer | Non
     def get_markets(
         self,
         market_type: MarketType,
-        limit: int = MAX_AVAILABLE_MARKETS,
+        limit: int | None = None,
```
evangriffiths, line 374:
```suggestion
```

I'd prefer if this arg was removed, and we always use `self.n_markets_to_fetch` to set `limit`

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -370,12 +371,13 @@ def answer_binary_market(self, market: AgentMarket) -> ProbabilisticAnswer | Non
...
         sort_by: SortBy = SortBy.CLOSING_SOONEST,
         filter_by: FilterBy = FilterBy.OPEN,
     ) -> t.Sequence[AgentMarket]:
         cls = market_type.market_class
         # Fetch the soonest closing markets to choose from
+        limit = limit if limit else self.n_markets_to_fetch
         available_markets = cls.get_binary_markets(
             limit=limit, sort_by=sort_by, filter_by=filter_by
```
evangriffiths, line 382:
```suggestion
        available_markets = cls.get_binary_markets(
            limit=self.n_markets_to_fetch, sort_by=sort_by, filter_by=filter_by
```

prediction_market_agent_tooling/config.py
```diff
@@ -200,6 +200,6 @@ def check_if_is_safe_owner(self, ethereum_client: EthereumClient) -> bool:
         if not self.SAFE_ADDRESS:
             raise ValueError("Cannot check ownership if safe_address is not defined.")
 
-        s = Safe(self.SAFE_ADDRESS, ethereum_client)  # type: ignore[abstract]
+        s = Safe(self.SAFE_ADDRESS, ethereum_client)
         public_key_from_signer = private_key_to_public_key(self.bet_from_private_key)
-        return s.retrieve_is_owner(public_key_from_signer)
+        return bool(s.retrieve_is_owner(public_key_from_signer))
```
kongzii, line 205: It's not a good sign if the project that is typed and was working previously suddenly needs casting like this.

[Check out their repository](https://github.com/safe-global/safe-eth-py/tree/v6.0.0b41/safe_eth/safe):

Screenshot by Dropbox Capture

Please:

1. Set the minimum supported version in pyproject.toml to `safe-eth-py = "^6.0.0b41"` (we need to bump the version in pyproject, because if we rename imports from `gnosis` to `safe_eth`, then it won't work with the previous versions)
2. Fix our imports from `gnosis` to `safe_eth` anywhere where it's used (I see it's done somewhere, but not everywhere)

gabrielfior: done

prediction_market_agent_tooling/config.py
```diff
@@ -200,6 +200,6 @@ def check_if_is_safe_owner(self, ethereum_client: EthereumClient) -> bool:
         if not self.SAFE_ADDRESS:
             raise ValueError("Cannot check ownership if safe_address is not defined.")
 
-        s = Safe(self.SAFE_ADDRESS, ethereum_client)  # type: ignore[abstract]
+        s = SafeV141(self.SAFE_ADDRESS, ethereum_client)
```
kongzii, line 203: What is the reason for using SafeV141 instead of just Safe? Is it safe ( 😁 )? Safe in a sense that what if the agent is using different version of it?

gabrielfior: mypy complains because Safe is an abstract class.
SafeV141 is the latest version , and the one we create for all our agents (https://github.com/gnosis/prediction-market-agent-tooling/blob/ae454b324cae5fe433c3bdb093556c24419a8998/tests_integration_with_local_chain/safe/test_safe.py#L38)

PR #533, opened 2024-10-29
Expose more DeployablePredictionAgent.get_markets args as properties

evangriffiths:
No functional changes. It just enables an agent that bets on the newest markets, created after a certain date


prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -383,7 +391,10 @@ def get_markets(
         cls = market_type.market_class
         # Fetch the soonest closing markets to choose from
         available_markets = cls.get_binary_markets(
-            limit=self.n_markets_to_fetch, sort_by=sort_by, filter_by=filter_by
+            limit=self.n_markets_to_fetch,
+            sort_by=sort_by,
```
kongzii, line 395: Could you unify this, please:

`n_markets_to_fetch` is currently taken from the class' variable, `get_markets_sort_by` is defined as class' variable but it's used from this function arguments instead, and `filter_by` is only in this function arguments

Screenshot by Dropbox Capture

evangriffiths: Oops fixed. Have left `filter_by` as hard-coded `FilterBy.OPEN` (since there's not a use-case atm for having a different value)

kongzii (approved):

PR #506, opened 2024-10-16
Introduce `fees` into AgentMarket itself (not just OmenAgentMarket)

kongzii:

PR #534, opened 2024-10-30
DB cache in form of decorator

kongzii:


prediction_market_agent_tooling/tools/caches/db_cache.py
```diff
@@ -0,0 +1,334 @@
...
+def db_cache(
+    func: FunctionT,
+    *,
+    max_age: timedelta | None = None,
+    cache_none: bool = True,
+    api_keys: APIKeys | None = None,
+    ignore_args: Sequence[str] | None = None,
+    ignore_arg_types: Sequence[type] | None = None,
+) -> FunctionT:
+    ...
```
gabrielfior, line 65: I'm comparing this with other open-source libraries intended for caching, for ex, [beaker](https://beaker.readthedocs.io/en/latest/caching.html#decorator-api), which offers the exact same functionality (incl. Postgres connection). See usage below:

image

I'm curious if you considered using this or another existing package for caching instead of building from scratch? What led you to choose the 'from scratch' approach?

kongzii: Yeah, the problem is in the details; I see `which offers the exact same functionality,` but that's not correct.

If this db_cache magic becomes a problem, not as useful as I thought at first, brings bugs that aren't just accidental mistakes, etc., then I'm cool with removing it completely and just using something out of the box. But I can explain why I'd go with custom implementation atm:

I did take a look at [dogpile.cache](https://github.com/sqlalchemy/dogpile.cache) the next day you wrote to me on Slack about it, however:

- it isn't typed (anything decorated with their cache loses mypy type checking benefits)
- it doesn't support postgres out of the box
- it doesn't store input parameters (I think the data that we are storing are gold, for example, we have 23k Tavily requests in DB already -- we can easily query them and use them in the future)
- output is pickled (configurable for json serialization, but then half of the code from this PR would be needed anyway)
- it doesn't verify if cached output is still compatible with the current implementation -- I tried to create Pydantic model, called the function to cache it, and then I modified the Pydantic model (added another required field), dogpile doesn't care about it, it just returns the cached value without any complaints
- I don't see an option for `ignore_args` -- some stuff can be ignored for cache purposes, for example, `api_keys`

For the beaker, I didn't try it at runtime, but quickly looking at docs, it seems like most of the stuff from above applies as well.

It's also nice to know what's going on here and be able to modify it if needed. If it works well, perhaps we could do a standalone library from it 🤞  If not, dogspile, beaker, joblib is it.

prediction_market_agent_tooling/tools/caches/db_cache.py
```diff
@@ -0,0 +1,368 @@
...
+
+
+def db_cache(
```
kongzii, line 69: Just add `@db_cache` to any function or method and it should work like magic!

Optionally, specify some of the optional keyword arguments.

It supports:

- any input/output type -- native types, lists, pydantic models, nested combinations, etc.
- it stores input arguments in DB
- it stores the output in DB (along with input we can use it to gather data)
- it handles if the pydantic model isn't backward compatible with cached value, and it will re-compute the result
- decorated methods are properly typed
- we can specify ignored arguments/types, so we don't cache based on APIKeys for example

Screenshot by Dropbox Capture

prediction_market_agent_tooling/tools/caches/db_cache.py
```diff
@@ -0,0 +1,368 @@
...
+        if sqlalchemy_db_url is not None
+        else None
+    )
+
+    # Create table if it doesn't exist
+    if engine is not None:
+        SQLModel.metadata.create_all(engine)
+
+    @wraps(func)
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
```
kongzii, line 116: In this PR, there are some `Any` types that I usually fight against, but here they are really necessary, or it would be super hard to type it correctly at least, but because it's all used inside of `db_cache`, which in the end is cast to the type of decorated function anyway, it shouldn't matter.

prediction_market_agent_tooling/tools/is_invalid.py
```diff
@@ -54,9 +54,9 @@
 """
 
 
-@persistent_inmemory_cache
 @tenacity.retry(stop=tenacity.stop_after_attempt(3), wait=tenacity.wait_fixed(1))
 @observe()
+@db_cache
```
kongzii, line 59: Beware! Order matters here.

If we do

```
@observe()
@db_cache
def something(): ... 
```

then all calls will be observed in Langfuse.

If we do

```
@db_cache
@observe()
def something(): ... 
```

Then only the first (non-cached) call is observed in Langfuse.

prediction_market_agent_tooling/tools/relevant_news_analysis/relevant_news_analysis.py
```diff
@@ -55,7 +54,7 @@
 def analyse_news_relevance(
     raw_content: str,
     question: str,
-    date_of_interest: datetime,
+    date_of_interest: date,
```
kongzii, line 57: Changed to `date` to not invalidate cache because of difference in seconds or hours (or even milliseconds!)

prediction_market_agent_tooling/tools/relevant_news_analysis/relevant_news_analysis.py
```diff
@@ -91,19 +90,18 @@ def analyse_news_relevance(
...
     """
     Get relevant news since a given date for a given question. Retrieves
     possibly relevant news from tavily, then checks that it is relevant via
     an LLM call.
     """
+    news_since = date.today() - timedelta(days=days_ago)
     results = get_relevant_news_since(
         question=question,
-        days_ago=days_ago,
+        news_since=news_since,
```
kongzii, line 102: Passing `date` into `get_relevant_news_since` (and therefore `tavily_search`) to have cache based on absolute values, not relative ones. E.g. if we would call `days_ago=1` for a week, then it could be always used from cache, because the argument is still the same, yet the date of `1 day ago` is different!

prediction_market_agent_tooling/tools/relevant_news_analysis/relevant_news_analysis.py
```diff
@@ -140,7 +138,6 @@ def get_certified_relevant_news_since_cached(
     question: str,
     days_ago: int,
     cache: RelevantNewsResponseCache,
```
kongzii, line 140: I left the implementation of `RelevantNewsResponseCache` because there is news-specific logic that was just too hard to do in a general way using decorator. Evan's explanation: 

Screenshot by Dropbox Capture

prediction_market_agent_tooling/tools/tavily/tavily_models.py
```diff
@@ -23,62 +16,3 @@ class TavilyResponse(BaseModel):
     images: list[str]
     results: list[TavilyResult]
     response_time: float
-
-
-class TavilyResponseModel(SQLModel, table=True):
```
kongzii, line 28: Tavily cache can be removed completely, I'll just re-fill the new database table with old data

kongzii:
Also did a script to refill the data: https://github.com/gnosis/prediction-market-agent-tooling/pull/540 But it's too hacky to be merged, so just closed the PR to keep it at least somewhere.

prediction_market_agent_tooling/tools/caches/db_cache.py
gabrielfior: Thanks for explaining - fine by me to go with the custom implementation.

prediction_market_agent_tooling/tools/caches/db_cache.py
```diff
@@ -0,0 +1,368 @@
...
+    __tablename__ = "function_cache"
+    id: int | None = Field(default=None, primary_key=True)
+    function_name: str = Field(index=True)
+    full_function_name: str = Field(index=True)
+    # Args are stored to see what was the function called with.
+    args: Any = Field(sa_column=Column(JSONB, nullable=False))
+    # Args hash is stored as a fast look-up option when looking for cache hits.
+    args_hash: str = Field(index=True)
+    result: Any = Field(sa_column=Column(JSONB, nullable=False))
+    created_at: DatetimeUTC = Field(default_factory=utcnow, index=True)
```
gabrielfior, line 40: I was expecting a TTL column to help with cache eviction - if you don't have it, how do you handle "old", invalid cache entries?

gabrielfior: I see you are doing some complex operations in the [tests](https://github.com/gnosis/prediction-market-agent-tooling/pull/534/files#diff-34e124c81067f3a3ceda9acbf1b39a6c8280f213a6a05889dcef53fa58c8ccdcR273-R276), which could be simplified if using TTL.
Also, I don't follow why the use-case of Pydantic models changing is very important - do you expect this to happen often?

prediction_market_agent_tooling/tools/caches/db_cache.py
```diff
@@ -0,0 +1,368 @@
...
+    if isinstance(value, BaseModel):
+        return value.model_dump()
+    elif isinstance(value, dict):
+        return {k: convert_pydantic_to_dict(v) for k, v in value.items()}
+    elif isinstance(value, (list, tuple)):
+        return type(value)(convert_pydantic_to_dict(v) for v in value)
+    elif isinstance(value, set):
+        return {convert_pydantic_to_dict(v) for v in value}
+    else:
+        return value
```
gabrielfior, line 314: For simplicity - can't we always store JSON objects in the DB (in the `value` column)? This way, we can then afterwards do `model_validate_json`, wrapped by try-catch and handle it appropriately.
I don't 100% like this approach of having many types with if-elif-else - ultimately this is a design decision depending on personal taste, but from my understanding, JSON would be enough, i.e. `value` should not be Any, instead `str` (JSON-compatible).

prediction_market_agent_tooling/tools/caches/db_cache.py
kongzii: We have `max_age` filter, `TTL` is not implemented programmatically, but could be added if needed in the future: https://stackoverflow.com/questions/18187490/postgresql-delete-old-rows-on-a-rolling-basis

> I see you are doing some complex operations in the [tests](https://github.com/gnosis/prediction-market-agent-tooling/pull/534/files#diff-34e124c81067f3a3ceda9acbf1b39a6c8280f213a6a05889dcef53fa58c8ccdcR273-R276), which could be simplified if using TTL.

I don't see, how TTL would simplify that? In essence, it's the same thing as max_age, but that's not the point of that test.

> I don't follow why the use-case of Pydantic models changing is very important - do you expect this to happen often?

Well you changed Pydantic model last week [here](https://github.com/gnosis/prediction-market-agent/pull/511/files#diff-78157da54dc1e9c88afc902f8d614c4277a945d775d4f68855eef7bc075941c5R10), for example. Or I did a change [here](https://github.com/gnosis/prediction-market-agent-tooling/pull/506/files#diff-d54dae948eb76b2cc49dc63822131a9fc5d54a95dc288feb464585108bd4bc27R64) 3 weeks ago.

So it's not a daily thing but definitely good to have it handled.

prediction_market_agent_tooling/tools/caches/db_cache.py
kongzii: There are two separate things: what you wrote here and on Slack:

> can't we always store JSON objects in the DB (in the value column)

This PR is doing exactly that, storing `args` and `result` as JSONB in Postgres. Thanks to that, values in the database are properly formatted, and we can even query them as normal columns and filter by them:

Screenshot by Dropbox Capture

However! You made me think and indeed, `convert_pydantic_to_dict` can be removed fully, simplified [in this commit](https://github.com/gnosis/prediction-market-agent-tooling/pull/534/commits/c7e2fa361a9a7adcfec7740e176a7681630c581a).

> For ex, [here](https://github.com/gnosis/prediction-market-agent-tooling/pull/534/files#diff-472656f06870bd951ad8b55aa02f25d56ceee0873460f1f2ea78d682803a60d7R317) I would find amazing if we simply did model_validate_json instead of the complex logic here - if it fails, then an error is thrown.

The issue here is that the current implementation also handles nested structures and non-pydantic structures. Let's say the output type is `list[AgentMarket]`, then the JSONized form (doesn't matter if stored as JSONB or STRING in the database) will be like `[{'id': '...', 'question': ...}, ...]` and we can't call `model_validate_json` on lists, only on Pydantic models.

So `convert_cached_output_to_pydantic` is indeed necessary because the current implementation can handle all varies of outputs, as seen in the tests: 

Screenshot by Dropbox Capture

So it's either this complex implementation, or forcing us in the future to always return a single Pydantic model out of cached functions, which we already aren't doing:

Screenshot by Dropbox Capture

Screenshot by Dropbox Capture

prediction_market_agent_tooling/tools/caches/db_cache.py
gabrielfior: -> `max_age` offers the functionality I had in mind when referring to TTL. All good.
-> Fine with Pydantic model remarks

gabrielfior (approved):

Issue #510, opened 2024-10-17
Support for arbitrary categorical outcomes


kongzii:
Currently, all markets on Omen (created by Olas or our Replicator) are categorical, but the categories are `[Yes, No]`, so we consider them as binary markets (`True, False` outcomes).

However, most markets on Seer are truly categorical:

![Screenshot by Dropbox Capture](https://github.com/user-attachments/assets/bce0ee22-6762-4c06-9010-75323321a9d7)

But there are also some binary (although in a different order of outcomes, pay attention to indexes there! 😄 ):

![Screenshot by Dropbox Capture](https://github.com/user-attachments/assets/0a2c2f81-8b54-4942-b91b-d18d61ae7f36)

If we truly want to support Seer and make Presagio a bit more fancy, we need to add support for truly categorical markets at least (and later on also scalar markets...)

This is harder than it looks like because at many places in code we do stuff like `from_bool`, `outcome: bool`, or all agents are producing `p_yes` instead of the `outcome: probability` pairs... so it needs a lot of thinking about how to structure PMAT and how to make current agents still usable and at least one new agent to test this out.

Issue #502, opened 2024-10-14
Add SeerAgentMarket to PMAT


gabrielfior:
- This should implement the same functions as OmenAgentMarket
- Tests
- Also add option to run agent on `Seer` from PMA's `run_agent`

Issue #501, opened 2024-10-14
Seer - add subgraph to SubgraphHandler


gabrielfior:
- Check if subgraph contains required functions (e.g. `getMarkets`, `getTradesFromUser`, etc) and add them to SubgraphHandler
- If not, deploy new subgraph
- Additional reference - see (https://thegraph.com/explorer/subgraphs/AAA1vYjxwFHzbt6qKwLHNcDSASyr1J1xVViDH8gTMFMR?view=Query&chain=arbitrum-one) for Algebra pools, we might need a different source for all pools for a given (outcomeToken, collateralToken) pair

kongzii:
Instead of adding seer to `OmenSubgraphHandler`, I vote for a new class, `SeerSubgraphHandler`. [See this comment ](https://github.com/gnosis/prediction-market-agent-tooling/pull/490/files#r1795285026)for the reasoning; TLDR is that OmenSubgraphHandler is already crowded, and it is hard to tell what's going on.

gabrielfior:
Agree with `SeerSubgraphHandler `

PR #535, opened 2024-10-30
501 seer add subgraph to subgraphhandler

gabrielfior:
-> Created SeerSubgraphHandler with methods for fetching markets and fetching associated pools (will be needed for later placement of bets)
-> Wrote tests


prediction_market_agent_tooling/markets/seer/seer_subgraph_handler.py
```diff
@@ -0,0 +1,96 @@
...
+from prediction_market_agent_tooling.markets.base_subgraph_handler import (
+    BaseSubgraphHandler,
+)
+from prediction_market_agent_tooling.markets.seer.data_models import (
+    SeerMarket,
+    SeerPool,
+)
+from prediction_market_agent_tooling.tools.hexbytes_custom import HexBytes
+
+BINARY_MARKETS_FILTER = {"parentMarket": ADDRESS_ZERO.lower()}
```
kongzii, line 13: I randomly tried one non-binary market, and it also has parentMarket set to address zero:

Screenshot by Dropbox Capture

kongzii: I'm afraid we will need a filter based on `Yes, No` outcomes (or whatever they are using for Yes/No markets), instead of `binary` markets. I agree that market like `Who will win the 2024 U.S. Presidential Election?` with outcomes `Kamala, Trump` is binary, but our code doesn't support it at all.

We need to work on https://github.com/gnosis/prediction-market-agent-tooling/issues/510 if we want to support Seer markets.

prediction_market_agent_tooling/markets/base_subgraph_handler.py
```diff
@@ -0,0 +1,54 @@
...
+from prediction_market_agent_tooling.config import APIKeys
+from prediction_market_agent_tooling.loggers import logger
+from prediction_market_agent_tooling.tools.singleton import SingletonMeta
+
+T = t.TypeVar("T", bound=BaseModel)
+
+
+class BaseSubgraphHandler(metaclass=SingletonMeta):
+    sg: Subgrounds
+    keys: APIKeys
```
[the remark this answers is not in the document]
gabrielfior: ChatGPT says
```
In general, attributes that vary by instance are safer as instance attributes, especially for subclasses. Only define an attribute at the class level if it’s truly static or shared across all instances.
```
So I think `keys` and `sg` should be specific to each instance (in this case, there is only one per type since it's a singleton, but still an instance), so keeping this inside `__init__` only. Implementing this.

prediction_market_agent_tooling/markets/seer/seer_subgraph_handler.py
gabrielfior: What about doing the following in this ticket:
-> Let's get only binary markets on Seer - those are defined as 
1. `outcomes` contains "Invalid" AND `len(outcomes)` == 3

-> However, the issues you raised (Kamala/Trump is different than YES/NO, breaking PMAT) still exist and will be handled on #501 .

kongzii: > -> However, the issues you raised (Kamala/Trump is different than YES/NO, breaking PMAT) still exist and will be handled on https://github.com/gnosis/prediction-market-agent-tooling/issues/501 .

Isn't this PR for issue #501? 😄 At least it's called `501 seer add subgraph to subgraphhandler`

gabrielfior: Edited the comment, sorry!

kongzii: I see! Fine by me in that case. This in this PR isn't used anywhere yet, so I'm okay if you want to go with the form you mentioned above (`outcomes contains "Invalid" AND len(outcomes) == 3`).

However, I just wanted to at least raise my concern that I'm afraid that #510 is quite a big issue to take (imagine how many places we have hardcoded stuff like `p_yes,` `from_bool_outcome,` etc.).

Even #502 is blocked until #510 is completed.

So if you want to trade on Seer after #510 is implemented, fine by me, but if you want to trade on Seer as quickly as possible, then doing a filter here for `Yes, No` is a small change, and you can have Seer-trading relatively quickly I guess.

gabrielfior: Added some more methods (see `get_binary_markets` and `get_bicategorical_markets`) to address this.

prediction_market_agent_tooling/markets/base_subgraph_handler.py
```diff
@@ -0,0 +1,54 @@
...
+
+        self.keys = APIKeys()
+
+    def _parse_items_from_json(
+        self, result: list[dict[str, t.Any]]
+    ) -> list[dict[str, t.Any]]:
+        """subgrounds return a weird key as a dict key"""
+        items = []
+        for result_chunk in result:
+            for k, v in result_chunk.items():
```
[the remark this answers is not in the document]
kongzii: Common that's too big neat picking! 😄 And if anything one could just use `result_chunk.values()` to not have the extra variable at all!

kongzii (approved):

PR #543, opened 2024-11-04
Tweak to lanfguse trace script

evangriffiths:


examples/monitor/match_bets_with_langfuse_traces.py
```diff
@@ -206,6 +206,10 @@ def get_outcome_for_trace(
                 bets_with_traces.append(ResolvedBetWithTrace(bet=bet, trace=trace))
 
         print(f"Number of bets since {start_time}: {len(bets_with_traces)}\n")
+
+        if len(bets_with_traces) == 0:
+            continue
```
evangriffiths, line 211: To quickly see if the script still works, I like to set  `start_time` to 1 day ago. But then some agents have 0 bets, in which case the script will throw a divide-by-zero error. This fixes that.

kongzii (approved):

PR #546, opened 2024-11-06
Initialize db_cache engine in wrapper

kongzii:


prediction_market_agent_tooling/tools/caches/db_cache.py
```diff
@@ -91,33 +91,22 @@ def decorator(func: FunctionT) -> FunctionT:
...
+    @wraps(func)
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
+        # If caching is disabled, just call the function and return it
+        if not api_keys.ENABLE_CACHE:
+            return func(*args, **kwargs)
 
-    engine = (
-        create_engine(
-            sqlalchemy_db_url.get_secret_value(),
+        engine = create_engine(
```
kongzii, line 100: `engine` needs to be created inside of the `wrapper` function. Otherwise, it lives in the scope of `@db_cache`, and then when think-thoroughly tries to do multi-processing, it fails to be pickled.

evangriffiths (approved):

PR #547, opened 2024-11-07
Lower amount of DB connections

kongzii:
deploy please


prediction_market_agent_tooling/config.py
```diff
@@ -53,7 +53,7 @@ class APIKeys(BaseSettings):
 
     SQLALCHEMY_DB_URL: t.Optional[SecretStr] = None
 
-    ENABLE_CACHE: bool = True
+    ENABLE_CACHE: bool = False
```
kongzii, line 56: By default set to False, to not error out cached functions if postgres access isn't set.

prediction_market_agent_tooling/tools/caches/db_cache.py
```diff
@@ -90,22 +91,25 @@ def decorator(func: FunctionT) -> FunctionT:
         return decorator
 
     api_keys = api_keys if api_keys is not None else APIKeys()
+    wrapped_engine = InitialiseNonPickable(
```
kongzii, line 94: The issue is that engine should be shared within all db_cache decorated functions, because otherwise there are too many connections and current Postgres DB doesn't make it.

But if it's just initialised here, it will error out in ThinkThoroughly because it isn't pickable.

This is a workaround around that.

gabrielfior (approved):

gabrielfior (approved):

PR #552, opened 2024-11-11
Added gnosis_rpc_config, google_credentials_filename and chain_id to …

gabrielfior:
…BaseSettings


prediction_market_agent_tooling/config.py
```diff
@@ -203,3 +204,37 @@ def check_if_is_safe_owner(self, ethereum_client: EthereumClient) -> bool:
...
         return s.retrieve_is_owner(public_key_from_signer)
+
+
+class RPCConfig(BaseSettings):
+    model_config = SettingsConfigDict(
+        env_file=".env", env_file_encoding="utf-8", extra="ignore"
+    )
+
+    GNOSIS_RPC_URL: t.Optional[str] = Field(default="https://rpc.gnosischain.com")
+    CHAIN_ID: t.Optional[int] = Field(default=100)
```
kongzii, line 215: If there are default values, it never will be None, so the optionality and properties can be removed.

Otherwise good idea to switch it to BaseSettings as well 👍

prediction_market_agent_tooling/config.py
```diff
@@ -203,3 +204,37 @@ def check_if_is_safe_owner(self, ethereum_client: EthereumClient) -> bool:
...
+    def chain_id(self) -> int:
+        return check_not_none(self.CHAIN_ID, "CHAIN_ID missing in the environment.")
+
+
+class CloudCredentials(BaseSettings):
+    model_config = SettingsConfigDict(
+        env_file=".env", env_file_encoding="utf-8", extra="ignore"
+    )
+
+    GOOGLE_APPLICATION_CREDENTIALS: t.Optional[str] = None
```
kongzii, line 233: I don't mind if we have multiple subclasses of BaseSettings (APIKeys, CloudCredentials, RPCConfig), just wondering if you have some specific pattern in mind that I could follow as well? Like when to create a new class instead of adding to existing?

prediction_market_agent_tooling/tools/contract.py
```diff
@@ -391,8 +387,8 @@ class ContractOnGnosisChain(ContractBaseClass):
     Contract base class with Gnosis Chain configuration.
     """
 
-    CHAIN_ID = GNOSIS_NETWORK_ID
-    CHAIN_RPC_URL = GNOSIS_RPC_URL
+    CHAIN_ID = ChainID(RPCConfig().chain_id)
```
kongzii, line 390: `ChainID` can be typed directly as return type of `chain_id` property

prediction_market_agent_tooling/tools/google_utils.py
```diff
@@ -51,3 +55,26 @@ def search_google(
...
+def get_private_key_from_gcp_secret(
+    secret_id: str,
+    project_id: str = "582587111398",  # Gnosis AI default project_id
+    version_id: str = "latest",
+) -> PrivateKey:
+    # If credentials filename specified, use that, else read using default credentials path.
+    google_application_credentials_filename = (
+        CloudCredentials().google_application_credentials
+    )
+    if google_application_credentials_filename is not None:
```
kongzii, line 69: It's coded as 

```
@property
    def google_application_credentials(self) -> str:
        return check_not_none(
            self.GOOGLE_APPLICATION_CREDENTIALS,
            "GOOGLE_APPLICATION_CREDENTIALS missing in the environment.",
        )
```

so `google_application_credentials` will never be None (it will error out if it's not defined). 

Can you either

- change how `CloudCredentials` handles None of google_application_credentials, then this if statement make sense
- or remove the if statement here, as it will never be used?

prediction_market_agent_tooling/config.py
gabrielfior: Good point, done

prediction_market_agent_tooling/tools/contract.py
gabrielfior: Good point, done

prediction_market_agent_tooling/tools/google_utils.py
gabrielfior: I removed the `@property`, since NONE values are also OK - it just takes the default.
Can you please double-check?

kongzii (approved):

PR #553, opened 2024-11-12
Raise OutOfFunds  in withdraw_wxdai_to_xdai_to_keep_balance

kongzii:


prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -1339,7 +1340,7 @@ def withdraw_wxdai_to_xdai_to_keep_balance(
     )
 
     if current_balances.wxdai < need_to_withdraw:
-        raise ValueError(
+        raise OutOfFundsError(
```
kongzii, line 1343: This can happen if agent has all his money locked in bets. If we raise OutOfFundsError instead of ValueError, these errors are reported to our slack channel only once a week, so it won't spam us.

gabrielfior (approved):

PR #555, opened 2024-11-13
First attempt at DB_Manager

gabrielfior:


prediction_market_agent_tooling/tools/caches/db_cache.py
```diff
@@ -97,16 +98,9 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
-            json_serializer=json_serializer,
-            json_deserializer=json_deserializer,
-            pool_size=1,
-        )
-
-        # Create table if it doesn't exist
-        SQLModel.metadata.create_all(engine)
+        with DBManager(api_keys).get_connection() as conn:
+            # Create table if it doesn't exist
+            SQLModel.metadata.create_all(conn, tables=[SQLModel.metadata.tables[FunctionCache.__tablename__]],)
```
gabrielfior, line 103: @kongzii maybe you have a batter idea on what's going on here:

-> similar to the previous version, here too the `FunctionCache` table is created (if not exists) every time a cached function is called. I don't like this very much, in my head, it should be done in the very beginning and then never again (my attempt at solving this, 
This part is making the tests fail (note that tests were passing earlier). Do you see a way to make this "table init" somewhere else, so that it's not called every time? Thanks!

gabrielfior: Note that tests, when executed individually, pass :(

prediction_market_agent_tooling/tools/caches/db_cache.py
```diff
@@ -152,7 +146,7 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
         if return_type is not None and contains_pydantic_model(return_type):
             is_pydantic_model = True
 
-        with Session(engine) as session:
+        with DBManager(api_keys).get_session() as session:
```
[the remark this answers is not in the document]
gabrielfior: True - my mistake, refactoring this.

prediction_market_agent_tooling/tools/db/db_manager.py
```diff
@@ -0,0 +1,72 @@
...
+    _instances: dict[str, "DBManager"] = {}
+
+    def __new__(cls, api_keys: APIKeys | None = None) -> "DBManager":
+        sqlalchemy_db_url = (api_keys or APIKeys()).sqlalchemy_db_url
+        secret_value = sqlalchemy_db_url.get_secret_value()
+        url_hash = hashlib.md5(secret_value.encode()).hexdigest()
+        if url_hash not in cls._instances:
+            instance = super(DBManager, cls).__new__(cls)
+            cls._instances[url_hash] = instance
+        return cls._instances[url_hash]
```
gabrielfior, line 25: Can you please add a comment around this? Just for later on docs

prediction_market_agent_tooling/tools/db/db_manager.py
```diff
@@ -0,0 +1,72 @@
...
+                table
+                for sqlmodel_table in sqlmodel_tables
+                if not self.cache_table_initialized.get(
+                    (
+                        table := SQLModel.metadata.tables[
+                            cast(str, sqlmodel_table.__tablename__)
+                        ]
+                    ).name
+                )
+            ]
```
gabrielfior, line 65: Can you dumb this down a bit? It's a little too complex, I'm afraid.

gabrielfior: GPT suggests this
```
def create_tables(self, sqlmodel_tables: Sequence[type[SQLModel]] | None = None) -> None:
    # Determine tables to create
    if sqlmodel_tables is not None:
        tables_to_create = []
        for sqlmodel_table in sqlmodel_tables:
            table_name = sqlmodel_table.__tablename__
            if table_name is not None:
                table = SQLModel.metadata.tables[table_name]
                if not self.cache_table_initialized.get(table.name):
                    tables_to_create.append(table)
    else:
        tables_to_create = None

    # Create tables in the database
    SQLModel.metadata.create_all(self._engine, tables=tables_to_create)

    # Update cache to mark tables as initialized
    if tables_to_create:
        for table in tables_to_create:
            self.cache_table_initialized[table.name] = True
```

tests/tools/db/test_db_manager.py
```diff
@@ -0,0 +1,35 @@
...
+        are_same_instance = db1 is db2
+        are_not_same_instance = db1 is not db3
+
+    assert are_same_instance, "DBManager created more than one instance!"
+    assert (
+        are_not_same_instance
+    ), "DBManager returned isntance with a different SQLALCHEMY_DB_URL!"
+
+
+# TODO: Gabriel what was the goal in this test?
```
gabrielfior, line 30: I wanted to assert that `get_connection` or `get_session` could be used when calling `SQLModel.metadata.create_all` - this is tested in other tests, so feel free to delete this.

PR #558, opened 2024-11-16
Fixes for Safe

kongzii:


prediction_market_agent_tooling/tools/web3_utils.py
```diff
@@ -263,7 +263,11 @@ def send_function_on_contract_tx_using_safe(
     )
     safe_tx.sign(from_private_key.get_secret_value())
     safe_tx.call()  # simulate call
-    tx_hash, tx = safe_tx.execute(from_private_key.get_secret_value())
+    eoa_nonce = web3.eth.get_transaction_count(eoa_public_key)
+    tx_hash, tx = safe_tx.execute(
+        from_private_key.get_secret_value(),
+        tx_nonce=eoa_nonce,
```
kongzii, line 269: Passing nonce manually to the execute method was the main motivation for this PR.

Problem was that if we don't pass it, Safe library switches `retry` to True:

Screenshot by Dropbox Capture

And if retry is used, then:

Screenshot by Dropbox Capture

Which causes two problems for us:

1. `error` log spams us unnecessarily on Slack
2. It's yet-another-retry-logic that slows down execution, because we are already using `tenacity` in our code

tests_integration_with_local_chain/conftest.py
```diff
@@ -53,7 +54,7 @@ def local_web3(
 
 @pytest.fixture(scope="module")
 def local_ethereum_client(local_web3: Web3) -> EthereumClient:
-    return EthereumClient()
+    return EthereumClient(URI(local_web3.provider.endpoint_uri))  # type: ignore
```
kongzii, line 57: Pass correct URI to the EthereumClient (it didn't work with Tenderly locally) and use `local_ethereum_client` in tests instead of re-initialising it

tests_integration_with_local_chain/safe/test_safe.py
```diff
@@ -103,5 +104,7 @@ def fetch_omen_open_binary_market_with_enough_liquidity(
     limit: int = 1, liquidity_bigger_than: Wei = xdai_to_wei(xdai_type(5))
 ) -> list[OmenMarket]:
     return OmenSubgraphHandler().get_omen_binary_markets(
-        limit=limit, resolved=False, liquidity_bigger_than=liquidity_bigger_than
+        limit=limit,
+        question_opened_after=utcnow(),
```
kongzii, line 108: resolved=False doesn't mean that market is open for trading, we need `question_opened_after=utcnow()` + `liquidity_bigger_than`

gabrielfior: Can we pass in an `open_for_trading` bool as arg? I think that would be cleaner - I think I'm the one that used `resolved=False` :)

kongzii: For that, we already have the `get_omen_binary_markets_simple` method, which accepts some enums and sets the where statements accordingly, `get_omen_binary_markets` is meant to represent what subgraph actually supports.

Switched to `get_omen_binary_markets_simple` here.

gabrielfior (approved):

PR #548, opened 2024-11-07
545 strategy backtests add financial metrics

gabrielfior:
-> Added Sharpe ratio, mean daily return and annual volatility as extra variables when selecting strategies

PR #561, opened 2024-11-19
Use Optuna and KFold to find best strategy for agent

kongzii:
I'll wait for https://github.com/gnosis/prediction-market-agent-tooling/pull/548 to be completed and then merge it here before doing final conclusions, but the idea is to

1. Use Optuna to find best strategy instead of manually hard-coding a ton of possibilities
2. Use KFold validation to try to find strategy that isn't just over-fitted to data


examples/monitor/match_bets_with_langfuse_traces.py
```diff
@@ -205,123 +273,70 @@ def get_outcome_for_trace(
...
+        kf = KFold(n_splits=5, shuffle=False)
+        min_abs_diff: float | None = None
+        best_study: optuna.Study | None = None
 
-        for strategy_idx, strategy in enumerate(strategies):
-            # "Born" agent with initial funding, simulate as if he was doing bets one by one.
-            starting_balance = 50.0
-            agent_balance = starting_balance
-            simulated_outcomes: list[SimulatedOutcome] = []
+        for n_fold, (train_index, test_index) in enumerate(kf.split(bets_with_traces)):
```
gabrielfior, line 293: I'm wondering if this time split is the optimal one, since accuracy varies with time, and accuracy directly dictates the "optimal" betting strategy - I would assume we want to be more conservative overall if accuracy is small, and riskier if accuracy is higher.

image

For that reason - have you considered using `TimeSeriesSplit`? Thoughts?

gabrielfior:
Great idea using Optuna for optimizing the strategy space and finding the optimal parameters. 💪

examples/monitor/match_bets_with_langfuse_traces.py
kongzii: `bets_with_traces` is sorted by created time ascending and KFold is used with shuffle=False, so it should have been fine 🤔 However that `TimeSeriesSplit` is more explicit about it, I'll switch to it

prediction_market_agent_tooling/config.py
```diff
@@ -188,14 +188,14 @@ def model_dump_public(self) -> dict[str, t.Any]:
         return {
             k: v
             for k, v in self.model_dump().items()
-            if APIKeys.model_fields[k].annotation not in SECRET_TYPES and v is not None
+            if self.model_fields[k].annotation not in SECRET_TYPES and v is not None
```
kongzii, line 191: Just issue with newest Pydantic where `model_fields` aren't class variables anymore.

prediction_market_agent_tooling/markets/data_models.py
```diff
@@ -161,3 +161,16 @@ class SharpeOutput(BaseModel):
     annualized_volatility: float
     mean_daily_return: float
     annualized_sharpe_ratio: float
+
+
+class SimulatedLifetimeDetail(BaseModel):
```
kongzii, line 166: Converting remaining dictionary into pydantic model.

prediction_market_agent_tooling/markets/omen/data_models.py
```diff
@@ -499,13 +499,17 @@ class OmenBet(BaseModel):
...
-    collateralAmountUSD: USD
     feeAmount: Wei
     outcomeIndex: int
     outcomeTokensTraded: Wei
     transactionHash: HexBytes
     fpmm: OmenMarket
 
+    @property
+    def collateral_amount_usd(self) -> USD:
+        # Convert manually instad of using the field `collateralAmountUSD` available on the graph, because it's bugged, it's 0 for non-xDai markets.
```
kongzii, line 510: This is nasty! Same issue experienced Presagio guys, but I forgot about it here: https://github.com/SwaprHQ/presagio/issues/109.

Because of this, some of the calculations were off when sDai markets were used.

prediction_market_agent_tooling/tools/caches/inmemory_cache.py
```diff
@@ -1,17 +1,57 @@
...
+
 T = TypeVar("T", bound=Callable[..., Any])
 
 
-def persistent_inmemory_cache(func: T) -> T:
+@overload
+def persistent_inmemory_cache(
+    func: None = None,
+    *,
+    in_memory_cache: bool = True,
```
kongzii, line 18: Just added option to disable in-memory cache and keep only disk cache, in case of many large objects it comes handy.

prediction_market_agent_tooling/tools/httpx_cached_client.py
```diff
@@ -3,7 +3,10 @@
 
 class HttpxCachedClient:
     def __init__(self) -> None:
-        storage = hishel.FileStorage(ttl=3600, check_ttl_every=600)
+        storage = hishel.FileStorage(
+            ttl=24 * 60 * 60,
```
kongzii, line 7: 10 min wasn't enough for full day of experiments 😄

prediction_market_agent_tooling/tools/transaction_cache.py
```diff
@@ -9,8 +9,8 @@
 
 class TransactionBlockCache:
     def __init__(self, web3: Web3):
-        self.block_number_cache = dc.Cache("block_cache_dir")
-        self.block_timestamp_cache = dc.Cache("timestamp_cache_dir")
+        self.block_number_cache = dc.Cache(".cache/block_cache_dir")
```
kongzii, line 12: `.cache` dir is ignored in `.gitignore`

examples/monitor/match_bets_with_langfuse_traces.py
```diff
@@ -93,18 +101,25 @@ def get_outcome_for_trace(
...
-            HexAddress(HexStr(market.id)), block_number=tx_block_number
+        try:
+            bet_tx_block_number = tx_block_cache.get_block_number(actual_placed_bet.id)
+        except tenacity.RetryError as e:
+            if isinstance(e.last_attempt.exception(), TransactionNotFound):
+                return None
+            raise
+        # We need market at state before the bet was placed, otherwise it wouldn't be fair (outcome price is higher from the original bet at `bet_tx_block_number`)
+        market_before_placing_bet = get_omen_market_by_market_id_cached(
+            HexAddress(HexStr(market.id)), block_number=bet_tx_block_number - 1
```
kongzii, line 112: That `- 1` is the second nasty thing!

We need market at state **before** the bet was placed, otherwise, after it's placed, even the same investment won't give the same result, as probabilities (prices) are moved from the original bet! 😱 

Also added test for this in `test_omen.py`.

examples/monitor/match_bets_with_langfuse_traces.py
```diff
@@ -115,77 +130,234 @@ def get_outcome_for_trace(
...
+            raise RuntimeError(
+                "Groups of bets should be sorted ascending by created time."
+            )
+
+        # During optimization, put more weight into recent weeks.
+        n = len(metrics)
+        weights = (
+            [0.5 + i * (1.0 - 0.5) / (n - 1) for i in range(n)] if n > 1 else [1.0]
+        )
+        return [w * m.maximize for w, m in zip(weights, metrics)]
```
kongzii, line 269: Optuna allows for multi-objective optimization.

Instead of optimizing against one bucket of markets, where some huge lucky bet can over-fit the strategy into un-profitable waters, optimize multiple buckets (the bucket is defined by `group_datetime` above, currently one bucket = one week).

However, put more priority into recent weeks (multiply by weights).

examples/monitor/match_bets_with_langfuse_traces.py
gabrielfior: That's true, great catch.
But bear in mind this is still an approximation - for example, if at block 1 the pool outcome tokens are [YES,NO]=[10,10], and at block 2 you place a trade for 10 Y tokens, then you can calculate the exact amount of tokens you get placing this bet by using block 1.
However, if 3 additional people place bets at block 2 (before you), your price is different - and there is no easy way (to my knowledge) to figure out the token outcome sizes right before your transaction.
But I don't expect this case to happen, there are not that many high-frequency bets being placed.

kongzii: Yay, that's true! Theoretically, the `original trade` can still be in favor -- in case someone bought the opposite outcome within the same block, before the agent.

So, this is more fair for the simulation, but not completely. But I also agree that's not the issue here.

(the original trade can be only more profitable, because we need to pass `expected_shares` to the contract, so if someone bought the same outcome within the block, then `outcomeTokensToBuy` will be lower than expected and transaction will revert)

Screenshot by Dropbox Capture

Screenshot by Dropbox Capture

examples/monitor/match_bets_with_langfuse_traces.py
```diff
@@ -115,77 +130,234 @@ def get_outcome_for_trace(
...
+        p_yes_mse=p_yes_mse,
+        total_bet_amount=total_bet_amount,
+        total_bet_profit=total_bet_profit,
+        total_simulated_amount=total_simulated_amount,
+        total_simulated_profit=total_simulated_profit,
+        roi=roi,
+        simulated_roi=simulated_roi,
+        sharpe_output_original=sharpe_output_original,
+        sharpe_output_simulation=sharpe_output_simulation,
+        maximize=total_simulated_profit,  # Metric to be maximized for.
```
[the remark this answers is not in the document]
kongzii: I took a look at Sharpe. The issue with Nans is if there are less than 3 days of bets inside of the fold.

It doesn't matter that there are 100 bets in two days, it will be nan because in the `wallet_daily_balance_df`, there will be only 1 day of balance (n/a for the first day, X for the second day based on the first day), and then `std` calculation returns nan.

As a quick fix I just added condition to skip the fold with less than 3 days of bets.

> i.e. not selecting one particular combination that has a spike in Sharpe but probably an outline

Yeah that's kind of what I wanted to achieve by using multi-optimization strategy. Optimizing Sharpe (or profit or whatever) per-week, instead for one huge group of bets, such that one combination on lucky best won't create un-profitable outliers.

---

I tried to run optimization against Sharpe, but in the end profit is way lower, I guess that's expected as it isn't optimized against it anymore.

How would we justify that we optimize for Sharpe instead of profit?

Screenshot by Dropbox Capture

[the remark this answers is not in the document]
gabrielfior: Wow, that's a lot less profit :(

I'm fine with optimizing for profit for now, and iterating upon this decision later on.

kongzii:
1 failed test, not related to this PR:

```
FAILED tests/markets/omen/test_omen.py::test_get_positions_1 - ValueError: Number of condition ids for markets 86 and positions 87 are not equal.
```

production agents are also failing 😱 will merge this and check it out

PR #564, opened 2024-11-29
Fix pending arbitration filter

kongzii:


prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -226,7 +226,6 @@ def _build_where_statements(
         category: str | None,
     ) -> dict[str, t.Any]:
         where_stms: dict[str, t.Any] = {
-            "isPendingArbitration": False,
```
kongzii, line 229: Kleros is testing arbitration process on this market https://presagio.pages.dev/markets?id=0xC93A18546Ce264BAA89d0182b35E732A1CE006D2.

Because of this hard-coded filter, our tests (and all agents in production) are failing:

Screenshot by Dropbox Capture

Because:

Screenshot by Dropbox Capture

I checked it and if `isPendingArbitration=true` then `answerFinalizedTimestamp=null`, so we should be fine with just removing this.

Screenshot by Dropbox Capture

gabrielfior: In my head, there are 2 cases:
1. "Normal operation", market is open, trading occurs, "answerFinalizedTimestamp" == null (id 0xaa00421b9a0c259ff797cb13ea805e9ad33321ad)

image

2. Now there is a market being arbitrated (id 0xc93a18546ce264baa89d0182b35e732a1ce006d2). Presagio doesn't allow us to place bets (so market is CLOSED). 
Observe that `answerFinalizedTimestamp` is null.

image

My understanding is that we don't want this market either, because it's not open. Hence I suggest we add the filter below

```
isPendingArbitration_not: true # allowed values Null, False
```

Thoughts?

kongzii: > "Normal operation", market is open, trading occurs, "answerFinalizedTimestamp" == null

`"answerFinalizedTimestamp" == null` doesn't mean that the market is open; it means that no answer has been submitted yet or that it is being arbitrated.

`isPendingArbitration_not: true` would result in the same error as I described above, ie it wouldn't return the market with `PendingArbitration: True`.

For open markets, we are filtering by the opening timestamp: 

Screenshot by Dropbox Capture

And the question that isn't open yet can not be arbitrated. So, all should be fine.

prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -757,9 +765,12 @@ def get_omen_question_filters(
 
         if with_answers is not None:
             if with_answers:
```
gabrielfior, line 767: Why the double if?

kongzii: Because it means something different if you pass `None` vs. if you pass `False`

with_answers=None --> default behaviour, we don't care whatever the question has or hasn't any answers

with_answers=False --> we want question with no answers only

gabrielfior (approved):

PR #566, opened 2024-12-03
Send data along with xDai

kongzii:


prediction_market_agent_tooling/tools/web3_utils.py
```diff
@@ -297,12 +297,15 @@ def send_xdai_to(
...
     value: Wei,
+    data_text: Optional[str] = None,
     tx_params: Optional[TxParams] = None,
     timeout: int = 180,
 ) -> TxReceipt:
     from_address = private_key_to_public_key(from_private_key)
 
     tx_params_new: TxParams = {"value": value, "to": to_address}
+    if data_text is not None:
+        tx_params_new["data"] = Web3.to_bytes(text=data_text)
```
kongzii, line 308: In this PR, just add an option to provide arbitrary text data into the transaction.

In PMA's PR, I'll use it in function like `send_message`, `decode_message`, etc., where zlib stuff will be used as well.

gabrielfior (approved):

PR #569, opened 2024-12-04
Add NFT contract

kongzii:


gabrielfior:
Just an observation -> ERC721 (like ERC20) has a `approve` method - this allows party B to spend (aka transfer) tokens owned by party A.
In that sense, since we just want to transfer the token from the agent to someone (if at all), then I think it's fine, but convincing the agent to call `approve` might be an attack vector, hence it's good NOT to have that functionality in there.

gabrielfior (approved):

PR #570, opened 2024-12-05
Try to fix amount of connecitons

kongzii:
deploy please

### DIFF
--- a/prediction_market_agent_tooling/tools/db/db_manager.py
+++ b/prediction_market_agent_tooling/tools/db/db_manager.py
@@ -33,9 +33,7 @@ class DBManager:
             sqlalchemy_db_url.get_secret_value(),
             json_serializer=json_serializer,
             json_deserializer=json_deserializer,
-            pool_size=1,
-            pool_recycle=3600,
-            echo=True,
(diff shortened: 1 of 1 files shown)


prediction_market_agent_tooling/tools/db/db_manager.py
```diff
@@ -28,16 +28,17 @@ def __new__(cls, api_keys: APIKeys | None = None) -> "DBManager":
         return cls._instances[url_hash]
 
     def __init__(self, api_keys: APIKeys | None = None) -> None:
+        if hasattr(self, "_initialized"):
```
gabrielfior, line 31: It's not clear to me why `self._initialized = True` is needed.
I expected DBManager to be a singleton, if that's the case, then this shouldn't be necessary.

kongzii: Well, my words exactly!

I asked o1-preview to fix whatever is wrong with DBManager, and these changes are what it did. And now it seems it works 🤷 

Screenshot by Dropbox Capture

gabrielfior (approved):

PR #578, opened 2025-01-09
Temporary disable Polymarket in resolution finding and allow Python 3.12

kongzii:


pyproject.toml
```diff
@@ -10,7 +10,7 @@ buy_omen = "scripts.bet_omen:buy"
 sell_omen = "scripts.bet_omen:sell"
 
 [tool.poetry.dependencies]
-python = ">=3.10,<3.12"
+python = ">=3.10,<3.13"
```
kongzii, line 13: We have Python 3.12 in the CI, but it was disallowed here. Probably some dependency had issues in the past, but all looks good now!

PR #585, opened 2025-01-20
Auto deposit and withdraw using cowpy

kongzii:


prediction_market_agent_tooling/tools/tokens/auto_deposit.py
```diff
@@ -0,0 +1,156 @@
...
+        # In this case, we can use deposit function directly, no need to go through DEX.
+        auto_deposit_depositable_wrapper_erc20(
+            collateral_token_contract, amount_wei, api_keys, web3
+        )
+
+    elif isinstance(collateral_token_contract, ContractERC4626BaseClass):
+        auto_deposit_erc4626(collateral_token_contract, amount_wei, api_keys, web3)
+
+    elif isinstance(collateral_token_contract, ContractERC20BaseClass):
+        auto_deposit_erc20(collateral_token_contract, amount_wei, api_keys, web3)
```
kongzii, line 37: New option, `auto_deposit_erc20` that uses CoW to allow the deposit of any ERC20 token.

prediction_market_agent_tooling/tools/tokens/auto_deposit.py
```diff
@@ -0,0 +1,156 @@
...
+    )
+
+    # If we have enough of the collateral token, we don't need to deposit.
+    if collateral_token_balance >= amount_wei:
+        return
+
+    # If we don't have enough, we need to deposit the difference.
+    left_to_deposit = Wei(amount_wei - collateral_token_balance)
+    logger.info(
+        f"Depositing {wei_to_xdai(left_to_deposit)} {collateral_token_contract.symbol()}."
```
kongzii, line 60: Only new log in this function.

prediction_market_agent_tooling/tools/tokens/auto_deposit.py
```diff
@@ -0,0 +1,156 @@
...
+            )
+        # And finally, we can deposit the asset token into the erc4626 vault directly as well, without DEX.
+        collateral_token_contract.deposit_asset_token(left_to_deposit, api_keys, web3)
+
+    else:
+        # Otherwise, we need to go through DEX.
+        auto_deposit_erc20(collateral_token_contract, asset_amount_wei, api_keys, web3)
+
+
+def auto_deposit_erc20(
```
kongzii, line 109: This function is 100% new

gabrielfior: Why can;t we use cow for all deposits/withdraws, instead of having special cases for sDAI and wxDAI? I believe cow supports it all.

prediction_market_agent_tooling/tools/tokens/main_token.py
```diff
@@ -0,0 +1,13 @@
...
+from prediction_market_agent_tooling.tools.contract import (
+    ContractDepositableWrapperERC20OnGnosisChain,
+)
+
+# This is the token where agents will hold their funds,
+# except for a small portion that will be kept in the native token of the network to pay for the fees.
+# If changed, then keep in mind that we assume this token is equal to 1 USD.
+KEEPING_ERC20_TOKEN = ContractDepositableWrapperERC20OnGnosisChain(
+    address=WRAPPED_XDAI_CONTRACT_ADDRESS
+)
```
gabrielfior, line 13: Can we have something like
```
KEEPING_ERC20_TOKEN = WrappedxDaiContract() 
```
I don't follow why we need the `DepositableWrapper`

prediction_market_agent_tooling/tools/tokens/auto_deposit.py
kongzii: That's a good question.

First, CoW doesn't support xDai (native token), so we need the special case of xDai --> wxDai (and back) using the `deposit` (or withdraw) function of wxDai.

Then, we could do the wxDai --> sDai (and back) using CoW, but I didn't think it's worth it (given the fact that we already have the implementation ready using the native contract methods), because:

- it wouldn't be testable in `tests_integration_with_local_chain`, as CoW is going through 3rd party APIs 
- deposit/withdraw is instant, but with CoW we need to wait until the swap happens

prediction_market_agent_tooling/tools/tokens/main_token.py
kongzii: Because wxDai is a special case of ERC20 with `deposit` and `withdraw` method to/from native token.[ See this comment.](https://github.com/gnosis/prediction-market-agent-tooling/blob/7bba967e44cfdcbfbe567008b276da0badae9ae7/prediction_market_agent_tooling/tools/contract.py#L265)

prediction_market_agent_tooling/tools/tokens/auto_deposit.py
gabrielfior: Thanks. Makes sense.

gabrielfior (approved):

PR #589, opened 2025-01-23
Pre-commit hooks

kongzii:


.github/workflows/python_ci.yaml
```diff
@@ -80,3 +80,16 @@ jobs:
...
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ./.github/actions/python_prepare
+      - name: Install Dependencies
+        run: |
+          pip install pre-commit
+      - name: Run Pre-Commit
+        run: |
+          pre-commit run --all-files
```
kongzii, line 95: pre-commit needs to be installed locally (see `install_hooks.sh`), but as a last resort, also run it in CI, so at least we won't miss it by accident if the key gets leaked.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,16 @@
...
+    rev: v5.0.0
+    hooks:
+      - id: detect-private-key
+      - id: check-added-large-files
+        args: ['--maxkb=5000']
+  - repo: local
+    hooks:
+      - id: detect-web3-private-key
+        name: Detect Web3 Private Key
+        entry: python scripts/detect_web3_private_keys.py
```
kongzii, line 12: After review, I'll try to get this into https://github.com/pre-commit/pre-commit-hooks, if they don't accept, I'll create a separate repo for it, so we can share it across our repos

prediction_market_agent_tooling/gtypes.py
```diff
@@ -82,5 +82,5 @@ def secretstr_to_v1_secretstr(s: SecretStr | None) -> SecretStrV1 | None:
 
 
 def int_to_hexbytes(v: int) -> HexBytes:
-    # Example: 1 -> HexBytes("0x0000000000000000000000000000000000000000000000000000000000000001").
+    # Example: 1 -> HexBytes("0x0000000000000000000000000000000000000000000000000000000000000001"). # web3-private-key-ok
```
kongzii, line 85: It looks like private keys don't have any strong format. I had to add this `# web3-private-key-ok` so we can tell the script that these aren't real private keys. 

@gabrielfior please check out scripts/detect_web3_private_keys.py, if you know a better way of how to detect private keys, I'll be happy 😄

gabrielfior (approved):

PR #590, opened 2025-01-23
Added Docs Support for ReadTheDocs

vanshika-srivastava:


mkdocs.yml
```diff
@@ -0,0 +1,7 @@
+site_name: Prediction Market Agent Tooling
+nav:
+  - Home: index.md
+  - Contributing: contributing.md
```
kongzii, line 4: Shouldn't these paths be prefixed with `docs/`?

kongzii (approved):

mkdocs.yml
vanshika-srivastava: right! let me fix

vanshika-srivastava: Hey, just checked it should be the above only. not prefixed.

PR #601, opened 2025-02-03
fixed safe creation transaction error

skundu42:


prediction_market_agent_tooling/tools/web3_utils.py
```diff
@@ -307,24 +307,25 @@ def send_xdai_to(
...
-
-    # We need gas and gasPrice here (and not elsewhere) because we are not calling
-    # contract.functions.myFunction().build_transaction, which autofills some params
-    # with defaults, incl. gas and gasPrice.
-    gas = web3.eth.estimate_gas(tx_params_new)
-    tx_params_new["gas"] = int(
-        gas * 1.5
-    )  # We conservatively overestimate gas here, knowing it will be returned if unused
-    tx_params_new["gasPrice"] = web3.eth.gas_price
+
```
kongzii, line 320: This `_prepare_tx_params` that is removed was meant to do this boilerplate stuff around transactions.

It's updating nonce [here](https://github.com/gnosis/prediction-market-agent-tooling/blob/05869ef630075ebd0531c2c2b1e7dde17a640570/prediction_market_agent_tooling/tools/web3_utils.py#L152).

And `gasPrice` was already set in this `send_xdai_to` method.

Wouldn't be enough to do in the previous implementation this:

```
tx_params_new["gasPrice"] = web3.eth.gas_price * 1.2 # Increase gas by 20% to have some bandwich.
``` 

instead of re-implementing all of this? 🤔

PR #602, opened 2025-02-03
Fix is_open attribute on OmenMarket

kongzii:


prediction_market_agent_tooling/markets/omen/data_models.py
```diff
@@ -245,7 +246,7 @@ def has_valid_answer(self) -> bool:
 
     @property
     def is_open(self) -> bool:
-        return self.currentAnswer is None
+        return self.close_time > utcnow()
```
gabrielfior, line 249: Can you provide more info? Why is this needed? Why does this change make sense? Why was the prev implementation working until now and suddently not anymore?

gabrielfior (approved):

prediction_market_agent_tooling/markets/omen/data_models.py
kongzii: [yep comment is here](https://github.com/gnosis/prediction-market-agent-tooling/pull/602/files#diff-6aaf85d952cb99bbd129bef760f5e94cd4477d698103727006d150de8ff9780fR235)

PR #609, opened 2025-02-07
Use warning if claiming failed

kongzii:


prediction_market_agent_tooling/markets/omen/omen_resolving.py
```diff
@@ -50,7 +50,7 @@ def claim_bonds_on_realitio_questions(
         except Exception as e:
             if not skip_failed:
                 raise e
-            logger.error(
+            logger.warning(
```
kongzii, line 53: We are getting error alerts because of this, but then on second try it just works 🤷 So we can deescalate this I guess

gabrielfior (approved):

PR #612, opened 2025-02-07
Trading Seer markets using cow

gabrielfior:


prediction_market_agent_tooling/markets/agent_market.py
```diff
@@ -297,10 +297,13 @@ def get_outcome_str(self, outcome_index: int) -> str:
             )
 
     def get_outcome_index(self, outcome: str) -> int:
-        try:
-            return self.outcomes.index(outcome)
-        except ValueError:
-            raise ValueError(f"Outcome `{outcome}` not found in `{self.outcomes}`.")
+        outcome_lower = outcome.lower()
```
kongzii, line 300: Why is this lowering required now with seer?

prediction_market_agent_tooling/markets/omen/data_models.py
```diff
@@ -48,9 +48,9 @@ def construct_presagio_url(market_id: HexAddress) -> str:
 
 
 def get_boolean_outcome(outcome_str: str) -> bool:
-    if outcome_str == OMEN_TRUE_OUTCOME:
+    if outcome_str.lower() == OMEN_TRUE_OUTCOME.lower():
```
kongzii, line 51: Why is Omen market change required on Seer PR?

prediction_market_agent_tooling/markets/seer/data_models.py
```diff
@@ -34,17 +44,98 @@ class SeerParentMarket(BaseModel):
...
+    blockTimestamp: int
+    hasAnswers: bool | None
+    payoutReported: bool
+    payoutNumerators: list[int]
+
+    @property
+    def has_valid_answer(self) -> bool:
+        # We assume that, for the market to be resolved as invalid, it must have
+        # 1. 3 outcomes (Yes, No, Invalid), 2. Invalid is the last one and 3. Invalid numerator is 1.
+        if len(self.outcomes) != 3:
```
kongzii, line 74: Instead of checking for 3, would it work with `if self.outcomes[-1] != "Invalid":`?

What I'm thinking is that `has_valid_answer` could be outcome-agnostic right away. Because we want to support arbitrary outcomes in future.

prediction_market_agent_tooling/markets/seer/data_models.py
```diff
@@ -34,17 +44,98 @@ class SeerParentMarket(BaseModel):
...
+        max_idx = self.payoutNumerators.index(1)
+
+        bool_outcome = get_boolean_outcome(self.outcomes[max_idx])
+        if bool_outcome:
+            return Resolution.YES
+        return Resolution.NO
+
+    @property
+    def is_binary(self) -> bool:
+        return len(self.outcomes) == 3
```
kongzii, line 102:
```suggestion
        # 3 because Seer has also third, `Invalid` outcome.
        return len(self.outcomes) == 3
```

prediction_market_agent_tooling/markets/seer/seer_subgraph_handler.py
```diff
@@ -70,35 +81,102 @@ def filter_binary_markets(markets: list[SeerMarket]) -> list[SeerMarket]:
...
+                where_stms["payoutReported"] = True
+            case FilterBy.NONE:
+                pass
+            case _:
+                raise ValueError(f"Unknown filter {filter_by}")
+
+        if not include_conditional_markets:
+            where_stms["parentMarket"] = ADDRESS_ZERO.lower()
+
+        where_stms["outcomes_contains"] = [INVALID_OUTCOME]
```
kongzii, line 107: Whis is hardcoding this necessary?

prediction_market_agent_tooling/markets/omen/data_models.py
gabrielfior: Thank you, valid point. 
I will address these concerns ref `outcomes` on a different comment.

prediction_market_agent_tooling/markets/seer/data_models.py
gabrielfior: I made it a bit more general (no limit on number of outcomes) - there are markets (for ex this one, [multi-categorical](https://app.seer.pm/markets/100/who-will-win-or-advance-to-the-next-round-in-the-re-run-of-round-1-of-the-2024-romanian-presidential-election?outcome=C%C4%83lin+Georgescu)) where outcomes > 3 and invalid is always last one.
Please check if updated logic makes sense to you.

prediction_market_agent_tooling/markets/agent_market.py
gabrielfior: discussed elsewhere, closing this

prediction_market_agent_tooling/markets/agent_market.py
```diff
@@ -284,8 +284,7 @@ def has_successful_resolution(self) -> bool:
     def has_unsuccessful_resolution(self) -> bool:
         return self.resolution in [Resolution.CANCEL, Resolution.MKT]
 
-    @staticmethod
-    def get_outcome_str_from_bool(outcome: bool) -> OutcomeStr:
+    def get_outcome_str_from_bool(self, outcome: bool) -> OutcomeStr:
```
gabrielfior, line 287: Changed this from `static` to regular function because `Seer` requires self.

prediction_market_agent_tooling/markets/blockchain_utils.py
```diff
@@ -0,0 +1,82 @@
+from web3 import Web3
```
gabrielfior, line 1: Simply moved some common functions for Seer and Omen to this file.

prediction_market_agent_tooling/markets/seer/data_models.py
```diff
@@ -30,21 +47,174 @@ class CreateCategoricalMarketsParams(BaseModel):
     token_names: list[str] = Field(..., alias="tokenNames")
 
 
+class SeerOutcomeEnum(str, Enum):
```
gabrielfior, line 50: This is one big change. Instead of dealing with binary outcomes than can be "Yes", "YES", "Invalid", "INVALID result", etc - I abstracted that away and am now dealing with Enums (+, -, neutral (for invalid)).
That's a bit different from Omen, where we tried to consolidate around bool outcomes.

prediction_market_agent_tooling/markets/seer/seer.py
```diff
@@ -1,22 +1,294 @@
...
+
+    @classmethod
+    def get_tiny_bet_amount(cls) -> BetAmount:
+        return BetAmount(amount=SEER_TINY_BET_AMOUNT, currency=cls.currency)
+
+    def get_position(self, user_id: str, web3: Web3 | None = None) -> Position | None:
+        """
+        Fetches position from the user in a given market.
+        We ignore the INVALID balances since we are only interested in binary outcomes.
+        """
```
gabrielfior, line 139: Position on Seer is simpler, it equals the token balance for all the outcome tokens.
One important point (that we neglect here) is the option to add liquidity to a OUTCOME_TOKEN/sDAI pool, where you don't get `outcome_tokens` but rather LP tokens.
This doesn't affect our trader agents initially but deserves consideration. Created a [ticket](https://github.com/gnosis/prediction-market-agent-tooling/issues/614).

tests/test_betting_strategy.py
```diff
@@ -57,8 +60,8 @@ def test_rebalance() -> None:
     mock_existing_position = Position(
         market_id="0x123",
         amounts={
-            OmenAgentMarket.get_outcome_str_from_bool(True): mock_amount,
-            OmenAgentMarket.get_outcome_str_from_bool(False): mock_amount,
+            OutcomeStr(OMEN_TRUE_OUTCOME): mock_amount,
+            OutcomeStr(OMEN_FALSE_OUTCOME): mock_amount,
```
gabrielfior, line 64: Adjusted this because method not `static` anymore

tests_integration_with_local_chain/markets/seer/test_seer_agent_market.py
```diff
@@ -0,0 +1,70 @@
...
+        Mock(return_value=MOCK_QUOTE),
+    ), patch(
+        "prediction_market_agent_tooling.markets.agent_market.AgentMarket.can_be_traded",
+        Mock(return_value=True),
+    ), patch(
+        "prediction_market_agent_tooling.markets.seer.seer.SeerAgentMarket.has_liquidity_for_outcome",
+        Mock(return_value=True),
+    ), patch(
+        "prediction_market_agent_tooling.tools.cow.cow_manager.CowManager.swap",
+    ):
```
gabrielfior, line 54: We mock mostly the Cow integrations (`swap` and `quote`) because we don't want to actually execute a trade nor price an asset on every test. This test mostly gives coverage to the remaining logic of `place_bet`, whereas we trust the tests already present in the `cow_py` repo and don't retest the cow interactions.

kongzii: If I understand the logic correctly, only thing that requires having real xDai and doing real transactions is the ending `swap` call in `place_bet` method.

Wdyt about mocking nothing,  and simply doing it like in https://github.com/gnosis/prediction-market-agent-tooling/blob/860332eab88dc9e04e13d58319baf0fcd606f1fc/tests_integration_with_local_chain/tools/test_cow_order.py#L29

Then, everything is tested up to the very end. We still can not assert that bet was placed correctly, but that we can't do now neither.

prediction_market_agent_tooling/markets/omen/data_models.py
kongzii: I didn't find any comment regarding why Omen-related change was needed?

prediction_market_agent_tooling/markets/seer/data_models.py
```diff
@@ -30,21 +47,166 @@ class CreateCategoricalMarketsParams(BaseModel):
...
+    def close_time(self) -> DatetimeUTC:
+        return DatetimeUTC.to_datetime_utc(self.opening_ts)
+
+    @property
+    def created_time(self) -> DatetimeUTC:
+        return DatetimeUTC.to_datetime_utc(self.block_timestamp)
+
+    @property
+    def current_p_yes(self) -> Probability:
+        # ToDo - Write test
```
kongzii, line 178:
```suggestion
```

prediction_market_agent_tooling/markets/seer/data_models.py
```diff
@@ -30,21 +47,166 @@ class CreateCategoricalMarketsParams(BaseModel):
...
+        price_data = {}
+        for idx in range(len(self.outcomes)):
+            wrapped_token = self.wrapped_tokens[idx]
+            price = self._get_price_for_token(
+                token=Web3.to_checksum_address(wrapped_token)
+            )
+            price_data[idx] = price
+
+        yes_idx = self.outcome_as_enums[SeerOutcomeEnum.POSITIVE]
+        price_yes = price_data[yes_idx] / sum(price_data.values())
```
kongzii, line 189: Shouldn't it be like in Omen case https://github.com/gnosis/prediction-market-agent-tooling/blob/b921c7c66ee0c88090abe3d324c0112a090f908b/prediction_market_agent_tooling/markets/omen/data_models.py#L355 ?

tests_integration/markets/seer/test_seer_data_model.py
```diff
@@ -0,0 +1,42 @@
...
+    no_token = market.wrapped_tokens[no_idx]
+    no_price = market._get_price_for_token(Web3.to_checksum_address(no_token))
+
+    invalid_idx = market.outcome_as_enums[SeerOutcomeEnum.NEUTRAL]
+    invalid_token = market.wrapped_tokens[invalid_idx]
+    invalid_price = market._get_price_for_token(Web3.to_checksum_address(invalid_token))
+
+    current_p_yes = market.current_p_yes
+    expected_p_yes = yes_price / (yes_price + no_price + invalid_price)
+    assert np.isclose(current_p_yes, expected_p_yes, atol=0.01)
```
kongzii, line 35: A good rule of thumb to follow, if the test logic copy-paste the logic of the function we want to test (what I mean here is that `current_p_yes` is also using `_get_price_for_token` and does `price_data[yes_idx] / sum(price_data.values())`, so it's the same thing, just inside of a for loop), then it creates a false feeling of having all covered by tests, but if the copy-pasted logic is flagged, we miss it.

But, no idea how to get it here in a different way 😄 Maybe in such case there doesn't have to be anything at all 🤷‍♂️

prediction_market_agent_tooling/tools/cow/cow_manager.py
```diff
@@ -0,0 +1,112 @@
...
+            return asyncio.run(
+                get_order_quote(
+                    order_quote_request=order_quote_request,
+                    order_side=order_side,
+                    order_book_api=self.order_book_api,
+                )
+            )
+
+        except UnexpectedResponseError as e1:
+            logger.error(f"Unexpected response error: {e1.message}")
```
kongzii, line 89: Please move this error log below the `if "NoLiquidity"` clause.

If I see correctly, `NoLiquidityAvailableOnCowException` is something we except to happen quite often (0 liquidity in market), if we do an error log for something that is expected, our alerts channel will go crazy. 😄

prediction_market_agent_tooling/markets/seer/data_models.py
kongzii: I tried to test it out, but I'm getting zerodivisionerror:

Screenshot by Dropbox Capture

In Omen, we handle it like this:

Screenshot by Dropbox Capture

prediction_market_agent_tooling/markets/seer/seer_subgraph_handler.py
gabrielfior: Just went through the Seer code, found this
```
    function createCategoricalMarket(CreateMarketParams calldata params) external returns (address) {
        require(params.outcomes.length >= 2, "Outcomes count must be 2 or more");

        string[] memory encodedQuestions = new string[](1);
        encodedQuestions[0] =
            encodeRealityQuestionWithOutcomes(params.marketName, params.outcomes, params.category, params.lang);

        return createMarket(
            params,
            params.marketName,
            InternalMarketConfig({
                encodedQuestions: encodedQuestions,
                outcomeSlotCount: params.outcomes.length + 1, // additional outcome for Invalid Result.
                templateId: REALITY_SINGLE_SELECT_TEMPLATE
            })
        );
    }
```
hence no point in filtering invalid outcomes, they are always present. Closing this.

prediction_market_agent_tooling/markets/seer/seer.py
```diff
@@ -1,22 +1,294 @@
...
+            api_keys.public_key,
+            # Use `public_key`, not `bet_from_address` because transaction costs are paid from the EOA wallet.
+            sum_wxdai=False,
+        ) > xdai_type(0.001)
+
+    @staticmethod
+    def from_data_model(model: SeerMarket) -> "SeerAgentMarket":
+        return SeerAgentMarket(
+            id=model.id.hex(),
+            description=None,
```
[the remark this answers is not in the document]
gabrielfior: No descriptions in Seer - will copy that from Omen, thanks!

prediction_market_agent_tooling/markets/seer/data_models.py
gabrielfior: Removed

tests_integration_with_local_chain/markets/seer/test_seer_agent_market.py
gabrielfior: Like the suggestion. Refactored it in a similar way, see https://github.com/gnosis/prediction-market-agent-tooling/pull/612/commits/e477e1fe5b8f36350e9d00cabd9ab62b72aa2018

prediction_market_agent_tooling/markets/seer/data_models.py
gabrielfior: No strong opinion here - just had to move away from bools. YES,NO,INVALID ok to me, refactored.

prediction_market_agent_tooling/markets/seer/seer.py
```diff
@@ -1,22 +1,294 @@
...
+
+    def has_liquidity_for_outcome(self, outcome: bool) -> bool:
+        outcome_token = self.get_wrapped_token_for_outcome(outcome)
+        try:
+            CowManager().get_quote(
+                collateral_token=self.collateral_token_contract_address_checksummed,
+                buy_token=outcome_token,
+                sell_amount=xdai_to_wei(
+                    xdai_type(1)
+                ),  # we take 1 xDai as a baseline value for common trades the agents take.
```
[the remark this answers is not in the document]
gabrielfior: As a rule-of-thumb, if liquidity is available for X xDAI, it should also be available for x xDAI if x < X.

Looking at PredictionProphetGPT4 below and ordering by trade_amount DESC, I find the max amount to be 5 xDAI, and the average to be 0.95. So I think the amount 1 xDAI is on the right order of magnitude.

image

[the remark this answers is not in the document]
gabrielfior: To your question, if we trade 25 xDAI (which we don't at the moment), the transaction could fail, but at least Cow would throw a nice exception and we would see it in the logs.

prediction_market_agent_tooling/markets/seer/seer.py
```diff
@@ -1,22 +1,294 @@
...
+    ) -> str:
+        api_keys = api_keys if api_keys is not None else APIKeys()
+        if not self.can_be_traded():
+            raise ValueError(
+                f"Market {self.id} is not open for trading. Cannot place bet."
+            )
+
+        # We add an additional check below since we want to be certain that liquidity exists for the outcome token
+        # being traded.
+        if not self.has_liquidity_for_outcome(outcome):
```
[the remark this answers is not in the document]
gabrielfior: If we never bet on Invalid, then this check is not necessary.
Removing this.

prediction_market_agent_tooling/markets/seer/data_models.py
gabrielfior: I refactored it a bit (see [commit](https://github.com/gnosis/prediction-market-agent-tooling/pull/612/commits/2e580d5276a278b649d303ccb6a76c44866d2f95)), now in your case you would see `price=0`. The reason is that Cow couldn't find a quote for any outcome price, hence the division by zero (sum == 0).

It isn't like Omen in this case because of 2 things:
1. We use Cow for quoting - hence we ask Cow for the price of each outcome token, and normalize the odds with all the prices. So if YES costs $0.7, NO costs $0.4 and INVALID costs $0.3, `p_yes` would be 0.7/(0.7+0.4+0.3) = 0.63.

2. For Omen, we have one pool YES/NO, hence the price is just the ratio. For Seer, we can have N pools for each outcome token (e.g. we can have YES/sDAI on Swapr v2, YES/sDAI on Swapr v3, YES/sDAI on Uni v3, and technically also YES/NO on all the previous platforms as well). Hence we delegate the pricing to Cow.

prediction_market_agent_tooling/markets/omen/data_models.py
gabrielfior: Can't find that anymore, moving the discussion here:
---
OMEN outcomes were referenced by the betting strategies, now I removed that dependency as well.
Long story short - reverted Omen definitions, no changes here.

prediction_market_agent_tooling/tools/cow/cow_manager.py
```diff
@@ -0,0 +1,121 @@
...
+                    order_book_api=self.order_book_api,
+                )
+            )
+
+        except UnexpectedResponseError as e1:
+            if "NoLiquidity" in e1.message:
+                raise NoLiquidityAvailableOnCowException(e1.message)
+            return self._log_and_raise_exception(e1)
+        except Exception as e:
+            return self._log_and_raise_exception(e)
```
gabrielfior, line 101: @kongzii if you have a better idea on how to make mypy stop complaining about try-except constructs (see https://github.com/python/mypy/issues/8964#issuecomment-640295130), feel free to edit directly

tests_integration/markets/seer/test_seer_data_model.py
gabrielfior: Decided to remove it - simply replicating the logic doesn't add too much value, and maintaining it would also be a pain.

prediction_market_agent_tooling/tools/cow/cow_manager.py
kongzii: Debug is fine but fyi debug logs are not stored in gcp. We will see them only locally. If you want to inspect this in GCP Logs, it needs to be info or warning.

kongzii (approved):

PR #645, opened 2025-03-14
Support for arbitrary collaterak tokens

kongzii:
deploy please

---

This is honestly a huge PR, but most of the changes are uninteresting brute-force changes of types and stuff.

Main benefits out of this PR:

- We know when we work with USD, when Token (with unknown value, but can be converted to USD using provided methods), when with OutcomeTokens
- PMAT should be ready to operate on markets with arbitrary tokens, but I don't enable it in this PR (will do in next)
- Mypy and Runtime will complain if we process together different types
- Some typing fixes, logic fixes (we were using outcome tokens where tokens were expected), removal of dead/unused code

Main assumption out of this PR:

- agent thinks in USD, agent decides how much USD to bet on market and doesn't care about the underlying collateral (although fancy agents arbitrating collateral's value could care about it!)

What to look at:

- Check out the docstring of the new [_GenericValue class here.](https://github.com/gnosis/prediction-market-agent-tooling/blob/4e94c1ac8d6bcbc43421d9ad4f81e3707790b260/prediction_market_agent_tooling/tools/_generic_value.py#L15)
- Check out what [types inherits from it in here](https://github.com/gnosis/prediction-market-agent-tooling/blob/f28eda2e42efc7556774de5af8b6f48e250e1e40/prediction_market_agent_tooling/gtypes.py#L30)
- `Token`, `xDai`, `Wei`, `OutcomeWei` are all different types now, and mypy will complain accordingly. Also, runtime will fail if we try to work with them together. 
- `Token` and `OutcomeToken` (and respectively for Wei) are different types because it's a big difference if we try to sell 1 Token (ie collateral) vs. 1 OutcomeToken (we also had a bug related to this and that's why I differentiated them in the end)
- Check [out new methods in AgentMarket](https://github.com/gnosis/prediction-market-agent-tooling/blob/f28eda2e42efc7556774de5af8b6f48e250e1e40/prediction_market_agent_tooling/markets/agent_market.py#L191) and their implementations, namely `get_token_in_usd`, `get_usd_in_token`, `get_sell_value_of_outcome_token`
- @gabrielfior Please check [the betting strategy file, if it looks good to you  ](https://github.com/gnosis/prediction-market-agent-tooling/blob/f28eda2e42efc7556774de5af8b6f48e250e1e40/prediction_market_agent_tooling/deploy/betting_strategy.py#L1)


prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -1058,8 +1080,10 @@ def omen_fund_market_tx(
     )
 
     if auto_deposit:
-        # In auto-depositing, we need to deposit the original `funds`, e.g. we can deposit 2 xDai, but receive 1.8 sDai, so for the funding we will use `amount_to_fund`.
-        auto_deposit_collateral_token(collateral_token_contract, funds, api_keys, web3)
+        # In auto-depositing, we need to deposit the original `funds_wei`, e.g. we can deposit 2 xDai, but receive 0.1 GNO, so for the funding we will use `amount_to_fund`.
+        auto_deposit_collateral_token(
+            collateral_token_contract, funds_wei, api_keys, web3
+        )
```
gabrielfior, line 1086: The easiest thing to do is to have auto-deposit from a fixed token, e.g. sDAI or wxDAI, to the collateral token (e.g. GNO), and convert them using Cow.
That way we support anything that can be traded on Cow.

prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -767,8 +778,12 @@ def omen_buy_outcome_tx(
...
     collateral_token_contract = market_contract.get_collateral_token_contract(web3)
 
-    # In case of ERC4626, obtained (for example) sDai out of xDai could be lower than the `amount_wei`, so we need to handle it.
-    amount_wei_to_buy = collateral_token_contract.get_in_shares(amount_wei, web3)
+    # Function receives how much xDai to bet, but if collateral is GNO for example, actual buying will be like 2 xDai -> 0.001 GNO.
+    amount_wei_to_buy = convert_to_another_token(
+        amount_wei,
+        from_token=KEEPING_ERC20_TOKEN.address,
+        to_token=collateral_token_contract.address,
+    )
```
gabrielfior, line 786: This can be simplified by changing the parameter `amount: xDai` to `amount: Wei`, i.e. we expect the `bet amount` to be given in collateral Wei units of the market.
See `auto_deposit` discussion as well -> https://github.com/gnosis/prediction-market-agent-tooling/pull/645/files#r2000890295

gabrielfior: It's to be acknowledged that this would imply lots of changes in the codebase.

kongzii: I'm doing something similar, but the conversion needs to be placed somewhere in any case (either inside of this function or outside before calling it).

The problem is, if we accept a generic type like `Wei`, then it's a big difference if we are talking about GNOs in Wei or wxDai in Wei. And that's where a lot of complexity comes to avoid such mistakes. 

This PR is a simple baby step, we can wait for the follow up PR to see the full thing in place.

gabrielfior: In my head, I would approach this in this order:
-> Let's assume the market's collateral is GNO
-> Accept an amount in Wei
-> We check if the user has a balance > amount. If yes, then no conversion necessary.
-> If no, then we either abort of go to the `auto_deposit` logic.

> The problem is, if we accept a generic type like Wei, then it's a big difference if we are talking about GNOs in Wei or wxDai in Wei. And that's where a lot of complexity comes to avoid such mistakes.

I didn't think it would be so complex - if we have the collateral token address, we can simply call `decimals` on the contract and find out how to convert Wei into proper units. Or is it more complicated than this?

kongzii: Image this from agent's point of view (and ours as their deployers).

We configure, for example on prophet agent:

```
def get_betting_strategy(self, market: AgentMarket) -> BettingStrategy:
        return KellyBettingStrategy(
            max_bet_amount=get_maximum_possible_bet_amount(
                min_=1, max_=4.77, trading_balance=market.get_trade_balance(APIKeys())
            ),
            max_price_impact=0.63,
        )
```

What is min_ and max_ here? I think the easiest currency for everyone to think about is USD (as xDai is stable coin against USD).

So in the very beginning, agent decides how much to bet in USD and then we need USD -> CollateralToken conversion somewhere along the way.

However, we need to type it properly and mypy needs to yell at us as much as possible where we have USD, where GNO and where we have no idea what the underlying collateral is (ie a plain `Wei` type).

Because if we send 1 (we assume GNO) to a function that operates with USD, it's not a big deal.

But if we send 1 (and we assume it's 1 USD), to a function that operates in GNOs, we will buy ~130 USD worth of tokens instead of 1.

kongzii: It also adds complexity when we do analytics. 

If profit is calculated in Weis, then profit of 100 Wei on wxDai market is much smaller then profit of 100 Wei on GNO market. So we need to work in the same currency when we need it.

gabrielfior: On an agent level, agree it's a good idea to work in USD terms (i.e. xDAI terms).
On a low-level (e.g. `omen_buy_outcome_tx`) it seems to me the easiest thing is to work with collateral_tokens, the function has some logic for handling `outcome_tokens`, not converting token amounts.
So, to your point: agree a conversion needs to take place, but hopefully before this function, so that the function is able to stay on the collateral token "units", if that makes sense.

gabrielfior: > It also adds complexity when we do analytics.
If profit is calculated in Weis, then profit of 100 Wei on wxDai market is much smaller then profit of 100 Wei on GNO market. So we need to work in the same currency when we need it.

This can be solved by adding a field `collateral_token_address` to the profit calculations. This is indeed complex, because let's say
-> Market is denominated in GNO
-> On day 1, we bet 1 GNO (market price 200 USD)
-> During days 1-30, our open interest is 1 GNO (USD value varies due to GNO price varying)
-> On day 30, market is resolved and we earn 0.1 GNO (on day 30, GNO is quoted as 130 USD)

Hence I believe it's more stable to work on the collateral amounts, which remain fixed. What does not remain fixed is the USD value (that changes with token price).

kongzii: Yep why not, I'm doing something like:

```
buy_outcome(value: USD, ...): 
    converted_to_collateral = ...
    buy_outcome_raw(converted_to_collateral)
    
buy_outcome_raw(value: Wei, ...):
    ...
```

kongzii: > -> On day 1, we bet 1 GNO (market price 200 USD)

If the agent thinks in USD, then we bet X USD, and whatever it is in GNO at the time. 

Doesn't that simplify exchange rates? Ie we don't need to care about them?

gabrielfior: -> On day 1, we bet 1 GNO (let's say 150 USD).
-> On day 30 (when market is resolved), we received 0.1 GNO back (let's say 25 USD, because GNO on that day is 250 USD). 
-> Now, ROI is 0.1 / 1 = 10%, but due to the market quotes, if we calculate in USD terms, it can vary (depending on exchange rate and price at those dates). Hence it makes more sense in my mind to keep it in `collateral_token` units.

kongzii: I see what you mean now! Agree. I will try to not mix it up 😄

prediction_market_agent_tooling/gtypes.py
```diff
@@ -9,57 +10,134 @@
...
-Probability = NewType("Probability", float)
-Mana = NewType("Mana", float)  # Manifold's "currency"
-USDC = NewType("USDC", float)
-ChainID = NewType("ChainID", int)
-IPFSCIDVersion0 = NewType("IPFSCIDVersion0", str)
 
+class Token(_GenericValue[int | float | str | Decimal, float], parser=float):
+    """
+    Represents any token in its decimal form, it could be 1.1 GNO, WXDAI, XDAI, Mana, whatever. We don't know the currency, just that it's in the decimal form.
+    """
```
gabrielfior, line 33: `Token` to me means any ERC-20 compatible token, i.e. supports `balanceOf`, `decimals`, among [others](https://docs.openzeppelin.com/contracts/3.x/erc20). 
If I understand correctly, you want to differentiate between OutcomeToken and CollateralToken and have ways to convert between their values. In that sense, I would rename `Token` to `ColleteralToken`. Alternatively, I would have CollateralToken inheriting from Token, and OutcomeToken inheriting from Token.

prediction_market_agent_tooling/tools/_generic_value.py
```diff
@@ -0,0 +1,261 @@
...
+    "InputValueType", bound=t.Union[str, int, float, WeiWeb3, Decimal]
+)
+InternalValueType = TypeVar("InternalValueType", bound=t.Union[int, float, WeiWeb3])
+
+
+class _GenericValue(
+    t.Generic[InputValueType, InternalValueType],
+    # Not great, but it allows to serialize this object with plain json.
+    dict[t.Literal["value"] | t.Literal["type"], InternalValueType | str],
+):
```
gabrielfior, line 19: Overall I think this approach works, although complex given all the magic for mypy to identify cases e.g. `b - c # mypy will report incompatible types`.

My alternative approach for discussion would be something like
```
class SafeTokenValue:
    def __init__(self, symbol: str, decimals: int, value: str):
        self.symbol = symbol
        self.decimals = decimals
    
    def __add__(self, other: SafeToken):
        if self.symbol != other.symbol:
            raise ValueError("symbol mismatch")
        return self.value + other.value
    
    # def __sub__(self, other: SafeToken): # similar to above
    # def __mul__(self, other: SafeToken): # similar to above

# One could also subclass the above and fix symbol and decimals, e.g. GNO, xDAI, etc```

The benefits of the above are
- less complex
- avoids cases like `xDAI(10) - GNO(2)`

unclear
- mypy "friendliness"

File: examples/monitor/match_bets_with_langfuse_traces.py [hunk 5]
```diff
@@ -96,34 +96,36 @@ def get_outcome_for_trace(
...
+    market_before_placing_bet = get_omen_market_by_market_id_cached(
+        HexAddress(HexStr(market.id)), block_number=bet_tx_block_number - 1
+    )
+    omen_agent_market_before_placing_bet = OmenAgentMarket.from_data_model(
+        market_before_placing_bet
+    )
+
+    but_trade_in_tokes = omen_agent_market_before_placing_bet.get_in_token(
+        buy_trade.amount
+    )
```
gabrielfior, line 116:
```suggestion
    bet_trade_in_tokes = omen_agent_market_before_placing_bet.get_in_token(
        buy_trade.amount
    )
```

examples/monitor/match_bets_with_langfuse_traces.py
```diff
@@ -96,34 +96,36 @@ def get_outcome_for_trace(
...
     if not correct:
-        profit = -buy_trade.amount.amount
+        profit = -but_trade_in_tokes
```
gabrielfior, line 118:
```suggestion
        profit = -bet_trade_in_tokes
```

examples/monitor/match_bets_with_langfuse_traces.py
```diff
@@ -96,34 +96,36 @@ def get_outcome_for_trace(
...
-        )
-
         received_outcome_tokens = (
             omen_agent_market_before_placing_bet.get_buy_token_amount(
                 buy_trade.amount, direction=buy_trade.outcome
-            ).amount
+            )
         )
-        profit = received_outcome_tokens - buy_trade.amount.amount
+        profit = received_outcome_tokens.as_token - but_trade_in_tokes
```
gabrielfior, line 125:
```suggestion
        profit = received_outcome_tokens.as_token - bet_trade_in_tokes
```

examples/monitor/match_bets_with_langfuse_traces.py
```diff
@@ -96,34 +96,36 @@ def get_outcome_for_trace(
...
 
     return SimulatedOutcome(
-        size=buy_trade.amount.amount,
+        size=but_trade_in_tokes,
```
gabrielfior, line 128:
```suggestion
        size=bet_trade_in_tokes,
```

prediction_market_agent_tooling/gtypes.py
```diff
@@ -9,57 +10,134 @@
...
+
+    @property
+    def as_wei(self) -> Wei:
+        """
+        OutcomeWei is essentialy Wei as well, when you know you need to pass it, you can convert it using this.
+        """
+        return Wei(self.value)
+
+
+class xDaiWei(_GenericValue[Web3Wei | int | str, Web3Wei], parser=int):
```
gabrielfior, line 120: [This also applies to xDaiWei and etc]

I don't love this. Wei is simply a [unit](https://eth-converter.com/).
I can follow that one of the motivations is for `mypy` to be happy and make the overall project type-safe, but this just adds unnecessary complexity to the code in my opinion.

In essence, in my understanding the goal is to have a way to represent a variable so that its units are known. We can do that either using mypy type safety checks or at runtime (e.g. having `assert units are the same` or similar). 
I would go with the runtime option in this case (see https://github.com/gnosis/prediction-market-agent-tooling/pull/645/files#r2008761886) because it's more flexible and makes "wei" a unit (while also catching inconsistencies) and allows new token collaterals (which will def come as we support more markets) to be used out-of-the-box, without adding new classes.

Do you think that could work?

prediction_market_agent_tooling/gtypes.py
kongzii: Sure, CollateralToken is fine name.

Inheriting isn't as nice as it sounds at first (I tried). Then if some function accepts `Token`, is easy to make mistake of passing `OutcomeToken` instead of `CollateralToken` or vice-versa.

prediction_market_agent_tooling/tools/_generic_value.py
kongzii: This is less complex because it has less functionality. 

It strips away `t.Generic[InputValueType, InternalValueType]`, but thanks to that:

```
x = Wei(5.4) # Mypy would complain here as it doesn't make sense afaik
```

image

Or in this case:

```
x = Wei(10)
reveal_type(x.value)
```

Mypy knows that `value` property is of type `Web3Wei` (I mean `from web3.types import Wei as Web3Wei` here), so we can just pass it to the functions that work with web3's Wei type.

Also it removes the `dict[t.Literal["value"] | t.Literal["type"], InternalValueType | str],`, but thanks to that, these `_GenericValue` can be serialized using plain `json.dumps(...)`.

However, having that symbol to avoid cases like``xDAI(10) - GNO(2`` at least on runtime is a good idea. But I'd add that in follow up PR.

examples/monitor/match_bets_with_langfuse_traces.py
kongzii: Oops thanks

examples/monitor/match_bets_with_langfuse_traces.py
kongzii: `buy` is correct as it was before, fixed to that

prediction_market_agent_tooling/gtypes.py
kongzii: Ideally, these `gtypes` (USD, Wei, xDai, ...) are developed in a way that you don't know about the complexity behind it at all. We should be able just to use them; they shouldn't bother you or make your life harder with PMAT. Quite the opposite, actually!

So, if you ever run into a situation where using these `_GenericValue` subclasses aren't working as expected, ping me, and I'll try to do it better.

Having said that, no, I don't think runtime is good enough; we need both. Look at how many places I had to correct the type (or the whole logic due to an uncaught bug). Don't you like if you refactor something and mypy informs you about it without trying to run all the agents?

kongzii:
As agreed over Slack, I will merge this, and let's see how it goes in production 😄 The Last test failed due to flaky CoW calls. We need some solution for that. But that touches more than this PR only.

PR #650, opened 2025-03-25
Add support for Safe in swap_tokens contract

kongzii:
deploy please


prediction_market_agent_tooling/tools/cow/cow_order.py
```diff
@@ -137,3 +153,27 @@ async def swap_tokens_waiting_async(
...
+    posted_order = await order_book_api.get_order_by_uid(order.uid)
+    CowGPv2SettlementContract(
+        address=Web3.to_checksum_address(
+            check_not_none(posted_order.settlementContract).root
+        )
+    ).setPreSignature(
+        api_keys,
+        HexBytes(posted_order.uid.root),
+        True,
+    )
```
gabrielfior, line 179: This logic should be implemented in the `cow-py` package (fine to have it here for now). But please create an issue there for us to track.

Another thing - I'm intrigued by the fact that the code [here](https://github.com/cowdao-grants/cow-py/blob/ed0ce92e3d4fce76f95f2fddf1e33bfe53423f5e/cowdao_cowpy/contracts/sign.py#L75C68-L75C77) is not failing, since `owner.key` for the Safe is not defined.
```
# [cow-py](https://github.com/cowdao-grants/cow-py/tree/ed0ce92e3d4fce76f95f2fddf1e33bfe53423f5e)/[cowdao_cowpy](https://github.com/cowdao-grants/cow-py/tree/ed0ce92e3d4fce76f95f2fddf1e33bfe53423f5e/cowdao_cowpy)/[contracts](https://github.com/cowdao-grants/cow-py/tree/ed0ce92e3d4fce76f95f2fddf1e33bfe53423f5e/cowdao_cowpy/contracts)
/sign.py
def ecdsa_sign_typed_data(
    owner: LocalAccount,
    domain_data: TypedDataDomain,
    message_types: Dict[str, Any],
    message_data: Dict[str, Any],
) -> SignedMessage:
    return Account._sign_hash(
        hash_typed_data(domain_data, message_types, message_data), owner.key
    )
```

What I'm trying to say is, `cowpy` signs the tx by default (if using an EOA), they should do the same for a Safe.

gabrielfior: Given the changes here (https://github.com/kongzii/cow-py/commit/ea664ad4c38012220347496fdfd77d8c956f527e), can we somehow move the logic to the cowpy repo directly?

kongzii: I had some discussions with the cow team before (although a different use-case, but similar), and they didn't want to have Safe dependency in their library. The reason was that CoW TS SDK also doesn't have it.

Please check the TG group discussions (it should be among last messages there) and if you think it's worth asking them, I don't mind moving this to CoW library. (actually I wanted to do it that way initially, but then I remembered the discussions)

gabrielfior (approved):

PR #654, opened 2025-04-01
Gabriel/hotfix seer

gabrielfior:
Small fixes for Seer compatibility


prediction_market_agent_tooling/markets/seer/price_manager.py
```diff
@@ -86,8 +88,8 @@ def get_price_for_token(
             )
             return self.get_token_price_from_pools(token=token)
 
-        price = collateral_exchange_amount / float(quote.quote.buyAmount.root)
-        return price
+        price = collateral_exchange_amount.as_wei / Wei(quote.quote.buyAmount.root)
```
kongzii, line 91: If this is needed, then you could use `get_buy_token_amount_else_raise` function instead of `get_quote` (it's the same thing, just already wrapped in Wei).

And if that's it's the case, maybe there is no need to have `get_quote` anymore? 🤔 So we would have only `get_buy_token_amount_else_raise` that returns typed Wei?

image

kongzii (approved):

PR #660, opened 2025-04-08
Error out on invalid outcome tokens

kongzii:


prediction_market_agent_tooling/markets/omen/omen.py
```diff
@@ -336,6 +339,11 @@ def from_created_market(model: "CreatedMarket") -> "OmenAgentMarket":
 
     @staticmethod
     def from_data_model(model: OmenMarket) -> "OmenAgentMarket":
+        if not all(is_valid_wei(number.value) for number in model.outcomeTokenAmounts):
+            raise ValueError(
```
gabrielfior, line 343: I thought we should rather skip this one market, instead of halting the processing of all markets - wdyt?

gabrielfior: This occurs very rarely, so tracking this as error for easier traceability

PR #662, opened 2025-04-08
Gabriel/redeem seer

gabrielfior:


prediction_market_agent_tooling/markets/seer/seer_contracts.py
```diff
@@ -79,3 +80,33 @@ def create_categorical_market(
...
+        params_dict = params.model_dump(by_alias=True)
+        # We explicity set amounts since OutcomeWei gets serialized as dict.
+        params_dict["amounts"] = [amount.value for amount in params.amounts]
+        receipt_tx = self.send(
+            api_keys=api_keys,
+            function_name="redeemToBase",
+            function_params=params_dict,
+            web3=web3,
+        )
+        return receipt_tx
```
gabrielfior, line 112: Explicity converting `OutcomeWei` into integers, otherwise gets `model_dump` as {'value: 100000, type: 'int'} or similar, whereas we simply want the value.

prediction_market_agent_tooling/markets/seer/data_models.py
```diff
@@ -186,3 +187,13 @@ def created_time(self) -> DatetimeUTC:
...
+
+
+class RedeemParams(BaseModel):
+    model_config = ConfigDict(populate_by_name=True)
+    market: ChecksumAddress
+    outcome_indices: list[int] = Field(alias="outcomeIndexes")
+    amounts: list[OutcomeWei]
+
+
+ADDRESS_ZERO
```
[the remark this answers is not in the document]
kongzii: Good rabbit!

prediction_market_agent_tooling/markets/seer/seer.py
```diff
@@ -203,9 +206,87 @@ def get_user_id(api_keys: APIKeys) -> str:
...
     @staticmethod
-    def redeem_winnings(api_keys: APIKeys) -> None:
-        # ToDo - implement me (https://github.com/gnosis/prediction-market-agent-tooling/issues/499)
-        pass
+    def _is_redeemable(market: SeerMarket, balances: list[OutcomeWei]) -> bool:
+        if not market.payout_reported:
+            return False
+        return any(
+            payout and balance > 0
+            for payout, balance in zip(market.payout_numerators, balances)
```
kongzii, line 214: Couldn't this `_is_redeemable` be instance method and accept only argument like `def market_redeemable_by(self, user: ChecksumAddress) -> bool:`? (as in `OmenAgentMarket`)

Because assuming that the market's payout_numerators and balances passed into this function are sorted correctly is a kind of big assumption that can be broken without notice.

prediction_market_agent_tooling/markets/seer/seer.py
```diff
@@ -203,9 +206,87 @@ def get_user_id(api_keys: APIKeys) -> str:
...
+        )
+
+    @staticmethod
+    def _get_outcome_token_balances(
+        market: SeerMarket, address: ChecksumAddress, web3: Web3 | None = None
```
kongzii, line 219: If this function requires `market`, why not just have it as a normal instance method that can use `self`? `SeerAgentMarket` already has `wrapped_tokens` property, but you can add any required fields to `SeerAgentMarket` as needed (to fix this pattern in other places as well)

prediction_market_agent_tooling/markets/seer/seer.py
```diff
@@ -203,9 +206,87 @@ def get_user_id(api_keys: APIKeys) -> str:
...
+        gnosis_router = GnosisRouter()
+        for market in markets_to_redeem:
+            params = RedeemParams(
+                market=Web3.to_checksum_address(market.id),
+                outcome_indices=list(range(len(market.payout_numerators))),
+                amounts=market_balances[market.id],
+            )
+            gnosis_router.redeem_to_base(api_keys, params=params, web3=web3)
+
+        # Auto-withdraw not implemented to reduce complexity (conversion already occurs before placing bets)
```
kongzii, line 289: "For reducing complexity, we will let agents bug out every few days and do it manually" 😄 

All that is needed is something like the:

```
if auto_withdraw:
  auto_withdraw_collateral_token(
    collateral_token_contract=self.collateral_token_contract_address_checksummed,
    amount_wei=amount_to_withdraw_in_collateral_token,
    api_keys=api_keys,
    web3=web3,
  )
```

Where `amount_to_withdraw_in_collateral_token` can be obtained by calling `ContractERC20BaseClass(address=self.collateral_token_contract_address_checksummed).balanceOf(api_keys.bet_from_address)` before and after all the redeemings and just using the difference (if it isn't returned by the `redeem_to_base` method itself).

That difference would be also nice to have logged anyway so we can see it works in the gcp logs.

It isn't so bad, is it? I hope it's safe to assume we have stuff in the code because it was needed for some good reason, then it's a little sad to me to see if it isn't used because it's "too complex", when it's just a single function call and rather let the original problem occur again.

prediction_market_agent_tooling/tools/cow/cow_order.py
```diff
@@ -221,3 +227,15 @@ async def swap_tokens_waiting_async(
         )
 
         await asyncio.sleep(3.14)
+
+
+def get_trades_by_owner(
```
kongzii, line 232: Maybe some tenacity retry here to get rid of potential crashes?

prediction_market_agent_tooling/markets/seer/seer.py
```diff
@@ -203,9 +206,87 @@ def get_user_id(api_keys: APIKeys) -> str:
...
+                filtered_markets.append(market)
+
+        return filtered_markets
+
+    @staticmethod
+    def redeem_winnings(api_keys: APIKeys, auto_withdraw: bool = True) -> None:
+        web3 = RPCConfig().get_web3()
+        subgraph = SeerSubgraphHandler()
+
+        closed_markets = subgraph.get_binary_markets(
```
kongzii, line 260: What I'm afraid here is that after some time, even if Seer doesn't have many markets (yet), this `redeem_winnings` method will take forever or error out on timeouts. Because it always queries all the closed markets and all the trades on CoW (even non-Seer-related).

Unfortunately, I tried to quickly look at Seer's subgraph and CoW API, and I'm afraid there is no good solution atm, but at least to keep it in our minds. Unless you'd know how to do it differently right away 🤔

gabrielfior: Another filter I can think of is something like `market closed N days ago`, and pass that to the subgraph handler.
The property `openingTs` can be used
https://github.com/seer-pm/demo/blob/9517af2f2969eff4d8e6b3e7bc83e969fa3e59ba/web/src/components/Market/MarketsFilterBox.tsx#L27

prediction_market_agent_tooling/markets/seer/seer.py
gabrielfior: It can't be an `instance` method because it's called from a `static` method.

> Because assuming that the market's payout_numerators and balances passed into this function are sorted correctly is a kind of big assumption that can be broken without notice.

Indeed that can cause problems, but the data structure comes from the subgraph and multiple things would break if the ordering were changed. So I think the current implementation is fine, or do you have any immediate suggestions?

prediction_market_agent_tooling/markets/seer/seer.py
gabrielfior: It's called from `static` method, defined here https://github.com/gnosis/prediction-market-agent-tooling/blob/8d10fd339020f56df2600a3366448e49d1e96ecf/prediction_market_agent_tooling/markets/agent_market.py#L260

prediction_market_agent_tooling/markets/seer/seer.py
gabrielfior: Agree to use the `auto_withdraw` method, but note the following
-> We want to convert wxDAI (sent by contract) into sDAI (fixed collateral on Seer)
-> We don't have `self.collateral_token_contract_address_checksummed` due to `static` on `redeem_winnings`
-> My understanding of `auto_withdraw_collateral_token` is that it withdraws X -> wxDAI (`KEEPING_ERC20_TOKEN`). That's already handled by the contract.
-> If e.g. the user has 10 wxDAI but no sDAI (required for placing bets), `auto_deposit` handles this already.

prediction_market_agent_tooling/markets/seer/seer.py
kongzii: yeah if it becomes a problem sounds good, I guess for now not needed

prediction_market_agent_tooling/markets/seer/seer.py
kongzii: It's used like this:

```python
markets_to_redeem = [
            market
            for market in filtered_markets
            if SeerAgentMarket._is_redeemable(market, market_balances[market.id])
        ]
```

`filtered_markets` there seem to be instances of markets, any reason why 

```python
markets_to_redeem = [
            market
            for market in filtered_markets
            if market.is_reedemable(...)])
        ]
```

shouldn't work?

(in Python it doesn't matter that `redeem_winnings` is staticmethod. `filtered_market` is an instance so `_is_redeemable` should be implementable like in OmenAgentMarket and be normal method)

prediction_market_agent_tooling/markets/seer/seer.py
kongzii: This is similar to the `_is_redeemable` before, it doesn't matter that `redeem_winnings` is static method, because you work with instances of markets in that staticmethod

prediction_market_agent_tooling/markets/seer/seer.py
kongzii: Agreed that auto_withdraw is not necessary, because, for some reason, Seer decided to do withdrawing of `sDai -> wxDai -> xDai` automatically. But to go over the points why it is necessary to have that functionality:

- imagine that agent is betting for a while, sooner or later, it will have all its funds in sDai, and he won't be able to pay for transactions fees, so it won't matter that he is rich in sDai, he won't be able to operate (if he had funds in `KEEPING_ERC20_TOKEN`, then they are automatically withdrawn to `xDai` so he can keep paying for fees, when `xDai` balance goes low)
- we agreed before that `KEEPING_ERC20_TOKEN` is token where agent is holding funds, the [get_trade_balance](https://github.com/gnosis/prediction-market-agent-tooling/blob/55ed76ea4a51b355114187f0e097c12bef554d97/prediction_market_agent_tooling/markets/seer/seer.py#L152) function assumes that and if he is rich in sDai, but poor in `KEEPING_ERC20_TOKEN`, agent will stop processing because it will thinks he doesn't have funds

(`redeem_winnings` is static, but it's working with instances of markets that do have `collateral_token_contract_address_checksummed`)

prediction_market_agent_tooling/markets/seer/seer.py
```diff
@@ -202,10 +205,88 @@ def get_position(
...
+        gnosis_router = GnosisRouter()
+        for market in markets_to_redeem:
+            params = RedeemParams(
+                market=Web3.to_checksum_address(market.id),
+                outcome_indices=list(range(len(market.payout_numerators))),
+                amounts=market_balances[market.id],
+            )
+            gnosis_router.redeem_to_base(api_keys, params=params, web3=web3)
+
+        # GnosisRouter sends wxDAI to user, so no auto-withdraw needed.
```
kongzii, line 289:
```suggestion
        # GnosisRouter withdraws sDai into wxDAI/xDai on its own, so no auto-withdraw needed by us.
```

PR #679, opened 2025-05-05
New try for logs

kongzii:
Let's see how it goes with this, in my tests, logs in GCP look correct:

All properly formatted:

image

And exception also grouped together:

image

### DIFF
--- a/prediction_market_agent_tooling/loggers.py
+++ b/prediction_market_agent_tooling/loggers.py
@@ -4,6 +4,7 @@ import sys
 import typing as t
 import warnings
 from enum import Enum
+from pythonjsonlogger import jsonlogger
 
 from loguru import logger
 from pydantic_settings import BaseSettings, SettingsConfigDict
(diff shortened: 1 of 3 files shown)


prediction_market_agent_tooling/loggers.py
```diff
@@ -31,17 +32,35 @@ class LogConfig(BaseSettings):
     LOG_LEVEL: LogLevel = LogLevel.DEBUG
 
 
-class NoNewLineStreamHandler(logging.StreamHandler):  # type: ignore # StreamHandler is not typed in the standard library.
-    def format(self, record: logging.LogRecord) -> str:
-        return super().format(record).replace("\n", " ")
-
+class _CustomJsonFormatter(jsonlogger.JsonFormatter):
```
kongzii, line 35: The main change: In GCP, format logs as JSON (in localhost, it formats them normally using loguru as before)

prediction_market_agent_tooling/loggers.py
```diff
@@ -31,17 +32,35 @@ class LogConfig(BaseSettings):
...
+        self,
+        log_record: dict[str, t.Any],
+        record: logging.LogRecord,
+        message_dict: dict[str, t.Any],
+    ) -> None:
+        super().add_fields(log_record, record, message_dict)
+        # Include "level" and "severity" with the same value as "levelname" to be friendly with log aggregators.
+        if log_record.get("levelname"):
+            log_record["level"] = log_record["levelname"]
+            log_record["severity"] = log_record["levelname"]
```
kongzii, line 46: Extend the JSON structure by these two, so GCP understands the level.

prediction_market_agent_tooling/loggers.py
```diff
@@ -31,17 +32,35 @@ class LogConfig(BaseSettings):
...
+
+    @staticmethod
+    def get_handler() -> logging.StreamHandler:  # type: ignore # Seems correct, but mypy doesn't like it.
+        logHandler = logging.StreamHandler()
+        formatter = _CustomJsonFormatter("%(asctime)s %(levelname)s %(message)s")
+        logHandler.setFormatter(formatter)
+        return logHandler
+
+
+def _handle_exception(
```
kongzii, line 56: Second main change: Thanks to this function, if code raises an exception (e.g. `raise ValueError`), it will be logged using the JSON structured log as well.

gabrielfior: What about `logger.info` and `logger.debug`, are they also logged now?

kongzii: `info` was always logged and `debug` is ignored on purpose, but that's configured in infra repository by LOG_LEVEL

gabrielfior: I would vote for debug to NOT be ignored, we can control the log level on GCP directly, no? Why ignoring?

kongzii: Debug is usually ignored in production because it's expected to have a ton of debug logs, which aren't necessary to track in production. The debug logs are meant for debugging. 

image

PR #684, opened 2025-05-06
Group crc pms

gabrielfior:
Allows for Metri super group ERC20 tokens to be used as collateral when creating markets

### DIFF
--- a/prediction_market_agent_tooling/markets/omen/omen_contracts.py
+++ b/prediction_market_agent_tooling/markets/omen/omen_contracts.py
@@ -431,6 +431,12 @@ class OmenFixedProductMarketMakerContract(ContractOnGnosisChain):
         )
 
 
+class MetriSuperGroup(ContractERC20OnGnosisChain):
+    address: ChecksumAddress = Web3.to_checksum_address(
+        "0x7147A7405fCFe5CFa30c6d5363f9f357a317d082"
+    )
+
+
 class GNOContract(ContractERC20OnGnosisChain):
     address: ChecksumAddress = Web3.to_checksum_address(
         "0x9c58bacc331c9aa871afd802db6379a98e80cedb"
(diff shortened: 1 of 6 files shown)


prediction_market_agent_tooling/tools/contract.py
```diff
@@ -583,6 +583,16 @@ def inc(
         )
 
 
+def contract_exposes_balanceOf(web3: Web3, address: ChecksumAddress) -> bool:
+    """This is a fallback check for contracts that implement some proxies but are indeed ERC20-compatible."""
```
gabrielfior, line 587: This was needed for the ERC20 tokens created for Circles (via hub.wrap), e.g. https://gnosisscan.io/address/0x7147a7405fcfe5cfa30c6d5363f9f357a317d082

prediction_market_agent_tooling/markets/omen/omen_contracts.py
```diff
@@ -431,6 +431,12 @@ def get_collateral_token_contract(
         )
 
 
+class MetriSuperGroup(ContractERC20OnGnosisChain):
+    address: ChecksumAddress = Web3.to_checksum_address(
+        "0x7147A7405fCFe5CFa30c6d5363f9f357a317d082"
```
kongzii, line 436: Will just bring the question from issue here:

That token is called Test, is that really the one to be used for production PMs?

prediction_market_agent_tooling/tools/contract.py
kongzii: `contract_implements_function` is already considering proxies, [see here](https://github.com/gnosis/prediction-market-agent-tooling/blob/dd7986afb99d823910a91e664bd4e9b4acbef96b/prediction_market_agent_tooling/tools/contract.py#L604).

Looks like the given contract doesn't use `implementation` field, but `singleton` field for the proxyfied contract.

So you could modify `contract_implements_function` to consider both `implementation`  and `singleton` fields.

Using this try-catch method isn't great, because it doesn't raise any specific error that could be watched, just the general reverted, and `except Excetion` will catch many other problems as well. And then it takes a long time to use this function ,as it tries to retry it:

image

prediction_market_agent_tooling/markets/omen/omen_contracts.py
gabrielfior: I double-checked with Circles team, this token is correct and represents the correct group.

prediction_market_agent_tooling/tools/contract.py
gabrielfior: I refactored the implementation and now also checks for the case where`singleton` is used. Note that `singleton` is private and the delegation occurs via `delegatecall`, so not simply checking for a method (see https://blog.openzeppelin.com/deep-dive-into-the-minimal-proxy-contract for more details).

kongzii (approved):

PR #688, opened 2025-05-12
Created logprobs parser for OAI models

TS9001:
Linked to https://github.com/gnosis/prediction-market-agent/issues/712

### DIFF
--- a/prediction_market_agent_tooling/logprobs_parser.py
+++ b/prediction_market_agent_tooling/logprobs_parser.py
@@ -1,0 +1,98 @@
+import math
+from typing import Any, Tuple, TypedDict
+from prediction_market_agent_tooling.loggers import logger
+from itertools import product
+
+class LogprobKey(TypedDict):
+    name: str
+    key_type: type
+    valid_values: set[Any] | None
+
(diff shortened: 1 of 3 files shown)


prediction_market_agent_tooling/logprobs_parser.py
```diff
@@ -0,0 +1,115 @@
+import math
+from itertools import product
+from typing import Any, Tuple, TypedDict
+
+from prediction_market_agent_tooling.loggers import logger
+
+
+class LogprobKey(TypedDict):
```
kongzii, line 8: If there isn't reason to use TypedDict, then we use Pydantic' BaseModel everywhere around PMAT/PMA 🙏 (I know Prophet had some TypedDicts, but that's a different company/team 😄 )

kongzii (approved):
Small nit around typeddict, otherwise lgtm. Too much logic, but tests seem to work 😄

prediction_market_agent_tooling/logprobs_parser.py
TS9001: Fixed

PR #691, opened 2025-05-14
Changed logprobs parser to use pydantic models

TS9001:
Changes to use pydantic models and validation instead of custom dictionary models

### DIFF
--- a/prediction_market_agent_tooling/logprobs_parser.py
+++ b/prediction_market_agent_tooling/logprobs_parser.py
@@ -55,7 +55,7 @@ class LogprobsParser:
                 for i in range(result_start_index, len(logprobs))
                 if logprobs[i]["token"] in {",", '"', ",\n", "\",\n'", '",\n'}
             ),
-            -1,
+            len(logprobs) - 1,
         )
         return result_start_index + 1, result_end_index
 
@@ -70,7 +70,10 @@ class LogprobsParser:
             return False
 
     def _parse_valid_tokens_with__agg_probs(
-        self, logprobs_list: list[tuple[dict[str, Any]]], field_info: FieldInfo
+        self,
+        logprobs_list: list[tuple[dict[str, Any]]],
+        field_info: FieldInfo,
+        top_logprobs: int,
     ) -> list[dict[str, Any]]:
         results: list[dict[str, Any]] = [
             {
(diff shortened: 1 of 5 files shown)


tests/test_logprobs_parser.py
```diff
@@ -1,15 +1,22 @@
 import math
-from typing import Any, Dict, List, TypedDict
+from typing import Any, Dict, List, Literal
 
 import pytest
+from pydantic import BaseModel
 
-from prediction_market_agent_tooling.logprobs_parser import LogprobKey, LogprobsParser
+from prediction_market_agent_tooling.logprobs_parser import LogprobsParser
```
gabrielfior, line 9: The unit tests make a lot of sense and I'm happy with those.
Could I ask one tiny integration test where e.g. OpenAI is called and the logprobs are retrieved? No strong preference, but just having a way to be alerted in case pydantic, openai or someone else changes how logprobs are delivered.

TS9001: Hmm this means that each time the test is executed we have to pay for tokens. Are we doing something similar somewhere else?

prediction_market_agent_tooling/logprobs_parser.py
```diff
@@ -1,29 +1,39 @@
...
+
+class FieldLogprobs(BaseModel):
+    key: str
+    logprobs: list[LogprobDetail]
 
 
 class LogprobsParser:
+    def __init__(self, skip_fields: list[str] = []):
+        self.skip_fields = ["logprobs"] + skip_fields
+
```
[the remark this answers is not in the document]
kongzii: Not a big deal here, but generally good advice to follow

tests/test_logprobs_parser.py
kongzii: We do, but they are disabled by default, behind a environment flag: https://github.com/gnosis/prediction-market-agent/blob/7cee450eb56ef48394145c5f03fb485c858ea671/tests/agents/test_goal_manager.py#L113

TS9001: Added integration test - it needs to be done with native OAI client so I wrote TODO there, so we can remember to switch for PydanticAI when it is done

prediction_market_agent_tooling/logprobs_parser.py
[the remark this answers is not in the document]
TS9001: Thank rabitai sensei, completely forgot about this-one :D

TS9001:
Rebased and FP

prediction_market_agent_tooling/logprobs_parser.py
```diff
@@ -73,45 +86,51 @@ def _parse_valid_tokens_with__agg_probs(
...
-            valid_logprobs = [
+            valid_logprobs_raw = [
                 logprobs[i]["top_logprobs"]
                 for i in range(result_start_index, result_end_index)
             ]
 
+            parsed_logprobs_data = self._parse_valid_tokens_with__agg_probs(
+                list(product(*valid_logprobs_raw)), field_info
+            )
+
```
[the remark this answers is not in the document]
TS9001: This is actually good idea.. in case there would anything happen during tokens parsing it could be rest of the text so it would definitely explode

kongzii (approved):

PR #697, opened 2025-05-21
Add auth to the RPC

kongzii:


prediction_market_agent_tooling/config.py
```diff
@@ -307,8 +310,30 @@ def chain_id_to_rpc_url(self, chain_id: ChainID) -> URI:
         else:
             raise ValueError(f"Unsupported chain ID: {chain_id}")
 
+    def chain_id_to_rpc_bearer(self, chain_id: ChainID) -> SecretStr | None:
+        if chain_id == ETHEREUM_ID:
+            return self.ETHEREUM_RPC_BEARER
+        elif chain_id == GNOSIS_CHAIN_ID:
+            return self.GNOSIS_RPC_BEARER
+        else:
+            raise ValueError(f"Unsupported chain ID: {chain_id}")
```
gabrielfior, line 319: I think a dict is more redable than `if-elif-elif...`

prediction_market_agent_tooling/config.py
```diff
@@ -307,8 +310,30 @@ def chain_id_to_rpc_url(self, chain_id: ChainID) -> URI:
...
+
     def get_web3(self) -> Web3:
-        return Web3(Web3.HTTPProvider(self.gnosis_rpc_url))
+        headers = {
+            "Content-Type": "application/json",
+            "User-Agent": construct_user_agent(str(type(self))),
```
gabrielfior, line 324: Why is this needed now?

kongzii: This is used automatically in `HTTPProvider` if you don't provide `headers` or `request_kwargs` at all.

So just copy-pasted it to keep the same behaviour as before.

prediction_market_agent_tooling/config.py
kongzii: Sure why not

gabrielfior (approved):

PR #701, opened 2025-05-21
Gabriel/hotfix seer liquidity

gabrielfior:
-> Added `outcomes_supply_gt` as filter (by default, > 0), similar to Seer
-> using db_cache as part of price discovery


prediction_market_agent_tooling/markets/seer/seer_subgraph_handler.py
```diff
@@ -88,6 +89,7 @@ def _build_where_statements(
             case FilterBy.OPEN:
                 and_stms["openingTs_gt"] = now
                 and_stms["hasAnswers"] = False
+                and_stms["outcomesSupply_gt"] = outcome_supply_lt_if_open.value
```
kongzii, line 92: `outcome_supply_lt_if_open` is called `lt`, but the filter used is `gt`

prediction_market_agent_tooling/markets/seer/seer_subgraph_handler.py
```diff
@@ -163,6 +166,7 @@ def get_markets(
         # Binary markets on Seer contain 3 outcomes: OutcomeA, outcomeB and an Invalid option.
         where_stms = self._build_where_statements(
             filter_by=filter_by,
+            outcome_supply_lt_if_open=outcome_supply_lt_if_open or Wei(0),
```
kongzii, line 169: Why not just use it as adefault parameter right away?

Ie `outcome_supply_lt_if_open: Wei | None = Wei(0),`

gabrielfior: implemented

prediction_market_agent_tooling/markets/seer/seer_subgraph_handler.py
gabrielfior: renamed

kongzii (approved):

PR #705, opened 2025-05-22
Fixes for Seer trading

gabrielfior:
Coinflip-Seer-high-liquidity was trying to bet on markets with liquidity 0 (e.g. https://app.seer.pm/markets/100/will-emmanuel-macron-resign-before-the-end-of-his-mandate).
Now we explicitly check if market has enough liquidity before proceeding.


prediction_market_agent_tooling/markets/seer/seer.py
```diff
@@ -277,33 +280,45 @@ def verify_operational_balance(api_keys: APIKeys) -> bool:
...
-        p = PriceManager(seer_market=model, seer_subgraph=seer_subgraph)
+        price_manager = PriceManager(seer_market=model, seer_subgraph=seer_subgraph)
 
-        probability_map = p.build_probability_map()
-        if not probability_map:
-            logger.info(
-                f"probability_map for market {model.id.hex()} could not be calculated. Skipping."
+        try:
+            probability_map = price_manager.build_probability_map()
+
```
gabrielfior, line 287: This can already fail (if market has no liquidity), and this gets caught on `except PriceCalculationError`

prediction_market_agent_tooling/markets/seer/seer.py
```diff
@@ -277,33 +280,45 @@ def verify_operational_balance(api_keys: APIKeys) -> bool:
...
+                fees=MarketFees.get_zero_fees(),
+                outcome_token_pool=None,
+                outcomes_supply=model.outcomes_supply,
+                resolution=None,
+                volume=None,
+                probabilities=probability_map,
+            )
+            if not market.has_liquidity():
+                logger.info(f"market {market.id} has not enough liquidity. Skipping.")
+                return None
```
gabrielfior, line 310: This is an additional check to make 100% sure the market has enough liquidity (for all outcomes except Invalid). It can be redundant, but it's now a bit more fail-safe.

prediction_market_agent_tooling/markets/seer/seer.py
```diff
@@ -277,33 +280,45 @@ def verify_operational_balance(api_keys: APIKeys) -> bool:
...
+            return market
+        except PriceCalculationError as e:
+            logger.warning(
+                f"Skipping market {model.id.hex()} due to price calculation error: {e}"
+            )
+            return None
+        except Exception as e:
+            logger.warning(
+                f"probability_map for market {model.id.hex()} could not be calculated. Exception {e}. Skipping."
```
kongzii, line 319: Catching all exceptions and logging them just as a warning isn't usually a good idea. Something will go wrong, and we will never be alerted about it.

`probability_map` raises `PriceCalculationError`, which is caught above, why is this needed?

prediction_market_agent_tooling/markets/seer/seer.py
kongzii: If the market sincerely doesn't have liquidity (not because of CoW errors, but really doesn't), that's not a problem. We often need to work with such markets (cleaner, ofv challenger, arbitrage agent come to mind).

This check makes sense somewhere, where `FilterBy.OPEN` is used, but not on general model construction.

prediction_market_agent_tooling/markets/seer/seer.py
gabrielfior: This is a catch-all - agree that logging and continuing is not a great pattern, but note that we don't have proper exception handling for each market, i.e. if one market fails, all processing halts - not optimal, in my opinion.
But anyways, happy to remove this Exception here, such exceptional cases should halt things.

prediction_market_agent_tooling/markets/seer/seer.py
gabrielfior: More context ref the implementation logic:
-> For Seer trading, we don't want markets with 0 liquidity
-> This filter would optimally get added to `seer_subgraph.get_markets`, however `outcomeSupply` (as prev implemented) doesn't filter out the markets with 0 liquidity (more on it later)
-> After fetching markets, each market is passed on to ` SeerAgentMarket.from_data_model_with_subgraph`, where the liquidity check occurs.

Some thoughts:
-> This check could be added to `before_process_market` - but this would imply creating a new class `SeerDeployableTraderAgent` with an overwritten `before_process_market` (or maybe you have other ideas?)
-> Ideally I would do this filter on a subgraph level (somewhere in the `subgraph` class), but not possible to merge data from different subgraphs directly. So one needs to fetch e.g. 10 markets, and then do another call for all pools which have those tokens, then you do the liquidity logic. Can also be implemented, but a bit more work than the prev thought.

prediction_market_agent_tooling/markets/seer/seer.py
gabrielfior: removed

prediction_market_agent_tooling/markets/seer/seer.py
kongzii: > -> For Seer trading, we don't want markets with 0 liquidity

For trading on any PM Platform we don't want markets with 0 liquidity.

If you are doing it in `from_data_model_with_subgraph` right now, then it should be about the same complexity to do it in get_markets function [in here](https://github.com/gnosis/prediction-market-agent-tooling/blob/c2d4bf6e0027930340e496c28f2bbfc18fbe7f68/prediction_market_agent_tooling/markets/seer/seer.py#L329)?

And there you can do it only if `filter_by: FilterBy = FilterBy.OPEN`.

Something like:

```
markets = [
            market
            for m in markets
            if (
                market := SeerAgentMarket.from_data_model_with_subgraph(
                    model=m, seer_subgraph=seer_subgraph
                )
            )
            is not None
        ]

# Double check for liquidity in market, because graph isn't always up-to-date.
if filter_by == FilterBy.OPEN:
    markets = [m for m in markets if m.has_liquidity()]
```

kongzii (approved):

PR #712, opened 2025-05-26
Hotfix/seer verify market

gabrielfior:
-> `have_bet_on_market_since` (https://github.com/gnosis/prediction-market-agent-tooling/blob/7f647917a161dc565663d513bb96240f46f05229/prediction_market_agent_tooling/markets/markets.py#L97-L125) didn't support Seer (see log below)
-> Refactored this to be a `AgentMarket` method, so that, if we add new markets, this will fail since method will throw an exception in the parent class.

image


prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -265,16 +261,13 @@ def check_min_required_balance_to_operate(self, market_type: MarketType) -> None
                 f"{api_keys=} doesn't have enough operational balance."
             )
 
-    def have_bet_on_market_since(self, market: AgentMarket, since: timedelta) -> bool:
-        return have_bet_on_market_since(keys=APIKeys(), market=market, since=since)
-
```
gabrielfior, line 270: This became a market function

prediction_market_agent_tooling/tools/cow/cow_order.py
```diff
@@ -292,3 +288,38 @@ def get_trades_by_owner(
...
+    stop=stop_after_attempt(3),
+    wait=wait_fixed(1),
+    after=lambda x: logger.debug(f"get_trades_by_owner failed, {x.attempt_number=}."),
+)
+def get_orders_by_owner(
+    owner: ChecksumAddress,
+) -> list[Order]:
+    """Retrieves all orders with pagination."""
+    items = paginate_endpoint(f"https://api.cow.fi/xdai/api/v1/account/{owner}/orders")
+    return [Order.model_validate(i) for i in items]
```
gabrielfior, line 303: Using pagination because a. cow doesn't support it and b. cow doesn't let us sort either

prediction_market_agent_tooling/markets/seer/seer.py
```diff
@@ -272,6 +275,20 @@ def redeem_winnings(api_keys: APIKeys) -> None:
...
 
+    def have_bet_on_market_since(self, keys: APIKeys, since: timedelta) -> bool:
+        """Check if the user has placed a bet on this market since a specific time using Cow API."""
+        # Cow endpoint doesn't allow us to filter by time.
+        start_time = utcnow() - since
+        prev_orders = get_orders_by_owner(owner=keys.bet_from_address)
+        for order in prev_orders:
+            if order.creationDate >= start_time and {
+                Web3.to_checksum_address(order.sellToken),
+                Web3.to_checksum_address(order.buyToken),
```
kongzii, line 286: Below in the models, these fields are already typed as ChecksumAddress, here I see two options now:

1. They indeed are ChecksumAddress already, and  `Web3.to_checksum_address` can be removed here
2. They are typed as ChecksumAddress, but in reality aren't (they are just strings at the end of the day, but misformated strings), and the type should be change to `str` in the models, so we know they aren't checksummed yet

Idk which of the scenarios is true, but check it out and implement the right one please.

PR #714, opened 2025-05-27
Gabriel/swap via swapr as fallback

gabrielfior:
-> Cow is failing (one example below) when outcome token amounts being traded are small. In those cases, we should trade directly via Swapr pools.

prediction_market_agent_tooling.loggers:_handle_exception:77 - Uncaught exception: Timeout waiting for order root='0x7ea8ecf4bd2857e11c3cf1c614c71db8d4db257ca0131953f8f7a745a68de6f7d0363ccd573163df94b754ca00c0aca2bb66b7486835ce1c' to be completed. https://api.cow.fi/xdai/api/v1/orders/0x7ea8ecf4bd2857e11c3cf1c614c71db8d4db257ca0131953f8f7a745a68de6f7d0363ccd573163df94b754ca00c0aca2bb66b7486835ce1c"


prediction_market_agent_tooling/tools/cow/cow_order.py
```diff
@@ -292,3 +358,33 @@ def get_trades_by_owner(
...
     response.raise_for_status()
     return [MinimalisticToken.model_validate(i) for i in response.json()]
+
+
+async def cancel_order(
+    order_uids: Sequence[str],
+    api_keys: APIKeys,
+    chain: Chain = Chain.GNOSIS,
+    env: Envs = "prod",
+) -> None:
```
gabrielfior, line 368: I tested both manually and added a new test (see `test_cow_cancellation) - it passes hence I'm confident in the functionality.
Deactivating however to not spam Cow API.

prediction_market_agent_tooling/markets/seer/seer_contracts.py
```diff
@@ -110,3 +115,42 @@ def redeem_to_base(
...
+        web3: Web3 | None = None,
+    ) -> TxReceipt:
+        # ToDo - check allowance
+        erc20_token = ContractERC20OnGnosisChain(
+            address=Web3.to_checksum_address(params.token_in)
+        )
+
+        if erc20_token.allowance(
+            api_keys.bet_from_address, self.address, web3=web3
+        ) < Wei(params.amount_in):
```
kongzii, line 146: If it's Wei, should be already typed as Wei on the model

prediction_market_agent_tooling/markets/seer/seer_contracts.py
```diff
@@ -110,3 +115,42 @@ def redeem_to_base(
...
+            erc20_token.approve(
+                api_keys, self.address, Wei(params.amount_in), web3=web3
```
kongzii, line 148: And then you don't have to call Wei(...) on it again and again

prediction_market_agent_tooling/markets/seer/swap_pool_handler.py
```diff
@@ -0,0 +1,94 @@
...
+        token_in: ChecksumAddress,
+        price_outcome_token: float,
+        buffer_pct: float = 0.05,
+    ) -> float:
+        is_buying_outcome = token_in == self.collateral_token_address
+
+        if is_buying_outcome:
+            return amount_wei.value * (1.0 - buffer_pct) / price_outcome_token
+        else:
+            return amount_wei.value * price_outcome_token * (1.0 - buffer_pct)
```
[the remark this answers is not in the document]
kongzii: Agree with rabbit, why it's returning float, if it's converted to int anyway?

The Token / OutcomeToken /Wei / OutcomeWei types took quite some time to implement, but already caught many bugs.

If they aren't used in new code, it was useless. Use them please. 🙏

prediction_market_agent_tooling/markets/seer/seer_contracts.py
```diff
@@ -110,3 +115,42 @@ def redeem_to_base(
...
+        "0xffb643e73f280b97809a8b41f7232ab401a04ee1"
+    )
+
+    def exact_input_single(
+        self,
+        api_keys: APIKeys,
+        params: ExactInputSingleParams,
+        web3: Web3 | None = None,
+    ) -> TxReceipt:
+        # ToDo - check allowance
```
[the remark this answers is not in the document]
gabrielfior: old ToDo, it was already implemented

prediction_market_agent_tooling/markets/seer/seer_contracts.py
gabrielfior: Problem with `Wei` is that, when using `model_dump`, it gets passed to the smart contracts like below
```
class A(BaseModel):
    b: Wei

a = A(b=Wei(123))
a.model_dump()
Out[21]: {'b': {'value': 123, 'type': 'Wei'}}
```
So that's the reason I still use `int` when defining classes again and again.

kongzii (approved):

PR #725, opened 2025-06-10
Update to categorical result mapping

kongzii:


TS9001:
Approved given that test errors are unrelated.

PR #736, opened 2025-06-20
Gabriel/pools trading fallback seer

gabrielfior:
When cowswap trades don't go through, exception is thrown, gets logged and trade happens via Swapr pools directly

### DIFF
--- a/tests_integration_with_local_chain/markets/seer/test_seer_agent_market.py
+++ b/tests_integration_with_local_chain/markets/seer/test_seer_agent_market.py
@@ -45,7 +45,6 @@ def test_seer_place_bet(
             auto_deposit=False,
             web3=local_web3,
         )
-    # trick to get the wrapped exception from tenacity
     exception_message = str(e)
 
     assert (


prediction_market_agent_tooling/markets/omen/cow_contracts.py
```diff
@@ -19,6 +19,10 @@ class CowGPv2SettlementContract(ContractOnGnosisChain):
         )
     )
 
+    address: ChecksumAddress = Web3.to_checksum_address(
```
gabrielfior, line 22: https://docs.cow.fi/cow-protocol/reference/contracts/core#deployments

kongzii: Are you sure this should be hard-coded? See https://github.com/gnosis/prediction-market-agent-tooling/blob/05ab13c2f482b2af79d49ab94efd975ff99ff02b/prediction_market_agent_tooling/tools/cow/cow_order.py#L332, where it's passed dynamically based on API's output. Seer doesn't return it somewhere?

kongzii: Actually, it doesn't seem like you are using `CowGPv2SettlementContract` anywhere except for removing the dynamic address initialisation, any reason to do it then?

kongzii (approved):

gabrielfior:
https://github.com/gnosis/prediction-market-agent-tooling/pull/736#discussion_r2160762110
@kongzii 
I intended to do look it was previously (see below)
```
CowGPv2SettlementContract(
        address=Web3.to_checksum_address(
            check_not_none(posted_order.settlementContract).root
        )
    )
    ```
   but mypy was complaining. Hence I looked at the docs and realized there is just 1 SettlementContract, hence fixed the address. If there's an easy fix for mypy, happy to implement it

File: prediction_market_agent_tooling/tools/cow/models.py [hunk 2]
```diff
@@ -14,3 +15,9 @@ class Order(BaseModel):
     sellToken: str
     buyToken: str
     creationDate: DatetimeUTC
+
+
+class RateLimit(SQLModel, table=True):
+    __tablename__ = "rate_limit"
+    id: str = Field(primary_key=True)
+    last_called_at: DatetimeUTC
```
gabrielfior, line 23: Added a new table to make sure Cow api rate-limit of 1req/s is respected across all agents

tests_integration_with_local_chain/markets/seer/test_seer_agent_market.py
```diff
@@ -34,21 +34,23 @@ def test_seer_place_bet(
...
             outcome=agent_market.outcomes[0],
             amount=amount,
-            auto_deposit=True,
+            auto_deposit=False,
             web3=local_web3,
         )
     # trick to get the wrapped exception from tenacity
-    exception_message = e.value.last_attempt.exception().message  # type: ignore
+    exception_message = str(e)
+
```
gabrielfior, line 50: I'm accessing the exception differently because I tell tenacity to raise the original exception, instead of throw `RetryError`

PR #738, opened 2025-06-23
Added support for Seer Scalar markets

TS9001:
Solves https://github.com/gnosis/prediction-market-agent/issues/908

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "prediction-market-agent-tooling"
-version = "0.66.3"
+version = "0.66.4"
 description = "Tools to benchmark, deploy and monitor prediction market agents."
 authors = ["Gnosis"]
 readme = "README.md"


prediction_market_agent_tooling/markets/seer/seer.py
```diff
@@ -348,10 +348,19 @@ def from_data_model_with_subgraph(
...
         )
 
         return market
 
+    # TODO Decide how to handle this in the future
+    @staticmethod
+    def convert_from_wei(value: int | None) -> int | None:
+        if value and value >= 10**18:
+            return value // 10**18
+        return value
```
TS9001, line 362: This is here because upper-lower bound can be (but usually is not) wei... I checked how they are doing it in https://github.com/seer-pm/demo/blob/42615a1c5278e924be81245deb557f0f2c427127/web/src/components/Market/AnswerForm.tsx#L47

I don't think that they really went with scalar forms btw... probably worth to ask them about ...

prediction_market_agent_tooling/markets/data_models.py
```diff
@@ -165,6 +203,30 @@ def from_probabilistic_answer(
...
         )
 
+    @staticmethod
+    def from_scalar_answer(
+        answer: ScalarProbabilisticAnswer,
+        market_outcomes: Sequence[OutcomeStr] | None = None,
+    ) -> "CategoricalProbabilisticAnswer":
+        probabilities = {}
+        if market_outcomes and UP_OUTCOME_SEER_IDENTIFIER in market_outcomes:
+            probabilities[OutcomeStr(UP_OUTCOME_SEER_IDENTIFIER)] = answer.p_up
```
gabrielfior, line 213: I would be a bit more dramatic and do something like

```
if not market.outcomes:
    raise ValueError("Market with no outcomes")

if not set([DOWN_OUTCOME_SEER_IDENTIFIER, UP_OUTCOME_SEER_IDENTIFIER]).issubsetof(market.outcomes): raise ValueError
```

prediction_market_agent_tooling/markets/seer/seer.py
```diff
@@ -348,10 +348,19 @@ def from_data_model_with_subgraph(
             resolution=None,
             volume=None,
             probabilities=probability_map,
+            upper_bound=SeerAgentMarket.convert_from_wei(model.upper_bound),
```
gabrielfior, line 351:
```suggestion
            upper_bound=from_wei_inc_negative(model.upper_bound),
```

gabrielfior: Assume these values will always be 1e18, at least on Seer

prediction_market_agent_tooling/markets/seer/seer_constants.py
```diff
@@ -0,0 +1,3 @@
+UP_OUTCOME_SEER_IDENTIFIER = "UP"
+DOWN_OUTCOME_SEER_IDENTIFIER = "DOWN"
+INVALID_RESULT_OUTCOME_SEER_IDENTIFIER = "Invalid result"
```
gabrielfior, line 3: I don't love this - can we always refer to UP_OUTCOME_LOWERCASE or similar, and do `.lower()` when needed? Otherwise we have too many constants

prediction_market_agent_tooling/markets/seer/seer_subgraph_handler.py
```diff
@@ -122,6 +127,25 @@ def _build_where_statements(
                 ]
             ]
 
+        if not include_categorical_markets and not include_conditional_markets:
+            # Create single OR conditions with all variations
```
gabrielfior, line 131: Please separate the `if` statement from categorical markets, I think it's easier

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -383,7 +390,21 @@ def build_answer(
...
+                return (
+                    CategoricalProbabilisticAnswer.from_scalar_answer(
+                        scalar_answer,
+                        market.outcomes,
+                    )
+                    if scalar_answer is not None
+                    else None
+                )
+            except NotImplementedError:
+                logger.info(
```
kongzii, line 405: Is this wise to do? Categorical prediction will have something like

```
Q: What will be price of GNO 27.6.2025?
Possible answers: UP, DOWN
```

I think such an agent can not make any reasonable prediction, or?

Instead of defaulting to `answer_categorical_market`, in agents without `answer_scalar_market` implemented, I'd just wouldn't fetch scalar markets at all.

You can check [fetch_categorical_markets property](https://github.com/gnosis/prediction-market-agent-tooling/blob/2d50d65a630b1683f921b3b2654fac535d51791a/prediction_market_agent_tooling/deploy/agent.py#L315) to see how it can be done if you agree.

And then here can be just raised exception with "implementation missing" just to be sure.

prediction_market_agent_tooling/markets/data_models.py
```diff
@@ -100,6 +105,36 @@ def to_boolean_outcome(value: str | bool) -> bool:
 Decision = Annotated[bool, BeforeValidator(to_boolean_outcome)]
 
 
+class ScalarProbabilisticAnswer(BaseModel):
+    scalar_value: float
+    upperBound: int
```
kongzii, line 110: Are these bounds in Wei or normal units like the scalar_value?

If they are Weis, please use `Wei` type, if they are normal values, shouldn't they be floats as well as the `scalar_value`?

prediction_market_agent_tooling/markets/data_models.py
```diff
@@ -165,6 +200,46 @@ def from_probabilistic_answer(
...
+            [outcome.lower() for outcome in market_outcomes] if market_outcomes else []
+        )
+
+        if not market_outcomes:
+            raise ValueError("Market with no outcomes")
+
+        if not set(
+            [
+                DOWN_OUTCOME_SEER_IDENTIFIER_LOWERCASE,
+                UP_OUTCOME_SEER_IDENTIFIER_LOWERCASE,
```
kongzii, line 219: There is already `UP_OUTCOME_LOWERCASE_IDENTIFIER` defined (without SEER`), shouldn't that be used here? As this is general model, not Seer-related.

I think for now we can have just the non-SEER constant, and add granuallity if needed later.

prediction_market_agent_tooling/markets/data_models.py
```diff
@@ -165,6 +200,46 @@ def from_probabilistic_answer(
...
+        answer: ScalarProbabilisticAnswer,
+        market_outcomes: Sequence[OutcomeStr] | None = None,
+    ) -> "CategoricalProbabilisticAnswer":
+        probabilities = {}
+        lowercase_market_outcomes = (
+            [outcome.lower() for outcome in market_outcomes] if market_outcomes else []
+        )
+
+        if not market_outcomes:
+            raise ValueError("Market with no outcomes")
```
kongzii, line 214: Why is `market_outcomes: Sequence[OutcomeStr] | None` accepted here, if it can not work with `None` value being provided? And it even defaults to None?

Couldn't the type be just `market_outcomes: Sequence[OutcomeStr]` to let the caller now (and have it statically type-checked) that market_outcomes is indeed required and it doesn't work without it?

prediction_market_agent_tooling/markets/data_models.py
```diff
@@ -165,6 +200,46 @@ def from_probabilistic_answer(
...
+
+        if not set(
+            [
+                DOWN_OUTCOME_SEER_IDENTIFIER_LOWERCASE,
+                UP_OUTCOME_SEER_IDENTIFIER_LOWERCASE,
+            ]
+        ).issubset(lowercase_market_outcomes):
+            raise ValueError("Market with no outcomes")
+
+        probabilities[OutcomeStr(UP_OUTCOME_SEER_IDENTIFIER_LOWERCASE)] = answer.p_up
```
kongzii, line 224: Again `SEER`-constant used in the general model. (and below as well)

prediction_market_agent_tooling/markets/seer/seer.py
kongzii: Agreed, and `convert_from_wei` can be removed. 

We have the `Wei`, `OutcomeWei`, `CollateralToken`, `OutcomeToken`, etc. types so we can distinguish what the given functions returns/accept.

The problem with just using floats or ints is that it's easy to get in bugs (already happened many times) without noticing them.

For example, [sell_tokens in the omen market](https://github.com/gnosis/prediction-market-agent-tooling/blob/2d50d65a630b1683f921b3b2654fac535d51791a/prediction_market_agent_tooling/markets/omen/omen.py#L262) accepts `OutcomeToken`. When you traded in the market, you bought let's say 10 OutcomeTokens for 5 CollateralToken (ie using sDai at probability 0.5).

Now when you want to sell it, it's a huge difference if you provide the sell_amount in collateral tokens (5), vs. outcome tokens (10).

Either the function will raise error (not enough of funds to sell), or we'd end up selling just half of what you wanted. 

And this is the case in many places, that's why we distinguish them at type-level.

prediction_market_agent_tooling/markets/seer/seer.py
```diff
@@ -508,6 +524,10 @@ def _swap_tokens_with_fallback(
                 f"Exception occured when swapping tokens via Cowswap, doing swap via pools. {e}"
             )
 
+            if not self.has_liquidity():
+                logger.error(f"Market {self.id} has no liquidity. Cannot place bet.")
+                raise e from e
```
kongzii, line 529:
```suggestion
                raise
```

Just raise works

prediction_market_agent_tooling/markets/seer/seer_constants.py
```diff
@@ -0,0 +1,3 @@
+UP_OUTCOME_SEER_IDENTIFIER_LOWERCASE = "up"
+DOWN_OUTCOME_SEER_IDENTIFIER_LOWERCASE = "down"
+INVALID_RESULT_OUTCOME_SEER_IDENTIFIER_LOWERCASE = "invalid result"
```
kongzii, line 3: We already have 

[INVALID_OUTCOME_LOWERCASE_IDENTIFIER = "invalid"](https://github.com/gnosis/prediction-market-agent-tooling/blob/c2d4bf6e0027930340e496c28f2bbfc18fbe7f68/prediction_market_agent_tooling/deploy/constants.py#L4)

from @gabrielfior , are they really using two different constants for invalid results, or do we have a bug?

prediction_market_agent_tooling/markets/seer/seer_subgraph_handler.py
```diff
@@ -98,31 +120,46 @@ def _build_where_statements(
...
-                    NO_OUTCOME_LOWERCASE_IDENTIFIER,
-                    NO_OUTCOME_LOWERCASE_IDENTIFIER.capitalize(),
-                    NO_OUTCOME_LOWERCASE_IDENTIFIER.upper(),
-                ]
-            ]
+        exclude_scalar_yes, exclude_scalar_no = {}, {}
+
+        # Return scalar markets.
+        if include_scalar_markets:
+            # We are interested in scalar markets only - this excludes categorical markets
```
kongzii, line 131: `include_scalar_markets` somehow sounds to me to `include ALSO scalar markets`, not to include ONLY scalar markets.

Which one is correct?

prediction_market_agent_tooling/markets/data_models.py
gabrielfior: Clement said always wei for new markets

prediction_market_agent_tooling/markets/data_models.py
TS9001: Done

prediction_market_agent_tooling/markets/seer/seer_constants.py
TS9001: I got rid of it completely and use market_antagonistic ones

prediction_market_agent_tooling/markets/seer/seer_subgraph_handler.py
```diff
@@ -98,12 +103,12 @@ def _build_where_statements(
             case _:
                 raise ValueError(f"Unknown filter {filter_by}")
 
-        if not include_conditional_markets:
+        if not include_conditional_markets and not include_scalar_markets:
```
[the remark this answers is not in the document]
TS9001: Good catch, gone

prediction_market_agent_tooling/markets/data_models.py
TS9001: No apparent reason - tried to keep format we have for categorical - gone

prediction_market_agent_tooling/markets/data_models.py
TS9001: yop, gone

prediction_market_agent_tooling/markets/seer/seer.py
TS9001: fixed

prediction_market_agent_tooling/markets/seer/seer_subgraph_handler.py
TS9001: yop, fixed

prediction_market_agent_tooling/markets/seer/seer.py
TS9001: Changed it to WEI - because the value are still sometimes pure int values the method must stay otherwise we would have very small numbers and have to convert it somewhere anyway.

prediction_market_agent_tooling/markets/seer/seer_constants.py
kongzii: Ah I see why now, because it's used like this:

Screenshot by Dropbox Capture

Can we keep the same pattern then? Or use the new one everywhere?

I don't really have preference, just think that it should be unified.

prediction_market_agent_tooling/markets/seer/seer.py
kongzii: But didn't Clement say that was the case only for old markets and shouldn't happen anymore?

kongzii: I see where you are getting now, but now it's a bit bugged:

```
    @staticmethod
    def normalize_value(value: int) -> Wei | None:
        if value and value >= 10**18:
            return Wei(value // 10**18)
        return Wei(value)
```

This isn't correct, because when you do `/ 10**18`, then it's no longer Wei, it's actually converting from Wei to a normal floating value. Before, this function was called `convert_from_wei` so I think that's where the bug come from.

However I still think this should be handled on the source model. What you can do in these weird cases is something like:

```python
def seer_normalize_wei(value: int | None) -> int | None:
    # See https://github.com/seer-pm/demo/blob/main/web/netlify/edge-functions/utils/common.ts#L22
    if value is None:
        return value
    is_in_wei = value > 1e10
    return value if is_in_wei else value * 10**18
    
SeerNormalizedWei = Annotated[Wei | None, BeforeValidator(seer_normalize_wei)]

class SeerMarket(BaseModel):
    upper_bound: SeerNormalizedWei = Field(alias="upperBound", default=None)
    lower_bound: SeerNormalizedWei = Field(alias="lowerBound", default=None)
```

This way, `SeerMarket` itself is correctly initialised, and users of it don't need to handle normalisation by themselves. (often we need to work with the non-AgentMarket type itself).

Note that I also changed the threshold value to check for, based on the Seer's codebase.

prediction_market_agent_tooling/markets/seer/seer.py
```diff
@@ -137,14 +141,17 @@ def get_token_in_usd(self, x: CollateralToken) -> USD:
...
 
-            if token_price:
-                return USD(x.value * token_price.value)
-            raise e
+    def get_colateral_price_from_pools(self) -> CollateralToken | None:
+        p = PriceManager.build(HexBytes(HexStr(self.id)))
+        token_price = p.get_token_price_from_pools(token=WRAPPED_XDAI_CONTRACT_ADDRESS)
+        if token_price is None:
+            token_price = p.get_token_price_from_pools(token=SDAI_CONTRACT_ADDRESS)
+        return token_price
```
TS9001, line 154: If cow fails we try to change for WEI or SDAI

prediction_market_agent_tooling/markets/seer/data_models.py
```diff
@@ -50,6 +51,17 @@ class CreateCategoricalMarketsParams(BaseModel):
...
 
+def seer_normalize_wei(value: int | None) -> int | None:
+    # See https://github.com/seer-pm/demo/blob/main/web/netlify/edge-functions/utils/common.ts#L22
+    if value is None:
+        return value
+    is_in_wei = value > 1e10
+    return value if is_in_wei else value * 10**18
+
+
+SeerNormalizedWei = Annotated[Wei | None, BeforeValidator(seer_normalize_wei)]
```
gabrielfior, line 62: This applies to old markets - for new markets, just assume it's Wei, no conversion necessary

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -321,6 +328,13 @@ def fetch_categorical_markets(self) -> bool:
             return True
         return False
 
+    @property
+    def fetch_scalar_markets(self) -> bool:
+        # Check if the subclass has implemented the answer_scalar_market method, if yes, fetch scalar markets as well.
+        if self.answer_scalar_market.__wrapped__.__func__ is not DeployablePredictionAgent.answer_scalar_market:  # type: ignore[attr-defined] # This works just fine, but mypy doesn't know about it for some reason.
+            return True
+        return False
```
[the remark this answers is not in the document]
kongzii: This is how categorical markets are handled as well.

It's nice because it follows the single source of truth principle and is less bug-prone on the user's end (if it's a normal true/false flag, the user can set it to True, yet it is not implemented this method, or the other way around).

> Problem here is, if we change the implementation of the function (removing the decorator), than it says there is no func although it exists, correct?

If the decorator is removed, then this will fail and needs to be fixed.

prediction_market_agent_tooling/markets/seer/data_models.py
kongzii: It's iff-ed in `seer_normalize_wei` to work with "new markets" as well. 

This follows Seer's code base, they don't distinguish based on creation date (old vs. new), but based on the same if as we have above.

prediction_market_agent_tooling/deploy/agent.py
```diff
@@ -383,7 +398,19 @@ def build_answer(
...
+                return (
+                    CategoricalProbabilisticAnswer.from_scalar_answer(
+                        scalar_answer,
+                        market.outcomes,
+                    )
+                    if scalar_answer is not None
+                    else None
+                )
+            except NotImplementedError:
+                logger.info("answer_scalar_market() not implemented!")
```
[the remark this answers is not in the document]
TS9001: Done

prediction_market_agent_tooling/markets/seer/seer.py
```diff
@@ -128,10 +133,39 @@ def store_trades(
...
+                f"Could not get quote for {self.collateral_token_contract_address_checksummed} from Cow, exception {e=}. Falling back to pools. "
+            )
+            token_price = self.get_colateral_price_from_pools()
+            if token_price is None:
+                raise e
+            return USD(x.value * token_price.value)
+
+    def get_colateral_price_from_pools(self) -> CollateralToken | None:
+        p = PriceManager.build(HexBytes(HexStr(self.id)))
+        token_price = p.get_token_price_from_pools(token=WRAPPED_XDAI_CONTRACT_ADDRESS)
```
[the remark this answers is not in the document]
TS9001: Yep as we discussed should be just SDAI

prediction_market_agent_tooling/markets/seer/seer.py
```diff
@@ -358,13 +400,16 @@ def get_markets(
...
+        fetch_scalar_markets: bool = False,
     ) -> t.Sequence["SeerAgentMarket"]:
         seer_subgraph = SeerSubgraphHandler()
         markets = seer_subgraph.get_markets(
             limit=limit,
             sort_by=sort_by,
             filter_by=filter_by,
             include_categorical_markets=fetch_categorical_markets,
+            include_only_scalar_markets=fetch_scalar_markets,
+            include_conditional_markets=not fetch_scalar_markets,  # scalar markets and conditional are currently exclusive
```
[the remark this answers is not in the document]
kongzii: Yes, but that means generally for all conditional markets. Not only scalar conditional markets.

So, conditional markets can get globally disabled in this PR, but I don't see why making them exclusive with scalar markets (e.g. no conditional markets only if scalar markets are requested).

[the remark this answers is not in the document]
TS9001: Yes you have a point. Done!

PR #742, opened 2025-06-30
Fix unprofitable bets in MultiCategoricalMaxAccuracyBettingStrategy

kongzii:


prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -196,11 +255,23 @@ def calculate_trades(
         market: AgentMarket,
     ) -> list[Trade]:
         """We place bet on only one outcome."""
-
         outcome_to_bet_on = self.calculate_direction(market, answer)
 
+        # Will be lowered if the amount that we would need to buy would be unprofitable.
+        actual_wanted_position = BettingStrategy.cap_to_profitable_position(
```
kongzii, line 261: I use this in `MultiCategoricalMaxAccuracyBettingStrategy` and not in `Kelly`, because only agents with `MultiCategoricalMaxAccuracyBettingStrategy` are erroring out, because Kelly scales the bet size automatically.

tests/tools/betting_strategies/test_betting_strategy.py
```diff
@@ -0,0 +1,58 @@
...
+from prediction_market_agent_tooling.markets.agent_market import AgentMarket, SortBy
+from prediction_market_agent_tooling.markets.omen.omen import OmenAgentMarket
+
+
+@pytest.mark.parametrize(
+    "bet_amount, market_class, sort_by",
+    list(
+        product(
+            [USD(1), USD(10), USD(100)],
+            [OmenAgentMarket],
```
kongzii, line 17: I wanted to add Seer as well, but at least some tests then errors out always. As the APIs there are quite flaky..

prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -94,6 +95,64 @@ def filter_trades(market: AgentMarket, trades: list[Trade]) -> list[Trade]:
...
+    @staticmethod
+    def cap_to_profitable_position(
+        market: AgentMarket,
+        existing_position: USD,
+        wanted_position: USD,
+        outcome_to_bet_on: OutcomeStr,
+    ) -> USD:
+        # If the wanted position is lower, it means the agent is gonna sell and that's profitable always.
+        if wanted_position > existing_position:
+            difference = wanted_position - existing_position
```
kongzii, line 146: We need to cap only the difference. Imagine the situation where the agent bought 10$ shares at probability 50%.

But now the probability is already skewed to 95%.

Buying a lot more probably won't be profitable and needs to be capped, but the existing bet is still profitable and can be kept.

TS9001 (approved):
LGTM. Great work!

PR #744, opened 2025-06-30
Add option to not take profits in betting strategy

kongzii:


prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -209,10 +212,22 @@ def _build_rebalance_trades_from_positions(
...
+
             if diff_amount == 0:
                 continue
 
             trade_type = TradeType.SELL if diff_amount < 0 else TradeType.BUY
+
+            # We work with positions, so imagine following scenario: Agent invested $10 when probs were 50:50,
+            # now the probs are 99:1 and his initial $10 is worth $100.
+            # If `take_profit` is set to False, agent won't sell the $90 to get back to the $10 position.
+            if (
```
kongzii, line 224: Double think with me here, please 😄 

If take profit is set to False, it means agent won't take profit from existing bets, that are now higher in value.

However, if agent decided to sell the whole position (which happens if he wants to buy the other one), he should sell it, right?

So:

1. `not self.take_profit` --> skip only if take profit is disabled (by default it's true)
2. `target_amount > 0` -->  if it is zero, it means he wants to sell the whole position
3. `trade_type == TradeType.SELL` --> this one seems clear

TS9001 (approved):

PR #767, opened 2025-07-07
Fix for binary markets not being returned correctly on seer

TS9001:


prediction_market_agent_tooling/markets/seer/seer_subgraph_handler.py
```diff
@@ -135,7 +135,7 @@ def _build_where_statements(
             no_stms = SeerSubgraphHandler._create_case_variations_condition(
                 DOWN_OUTCOME_LOWERCASE_IDENTIFIER, "outcomes_contains", "or"
             )
-        elif include_conditional_markets and not include_categorical_markets:
+        elif not include_categorical_markets:
```
gabrielfior, line 138: I would have expected
```
elif not include_conditional_markets and not include_categorical_markets:
```

kongzii: Seems to me a lot of the confusion/complexity here is because the scalar filter works differently than the conditional and categorical filters.

For scalar, it's `only scalar`, for others it's `include also this`.

And also that `yes_stms` / `no_stms` is used for scalar markets as well (see lines 132-137).

And then we default to binary markets, but we don't have to.

I'd suggest refactoring `include_only_scalar_markets` --> `include_scalar_markets`, and add `include_binary_markets`, then we can have:

```python
if not include_conditional_markets:
    and_stms["parentMarket"] = ADDRESS_ZERO.lower()

if not include_scalar_markets and not include_binary_markets and not include_categorical_markets:
   raise RuntimeError("At least some markets shoudl have been included!")

outcomes_stm = {}

if include_scalar_markets and include_binary_markets and include_categorical_markets:
    # Pass, anything can be in the outcomes

else:
    if not include_scalar_markets:
        # UP/DOWN can not be in the outcomes
   
    if include_binary_markets and not include_categorical_markets:
         # Yes/No must be in the outcomes
```

Because, in theory, nothing stops us from having an agent that implements both `answer_scalar_market` and `answer_categorical_market` anyway.

Wdyt? Am I missing some condition in the code?

kongzii: On the second thought, though.. Maybe even easier would be to have

`include_ONLY_scalar_markets`, `include_ONLY_binary_markets`, `include_ONLY_categorical_markets`.

And we can assume that one agent can implement only one of these methods. This should be asserted in the `__init__` method of DeployableAgent class.

And the code can be:

```python
if not include_conditional_markets: 
    and_stms["parentMarket"] = ADDRESS_ZERO.lower() 

if include_scalar_markets:
   # Outcome must contain UP/DOWN
elif include_binary_markets:
   # Outcoem must contain Yes/No
elif include_categorical_markets:
   # template id must be 2 
else:
   raise RuntimeError("At least some markets shoudl have been included!")
```

Wdyt2?

TS9001: I like this idea, pushing commit with the fix, incl enum type for this proposed by @gabrielfior

prediction_market_agent_tooling/markets/metaculus/metaculus.py
```diff
@@ -66,14 +67,15 @@ def from_data_model(model: MetaculusQuestion) -> "MetaculusAgentMarket":
...
+    def get_markets(  # type: ignore[override]
         limit: int,
         sort_by: SortBy = SortBy.NONE,
         filter_by: FilterBy = FilterBy.OPEN,
         created_after: t.Optional[DatetimeUTC] = None,
         excluded_questions: set[str] | None = None,
         tournament_id: int | None = None,
-        fetch_scalar_markets: bool = False,
+        market_types: list[MarketType] = [MarketType.ALL],
+        include_conditional_markets: bool = False,
```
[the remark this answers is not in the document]
TS9001: Fixed

prediction_market_agent_tooling/markets/seer/seer_subgraph_handler.py
```diff
@@ -123,43 +127,46 @@ def _build_where_statements(
...
-            or include_conditional_markets
-        ):
-            # We should not provide any scalar markets because they are exclusive for categorical markets
-            exclude_scalar_yes = SeerSubgraphHandler._create_case_variations_condition(
-                UP_OUTCOME_LOWERCASE_IDENTIFIER, "outcomes_not_contains", "and"
-            )
-            exclude_scalar_no = SeerSubgraphHandler._create_case_variations_condition(
-                DOWN_OUTCOME_LOWERCASE_IDENTIFIER, "outcomes_not_contains", "and"
+        elif market_type == MarketType.CATEGORICAL:
+            # Template ID 2 OR Template ID 3 (includes both binary and categorical)
```
kongzii, line 155:
```suggestion
            # Template ID 2 (categorical) OR Template ID 3 (multi-categorical, we treat them as categorical for now for simplicity)
```

See https://reality.eth.limo/app/docs/html/contracts.html#templates

prediction_market_agent_tooling/markets/seer/seer_subgraph_handler.py
```diff
@@ -123,43 +127,46 @@ def _build_where_statements(
...
+            # Return flat structure to avoid nesting issues
+            outcome_filters.append(
+                {
+                    "or": [
+                        {"templateId": 2},
+                        {"templateId": 3},
```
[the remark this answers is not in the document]
TS9001: great idea, fixed

PR #769, opened 2025-07-11
Gabriel/fix polymarket

gabrielfior:


prediction_market_agent_tooling/markets/polymarket/polymarket.py
```diff
@@ -67,9 +68,11 @@ def build_resolution_from_condition(
         ]
         # For a binary market, there should be exactly one payout numerator greater than 0.
         if len(payout_numerator_indices_gt_0) != 1:
-            raise ValueError(
+            # ToDo: implement this when supporting multi-categorical markets
```
kongzii, line 71: Why remove the raise? What's flow where we ask for non-binary market from Polymarket, and are okay with not having resolution? 🤔

prediction_market_agent_tooling/markets/polymarket/data_models.py
```diff
@@ -66,7 +60,7 @@ class PolymarketGammaResponseDataItem(BaseModel):
     slug: str
     volume: float | None = None
     startDate: DatetimeUTC
-    endDate: DatetimeUTC
+    endDate: DatetimeUTC | None = None
```
kongzii, line 63: What is an example of a market without an ending date? 🤔 It should be kind of necessary.

prediction_market_agent_tooling/markets/polymarket/polymarket.py
kongzii: If you are going to set replication of categorical markets from Polymarket before #770, and this situation needs to work, then let's keep it.

If it's not really required, then I wouldn't change something that works, so we are explicitly notified if it's used without intention.

kongzii (approved):

Issue #746, opened 2025-07-01
Fill resolution in seer agent market


kongzii:
Right now, the resolution in SeerAgentMarket is hard-coded to None: https://github.com/gnosis/prediction-market-agent-tooling/blob/a093e618acbab28692f9e749cbcd14b9f80343a1/prediction_market_agent_tooling/markets/seer/seer.py#L346

However, subgraph does return questions with resolutions on markets: 

![Image](https://github.com/user-attachments/assets/35fd6de7-6fd0-47d9-83ef-595c34ee85c5)

However 2, it doesn't work with subgrounds for some reason: https://github.com/gnosis/prediction-market-agent-tooling/blob/a093e618acbab28692f9e749cbcd14b9f80343a1/prediction_market_agent_tooling/markets/seer/seer_subgraph_handler.py#L70

Another issue is, that AgentMarket expect just 1 resolution, but because of multi-categorical markets, several resolutions can be available on Seer market.

This issue is to somehow figure it all out!

Issue #740, opened 2025-06-26
Replace subgrounds with regular httpx calls (that enable caching)


gabrielfior:
subgrounds (https://github.com/0xPlaygrounds/subgrounds) hasn't been updated in a while, and we use it simply for subgraph pagination.
We can do that on our own, increase control over the process and stop relying on a 3rd party package for it.

Calling all contributors!

PR #774, opened 2025-07-14
Fill resolution seer

gabrielfior:
Initial draft for Seer resolutions

Closes https://github.com/gnosis/prediction-market-agent-tooling/issues/746


prediction_market_agent_tooling/markets/seer/seer.py
```diff
@@ -370,27 +431,12 @@ def from_data_model_with_subgraph(
...
-                                False,
-                            )
-                        )
-                    ),
-                    parent_outcome=model.parent_outcome,
-                )
-                if model.parent_market
-                else None
-            ),
+            parent=parent,
```
gabrielfior, line 439: Abstracted to `get_parent`

prediction_market_agent_tooling/markets/seer/seer_subgraph_handler.py
```diff
@@ -264,17 +264,78 @@ def get_markets(
...
         fields = self._get_fields_for_markets(markets_field)
         markets = self.do_query(fields=fields, pydantic_model=SeerMarket)
-        return markets
+        market_ids = [m.id for m in markets]
+        questions = self.get_questions_for_markets(market_ids)
+
+        # organize questions by property question.market.id
+        questions_by_market_id = defaultdict(list)
+        for q in questions:
+            questions_by_market_id[q.market.id.hex()].append(q)
```
gabrielfior, line 273: It's possible to do `grouping` using [subgrounds](https://docs.playgrounds.network/subgrounds/getting_started/querying/grouping/) - however, this simply fetches more items, and it's not trivial to separate the `markets` from the `questions` if they were to be fetched in the same query.
Hence, the solution I went with is to fetch questions separately, incurring 1 more graph call per function (`get_markets`, `get_market_by_id`) - note that we have a questions_lookup for parent markets which should help a bit here.
The best approach would be to remove `subgrounds` and implement it from scratch (see https://github.com/gnosis/prediction-market-agent-tooling/issues/740)

prediction_market_agent_tooling/markets/seer/data_models.py
```diff
@@ -48,17 +48,36 @@ class CreateCategoricalMarketsParams(BaseModel):
 SEER_BASE_URL = "https://app.seer.pm"
 
 
-def seer_normalize_wei(value: int | None) -> int | None:
+def seer_normalize_wei(value: int | dict[str, t.Any] | None) -> int | None:
```
gabrielfior, line 51: Changes required due to `model_dump` with Wei values

prediction_market_agent_tooling/markets/seer/questions_cache.py
```diff
@@ -0,0 +1,49 @@
...
+    Attributes:
+        market_id_to_questions: A dictionary mapping market IDs to lists of SeerMarketQuestions
+        seer_subgraph_handler: Handler for interacting with the Seer subgraph
+    """
+
+    def __init__(self, seer_subgraph_handler: SeerSubgraphHandler | None = None):
+        self.market_id_to_questions: dict[
+            HexBytes, list[SeerMarketQuestions]
+        ] = defaultdict(list)
+        self.seer_subgraph_handler = seer_subgraph_handler or SeerSubgraphHandler()
```
gabrielfior, line 28: This should hold questions during 1 agent run - we don't expect resolutions to change during that timespan, hence Singleton makes sense

prediction_market_agent_tooling/markets/seer/questions_cache.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def fetch_questions(self, market_ids: list[HexBytes]) -> list[SeerMarketQuestions]:
+        filtered_list = [
+            market_id
+            for market_id in market_ids
+            if market_id not in self.market_id_to_questions
+        ]
```
gabrielfior, line 35: We don't want to fetch previously fetched markets

TS9001 (approved):

PR #795, opened 2025-07-21
Fix betting strategy optim script

kongzii:


prediction_market_agent_tooling/markets/agent_market.py
```diff
@@ -147,6 +147,23 @@ def handle_legacy_fee(cls, data: dict[str, t.Any]) -> dict[str, t.Any]:
         if "fees" not in data and "fee" in data:
             data["fees"] = MarketFees(absolute=0.0, bet_proportion=data["fee"])
             del data["fee"]
+        # Backward compatibility for older `AgentMarket` without `probabilities`.
+        if "probabilities" not in data and "current_p_yes" in data:
```
gabrielfior, line 151: What's the agent flow that would lead to `current_p_yes` and not `probabilities` being generated?

kongzii: Older data from Langfuse has the older format. How else would you solve it?

TS9001 (approved):

PR #812, opened 2025-07-29
CategoricalKelly + fixes

kongzii:
I'm not sure if this still should be called Kelly, but I don't have a better name, so here we go! 😄 

The goal here is to have a betting strategy that is designed for categorical markets.

Because these strategies generally "just give some number given input probs", it's hard to tell if it's "correct". What I did is:

1. Implement a few tests that test no-brainer cases: Buy if significantly underpriced, short if overpriced, do nothing if the market's probs are just right.
2. Include CategoricalKelly as one of the strategies in the betting strategy optimization script. It should be at least competitive with BinaryKelly.

After running the benchmark for `DeployablePredictionProphetGPT4oAgent`, the best strategy, acording to Optuna's `best_trial`, which is defined as `trial located at the Pareto front in the study`, is still BinaryKelly (in this experiment at least, in other runs, CategoricalKelly was winning, there is some randomness):

```
[5 / 5] Best value for DeployablePredictionProphetGPT4oAgent (params: {'strategy_name': 'BinaryKellyBettingStrategy', 'bet_amount': 24.810272693356964, 'max_price_impact': 0.9585528322007394, 'force_simplified_calculation': False}, n train bets: 931, n test bets: 157): Training maximization: [224.08885170673798, 345.7774583185099, 315.18831471615124, 275.49461200447683, 400.59789034508907] Testing profit: 338.49 Original profit on Testing: 29.77 (testing dates 2025-07-21 to 2025-07-27)
```

However, if we simply take strategies with top maximization value, `CategoricalKellyBettingStrategy (simplified)` wins and `CategoricalKellyBettingStrategy (full)` is at least not last. 

---

## DeployablePredictionProphetGPT4oAgent

1140 bets

| strategy                                     |   p_yes_mse |   total_bet_amount |   total_bet_profit |   total_simulated_amount |   total_simulated_profit |    roi |   simulated_roi |   maximize |
|:---------------------------------------------|------------:|-------------------:|-------------------:|-------------------------:|-------------------------:|-------:|----------------:|-----------:|
| CategoricalKellyBettingStrategy (simplified) |    0.196453 |             200.47 |            7.15539 |                 1111.71  |                 422.962  | 3.5693 |         38.046  |   422.962  |
| BinaryKellyBettingStrategy (full)            |    0.196453 |             200.47 |            7.15539 |                  590.521 |                 400.692  | 3.5693 |         67.854  |   400.692  |
| MaxAccuracyWithKellyScaledBetsStrategy       |    0.196453 |             200.47 |            7.15539 |                 1355.68  |                 308.301  | 3.5693 |         22.7414 |   308.301  |
| BinaryKellyBettingStrategy (simplified)      |    0.196453 |             200.47 |            7.15539 |                 1173.96  |                 216.418  | 3.5693 |         18.4348 |   216.418  |
| CategoricalKellyBettingStrategy (full)       |    0.196453 |             200.47 |            7.15539 |                  862.436 |                 140.008  | 3.5693 |         16.234  |   140.008  |
| MaxExpectedValueBettingStrategy              |    0.196453 |             200.47 |            7.15539 |                  818.023 |                 134.305  | 3.5693 |         16.4183 |   134.305  |
| CategoricalMaxAccuracyBettingStrategy        |    0.196453 |             200.47 |            7.15539 |                  517.302 |                  68.5841 | 3.5693 |         13.258  |    68.5841 |

---

A big limitation in this PR is that, while these Kelly implementations return bet sizes (including negative for shorting) across all outcomes, the `BettingStrategy` at the moment can handle only betting at one of them (the biggest one).

Follow-up PRs would be:

1. Allow the BettingStrategy class to handle multiple bets.
2. Allow shorting

### DIFF
--- a/prediction_market_agent_tooling/tools/betting_strategies/kelly_criterion.py
+++ b/prediction_market_agent_tooling/tools/betting_strategies/kelly_criterion.py
@@ -1,6 +1,16 @@
+import numpy as np
+from scipy.optimize import minimize
+
 from prediction_market_agent_tooling.gtypes import CollateralToken, OutcomeToken
+from prediction_market_agent_tooling.markets.agent_market import AgentMarket
 from prediction_market_agent_tooling.markets.market_fees import MarketFees
-from prediction_market_agent_tooling.tools.betting_strategies.utils import SimpleBet
+from prediction_market_agent_tooling.markets.omen.omen import (
+    get_buy_outcome_token_amount,
+)
+from prediction_market_agent_tooling.tools.betting_strategies.utils import (
+    BinaryKellyBet,
+    CategoricalKellyBet,
+)
 
 
 def check_is_valid_probability(probability: float) -> None:
@@ -13,7 +23,7 @@ def get_kelly_bet_simplified(
     market_p_yes: float,
     estimated_p_yes: float,
     confidence: float,
-) -> SimpleBet:
+) -> BinaryKellyBet:
     """
     Calculate the optimal bet amount using the Kelly Criterion for a binary outcome market.
 
@@ -54,7 +64,7 @@ def get_kelly_bet_simplified(
     # Ensure bet size is non-negative does not exceed the wallet balance
     bet_size = CollateralToken(min(kelly_fraction * max_bet.value, max_bet.value))
 
-    return SimpleBet(direction=bet_direction, size=bet_size)
+    return BinaryKellyBet(direction=bet_direction, size=bet_size)
 
 
 def get_kelly_bet_full(
@@ -64,7 +74,7 @@ def get_kelly_bet_full(
     confidence: float,
     max_bet: CollateralToken,
     fees: MarketFees,
-) -> SimpleBet:
+) -> BinaryKellyBet:
     """
     Calculate the optimal bet amount using the Kelly Criterion for a binary outcome market.
 
@@ -98,7 +108,7 @@ def get_kelly_bet_full(
     check_is_valid_probability(confidence)
 
     if max_bet == 0:
-        return SimpleBet(size=CollateralToken(0), direction=True)
+        return BinaryKellyBet(size=CollateralToken(0), direction=True)
 
     x = yes_outcome_pool_size.value
     y = no_outcome_pool_size.value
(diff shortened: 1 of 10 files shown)


gabrielfior:
Very interesting PR! Some thoughts before reviewing:

> Allow the BettingStrategy class to handle multiple bets.

+1, would be nice.

> Allow shorting

We can refactor the strategy class to allow for shorting. Shorting == mint full set + sell the outcome to be shorted.

prediction_market_agent_tooling/tools/betting_strategies/kelly_criterion.py
```diff
@@ -144,7 +154,156 @@ def get_kelly_bet_full(
...
+                # If bet is negative, we "short" outcome i by buying all other outcomes
+                for j in range(n):
+                    if j == i:
+                        continue
+                    payout += get_buy_outcome_token_amount(
+                        CollateralToken(abs(bets[i]) / (n - 1)),
+                        j,
+                        outcome_pool_sizes,
+                        fees,
+                    ).value
```
gabrielfior, line 257: Maybe I'm missing something, but isn't selling of outcome `i` missing from this calculation?
The `mint of full set` makes sense.

gabrielfior:
Overall I'm happy with the code and don't have any corrections (maybe @TS9001 has a few).
My concern now is with the logic of `KellyBets`: `BinaryKellyBettingStrategy (full)` vs `CategoricalKellyBettingStrategy (full).`. In my head, all of the following should take place:
-> Let's assume we have a binary market "Will Elon Musk remain CEO of Tesla by Sep 01 2025?" (YES=0.7, NO = 0.3)
-> Let's assume agent's `p_yes` is 0.8.
1. Categorical and Binary Kelly should be able to handle the market
2. I expect both to deliver similar bet sizes (and the same bet direction).
3. The above should also hold in the case of underpriced markets, similar to the tests you wrote.

If that's not the case, I'm worried that either a. our original Kelly betting strategy is wrong or b. this new implementation is incorrect (for some unknown reason).
@kongzii can you check if 1,2,3 hold true?

kongzii:
> -> Let's assume we have a binary market "Will Elon Musk remain CEO of Tesla by Sep 01 2025?" (YES=0.7, NO = 0.3)
> -> Let's assume agent's p_yes is 0.8.
> 
> 1. Categorical and Binary Kelly should be able to handle the market
> 2. I expect both to deliver similar bet sizes (and the same bet direction).
> 3. The above should also hold in the case of underpriced markets, similar to the tests you wrote.

(1) They are, there are tests for that
(2) / (3) I did too, but as said on the sync, it's not the case and can be quite different. After all, their implementation is also very different, which is why I'm not sure if Kelly is still the correct name.

That's why I did the experimental test to see if these new strategies will be profitable or not. 

I got a few more ideas, though, today, so I will push a couple more commits.

scripts/optimize_betting_strategy.py
kongzii, line 1: I moved it here from `examples/monitor/match_bets_with_langfuse_traces.py`, as that name didn't match the real use case of the script anymore. I was hoping that GitHub would render diffs correctly.

But only changes in this script are:

1. Refactor from looping through all agents to running this script per agent. So we can run it in parallel. Because optuna uses threading when you specify n_jobs>0, and that doesn't help us.
2. Added more logging (saving per-bet details, mean/median of bets, etc.)
3. Some try-expect to not crash on edge cases.
4. Early-stopping in Optuna instead of fixed number of iterations.
5. Fix httpx caching.

kongzii:
# New more exhaustive summary of PR, after latest changes

There are two new Kelly-based strategies

- get_kelly_bets_categorical_simplified
- get_kelly_bets_categorical_full

As the names suggest, they are analogical to the current version for binary markets.

The simplified version accepts also flags:

- allow_multiple_bets --> if multiple bets per prediction are allowed (ie 0.1 xDai on outcome 1, 0.15 xDai on outcome 2, etc.)
- allow_shorting --> if shorting is allowed (returning negative bet size for the given outcome)

On top of these two, the full version accepts also flag:

- multicategorical --> if multiple outcomes can be selected as correct ones on the market

## get_kelly_bets_categorical_simplified vs. get_kelly_bet_simplified

If we set allow_multiple_bets=False and allow_shorting=False, these two are the same. That's clear from the equation and also asserted in tests..

If we set allow_multiple_bets=True, results can be different, which is easy to see in the code,
as it's doing `bet_size = (kelly_fraction / total_kelly_fraction) * max_bet.value` to divide the max bet between outcomes.

If we set allow_shorting=True, result can be different, it's not intuitive at first, as on PM, shorting one outcome should be equal to buying all others, but in Kelly, it's not, because odds are different.

Let's imagine situation:

On YES outcome:

```
max_bet = 5 xDai
agent_p_yes = 0.6
market_p_yes = 0.8

edge = agent_p_yes - market_p_yes = -0.2
odds = (1 / market_p_yes) - 1 = 0.25
kelly_fraction = edge / odds = -0.8

kelly_bet = max_bet * kelly_fraction = -4.0
```

This tells us to short 4 xDai of YES --> buy 4 xDai on NO.

Now for NO outcome:

```
max_bet = 5 xDai
agent_p_no = 0.4
market_p_no = 0.2

edge = agent_p_no - market_p_no = 0.2
odds = (1 / market_p_no) - 1 = 4
kelly_fraction = edge / odds = 0.05

kelly_bet = max_bet * kelly_fraction = 0.25
```

This tells us to buy 0.25 xDai of NO.

Milion dollar question is which one is better. 4 xDai of NO gives potential profit 16 xDai, 0.25 xDai gives 1.25 xDai, in both cases, 400% profits.
If max_bet is 5, I don't see reason not to bet 4 actually.

## get_kelly_bets_categorical_full vs. get_kelly_bet_full

Now this is tricky, even if we set allow_multiple_bets, allow_shorting, multicategorical all to False, they do agree in directions of the bets, but it just doesn't add up with sizes, and differences can be up to 99% of bet size.

Unlike in get_kelly_bet_full, categorical version don't have exact closed-form solution, and numerical optimization is needed.

And that depends on many things:

- initial bet vector x0 (currently initialised using get_kelly_bets_categorical_simplified), even small difference here will cause converge to a different bet vector (not a wrong one, just less or more profitable)
- given constraints (for example hard constraint of single-bet can not be even implement)
- randomness in scipy
- how exactly is neg_expected_log_utility defined (it's using payouts or profits, how is confidence used?)

So here I'm still falling back to experimental tests from above, where this categorical version is indeed comparable in profitability to the binary version, if it's used on many markets:

Screenshot by Dropbox Capture

Screenshot by Dropbox Capture

Screenshot by Dropbox Capture

I'm going to run this over all agents and report it in a new PR. But I'm hoping it will hold.

@gabrielfior @TS9001 Long and exhaustive comment 😄

PR #813, opened 2025-07-30
Added capping for kelly-bet-strategy

TS9001:
https://github.com/gnosis/prediction-market-agent/issues/929

Reused methods from https://github.com/gnosis/prediction-market-agent-tooling/pull/742 - hopefully will be enough - was able to test just by units

### DIFF
--- a/prediction_market_agent_tooling/markets/seer/seer.py
+++ b/prediction_market_agent_tooling/markets/seer/seer.py
@@ -2,6 +2,7 @@ import asyncio
 import typing as t
 from datetime import timedelta
 
+import cachetools
 from cowdao_cowpy.common.api.errors import UnexpectedResponseError
 from eth_typing import ChecksumAddress
 from web3 import Web3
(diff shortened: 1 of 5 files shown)


prediction_market_agent_tooling/deploy/betting_strategy.py
```diff
@@ -443,8 +443,22 @@ def calculate_trades(
...
+        kelly_bet_size_usd = market.get_token_in_usd(kelly_bet_size)
+        existing_amount = (
+            existing_position.amounts_current.get(bet_outcome, USD(0))
+            if existing_position
+            else USD(0)
+        )
+        capped_amount = BettingStrategy.cap_to_profitable_position(
+            market=market,
+            existing_position=existing_amount,
+            wanted_position=existing_amount + kelly_bet_size_usd,
```
kongzii, line 456:
```suggestion
            wanted_position=kelly_bet_size_usd,
```

Shouldn't this be just like this?

Otherwise, it's always increasing the `max_position_amount` by the previous bet, which isn't what is wanted. (just a design decision that we agreed on, but now it's nice to have it consistent between strategies).

kongzii (approved):

PR #794, opened 2025-07-17
Market disbalance arbitrage

TS9001:

PR #826, opened 2025-08-06
Added token pools fetching to enable kelly-bet

TS9001:
Unfortunatelly https://github.com/gnosis/prediction-market-agent-tooling/pull/813 did not worked and agents are still failing and https://github.com/gnosis/prediction-market-agent/issues/929 is not solved

After intensive testing I have found that problem is that Kelly-Bet requires outcome-token-pool values to be set see https://github.com/gnosis/prediction-market-agent-tooling/blob/286856f3b0575c1bb0be0710db946fa353e93626/prediction_market_agent_tooling/deploy/betting_strategy.py#L495

This change uses fix proposed inside of https://github.com/gnosis/prediction-market-agent-tooling/pull/794 to fill the token_pools values and fetch values from the pool instead of trying to fetch the initial values from cow. Until the cow solver is done this is probably good way to go because it shortens market loading from several minutes to ~20 sec.

### DIFF
--- a/prediction_market_agent_tooling/markets/seer/price_manager.py
+++ b/prediction_market_agent_tooling/markets/seer/price_manager.py
@@ -220,7 +220,7 @@ class PriceManager:
                 )
                 probability_map[
                     OutcomeStr(model.outcomes[wrapped_tokens.index(token)])
-                ] = Probability(round(pool.token0Price, 2).value)
+                ] = Probability(pool.token0Price.value)
             else:
                 outcome_token_pool[
                     OutcomeStr(model.outcomes[wrapped_tokens.index(token)])
@@ -231,7 +231,7 @@ class PriceManager:
                 )
                 probability_map[
                     OutcomeStr(model.outcomes[wrapped_tokens.index(token)])
-                ] = Probability(round(pool.token1Price, 2).value)
+                ] = Probability(pool.token1Price.value)
 
         for outcome in model.outcomes:
             if outcome not in outcome_token_pool:
--- a/prediction_market_agent_tooling/markets/seer/price_manager.py
+++ b/prediction_market_agent_tooling/markets/seer/price_manager.py
@@ -2,6 +2,9 @@ from cachetools import TTLCache, cached
 from pydantic import BaseModel
 from web3 import Web3
 
+from prediction_market_agent_tooling.deploy.constants import (
+    INVALID_OUTCOME_LOWERCASE_IDENTIFIER,
+)
 from prediction_market_agent_tooling.gtypes import (
     ChecksumAddress,
     CollateralToken,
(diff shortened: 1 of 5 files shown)


prediction_market_agent_tooling/markets/seer/price_manager.py
```diff
@@ -182,3 +183,59 @@ def build_probability_map(self) -> dict[OutcomeStr, Probability]:
...
+                outcome_token_pool[
+                    OutcomeStr(model.outcomes[wrapped_tokens.index(token)])
+                ] = (
+                    OutcomeToken(pool.totalValueLockedToken0)
+                    if pool.totalValueLockedToken0 is not None
+                    else OutcomeToken(0)
+                )
+                probability_map[
+                    OutcomeStr(model.outcomes[wrapped_tokens.index(token)])
+                ] = Probability(round(pool.token0Price, 2).value)
```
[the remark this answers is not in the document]
kongzii: hm hm, is the rounding required for something?

prediction_market_agent_tooling/markets/seer/price_manager.py
```diff
@@ -182,3 +183,59 @@ def build_probability_map(self) -> dict[OutcomeStr, Probability]:
...
+                    else OutcomeToken(0)
+                )
+                probability_map[
+                    OutcomeStr(model.outcomes[wrapped_tokens.index(token)])
+                ] = Probability(round(pool.token1Price, 2).value)
+
+        for outcome in model.outcomes:
+            if outcome not in outcome_token_pool:
+                outcome_token_pool[outcome] = OutcomeToken(0)
+                probability_map[outcome] = Probability(0)
```
kongzii, line 239: Are there scenarios where we know the price/probability for one outcome, not for the other, and it then defaults to 0?

What I'm worried about is that if we default to 0, and the agent predicts that outcome with a probability of 90, for example.

Consequently, it will either fail to buy that outcome, or it will try to buy it aggressively as it supposes it's hugely underpriced.

kongzii: In the current version, we error out in case like this: https://github.com/gnosis/prediction-market-agent-tooling/blob/930572912820660a8f8969dc6f946f7d1f81bcf2/prediction_market_agent_tooling/markets/seer/price_manager.py#L154

TS9001: Yes this is more or less fix for "Invalid Token" - it has 0% in most of the cases but not necessarily so I cannot just remove it. 
Rounding is gone.

TS9001: Changed it to throw when probability is missing for relevant token

PR #866, opened 2025-08-21
Optimize slow swapr call

kongzii:
I found that most of the time was eaten by a single test:

image

And there, the swapr call's gas estimation was taking all the time:

image


prediction_market_agent_tooling/tools/web3_utils.py
```diff
@@ -120,17 +122,55 @@
...
+            logger.warning(
+                f"Gas estimation timed out after {timeout} seconds, using default: {default_gas}"
+            )
+
+        executor.shutdown(wait=False, cancel_futures=True)
+        return default_gas
+    except Exception:
+        logger.warning(f"Error during gas estimation, using default: {default_gas}.")
+        return default_gas
+
```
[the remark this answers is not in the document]
kongzii: Cant do, with context manager it autimatically sets wait=True and then it's no benefit for us

prediction_market_agent_tooling/tools/cow/semaphore.py
```diff
@@ -28,7 +31,7 @@ def decorator(func: F) -> F:
         def wrapper(*args: Any, **kwargs: Any) -> Any:
             sqlalchemy_db_url = (
                 api_keys.sqlalchemy_db_url.get_secret_value()
-                if api_keys.SQLALCHEMY_DB_URL
+                if shared_db
```
kongzii, line 34: @gabrielfior Any reason to enforce this limit across all code we use?

Seems very inefficient to rate limit myself on my local pc, because all agents are running in Kube. It was causing me slowness.

Similarly, for agents, I doubt CoW would rate-limit the whole company's GCP?

.github/workflows/python_ci.yaml
```diff
@@ -32,21 +32,27 @@ jobs:
...
-            command: 'poetry run python -m pytest tests_integration_with_local_chain/ --disable-isolation -vvv'
+            command: "poetry run python -m pytest tests_integration/ -p no:ape_test -vvv --durations=0"
+          - name: Integration with Local Chain (Omen)
+            command: "poetry run python -m pytest tests_integration_with_local_chain/ --disable-isolation -vvv --durations=0 -k omen"
+          - name: Integration with Local Chain (Seer)
+            command: "poetry run python -m pytest tests_integration_with_local_chain/ --disable-isolation -vvv --durations=0 -k seer"
+          - name: Integration with Local Chain (Polymarket)
+            command: "poetry run python -m pytest tests_integration_with_local_chain/ --disable-isolation -vvv --durations=0 -k polymarket"
+          - name: Integration with Local Chain (Others)
+            command: "poetry run python -m pytest tests_integration_with_local_chain/ --disable-isolation -vvv --durations=0 -k 'not omen and not seer and not polymarket'"
```
kongzii, line 48: Thanks to the --duration=0, we see the per-test run times, to see in the future right away what's the bottleneck

Thanks to the -k, we now run even more tests in parallel, according to per-platform blocks

prediction_market_agent_tooling/tools/cow/semaphore.py
kongzii: Ah that's 💩 Let me think a bit before merging then.

prediction_market_agent_tooling/tools/web3_utils.py
```diff
@@ -120,17 +123,45 @@ def prepare_tx(
...
+    """
+    Tries to estimate the gas, but default to the default value on timeout.
+    """
+    try:
+        # We need n_jobs=-1, otherwise timeouting isn't applied.
+        result: list[int] = Parallel(n_jobs=-1, backend="threading", timeout=timeout)(
+            delayed(function_call.estimate_gas)(tx_params) for _ in range(1)
+        )
+        estimated_gas = result[0]
+        return int(estimated_gas * 1.2)  # Add 20% buffer
```
gabrielfior, line 157: Using `multiprocessing` in this context seems overly complex to me.
I would stick with a simple call to `function_call.estimate_gas(tx_params)` and, if timeout, stick to default gas.

gabrielfior: Ow, now I get it - sure, makes sense to me.

gabrielfior (approved):

PR #869, opened 2025-08-21
850 finish redeem winnings in seer

gabrielfior:
The original logic works for the case described. I just added one test.


prediction_market_agent_tooling/markets/agent_market.py
```diff
@@ -425,7 +425,7 @@ def get_binary_market(id: str) -> "AgentMarket":
         raise NotImplementedError("Subclasses must implement this method")
 
     @staticmethod
-    def redeem_winnings(api_keys: APIKeys) -> None:
+    def redeem_winnings(api_keys: APIKeys, web3: Web3|None=None) -> None:
```
gabrielfior, line 428: Here's one big change I did, this was necessary because I wanted to call `SeerAgentMarket.redeem_winnings(local_web3)`.
There are workarounds but this is the easiest thing to add.

prediction_market_agent_tooling/tools/contract.py
```diff
@@ -431,6 +431,25 @@ class ContractWrapped1155BaseClass(ContractERC20BaseClass):
         )
     )
 
+    def factory(self, web3: Web3 | None = None) -> ChecksumAddress:
+        return Web3.to_checksum_address(self.call("factory", web3=web3))
```
gabrielfior, line 435: returns the `admin` which is able to mint outcome tokens (during test)

Issue #861, opened 2025-08-18
Seer outcomes don't match the outcomes from probabilities


gabrielfior:
See this market with outcomes (note the extra space) `Yes` and `No `.
Some ways to deal with this:
1 - Calculate probabilities using market outcome strings, instead of hardcoded strings
2 - Ignore markets which don't follow `Yes` and `No` correctly formatted
3 - Ignore this error

https://console.cloud.google.com/logs/query;cursorTimestamp=2025-08-18T00:04:58.849428278Z;duration=PT1H;query=resource.labels.namespace_name%3D%22agents%22%0Aresource.labels.container_name%3D%22main%22%0Alabels.%22k8s-pod%2Fagent_job%22%3D%22true%22%0Aseverity%3DERROR%0ANOT%20%22OutOfFundsError%22%0ANOT%20%22ResourceWarning:%20%22%0ANOT%20%22DeprecationWarning%22%0A%20timestamp%3E%3D%222025-08-15T14:38:27.641Z%22%20%0A%20--timestamp%3C%3D%222025-08-18T15:38:27.641Z%22%20%0A%20severity%3E%3DDEFAULT%0Atimestamp%3D%222025-08-18T00:04:58.849428278Z%22%0AinsertId%3D%225tp4fsjos94lpbgs%22?project=gnosis-ai

PR #870, opened 2025-08-22
Strip spaces in outcome strings for Seer

TS9001:
Fixes: https://github.com/gnosis/prediction-market-agent-tooling/issues/861

### DIFF
--- a/prediction_market_agent_tooling/markets/seer/price_manager.py
+++ b/prediction_market_agent_tooling/markets/seer/price_manager.py
@@ -183,7 +183,7 @@ class PriceManager:
             outcome = self.seer_market.outcomes[
                 self.seer_market.wrapped_tokens.index(outcome_token)
             ]
-            normalized_prices[outcome] = new_price
+            normalized_prices[OutcomeStr(str(outcome).strip())] = new_price
 
         return normalized_prices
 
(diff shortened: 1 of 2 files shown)


gabrielfior (approved):

prediction_market_agent_tooling/markets/seer/price_manager.py
```diff
@@ -183,7 +183,7 @@ def build_probability_map(self) -> dict[OutcomeStr, Probability]:
             outcome = self.seer_market.outcomes[
                 self.seer_market.wrapped_tokens.index(outcome_token)
             ]
-            normalized_prices[outcome] = new_price
+            normalized_prices[OutcomeStr(str(outcome).strip())] = new_price
```
kongzii, line 186:
```suggestion
            normalized_prices[OutcomeStr(outcome.strip())] = new_price
```

This should work 🤔 

Wouldn't be a big deal, just if we ever change type of `outcome` to something else, mypy wouldn't complain here as almost anything can be stringified.

TS9001: Agree fixed.

prediction_market_agent_tooling/markets/seer/data_models.py
```diff
@@ -103,6 +110,14 @@ class SeerMarket(BaseModel):
     upper_bound: SeerNormalizedWei = Field(alias="upperBound", default=None)
     lower_bound: SeerNormalizedWei = Field(alias="lowerBound", default=None)
 
+    @model_validator(mode="before")
+    def normalize_outcomes(cls, values: dict[str, t.Any]) -> dict[str, t.Any]:
+        if "outcomes" in values:
+            values["outcomes"] = [
```
kongzii, line 116: I'm not sure if it's good idea to modify market's data on response itself. `SeerMarket` is model for what comes out of subgraph and seeing a different data here could cause hard-to-notice bugs in the future.

On the other hand, we can say it's fine and normalize it on all market types (Omen, Polymarket, Manifold). 

But maybe we can normalize it on `AgentMarket` classes, and not the source classes at least.

Because in Omen for example, we do this kind of thing just before doing the bets, not at the data source: https://github.com/gnosis/prediction-market-agent-tooling/blob/3d4d245cb8bd414133bf551e352c6054a779763a/prediction_market_agent_tooling/markets/agent_market.py#L523

So we could agree on the "correct" way and do it in the same fashion everywhere.

What are your thoughts on this?

prediction_market_agent_tooling/markets/seer/price_manager.py
TS9001: done

PR #879, opened 2025-08-28
Test losing precision on Weis in GenericValue

kongzii:


prediction_market_agent_tooling/tools/_generic_value.py
```diff
@@ -239,7 +239,7 @@ def __get_pydantic_core_schema__(
         cls, source_type: t.Any, handler: GetCoreSchemaHandler
     ) -> CoreSchema:
         # Support for Pydantic usage.
-        dt_schema = handler(str | int | float | dict)
+        dt_schema = handler(cls.parser | str | dict)
```
kongzii, line 242: Wei values are defined as subclasses of 

```
class OutcomeWei(_GenericValue[Web3Wei | int | str, Web3Wei], parser=int):
```

However, it's interesting that when the handler included `float`, pydantic automatically used float (even when integers were passed around), and we lost the precision that caused the bug Gabriel reported.

TS9001 (approved):

PR #881, opened 2025-09-08
Universal function to get implementation address

kongzii:
deploy please

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -172,13 +172,13 @@ typing-extensions = {version = ">=4.2", markers = "python_version < \"3.13\""}
 
 [[package]]
 name = "alembic"
-version = "1.16.4"
+version = "1.16.5"
 description = "A database migration tool for SQLAlchemy."
 optional = true
 python-versions = ">=3.9"
 files = [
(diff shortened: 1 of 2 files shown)


prediction_market_agent_tooling/tools/contract.py
```diff
@@ -909,84 +909,77 @@ def contract_implements_function(
...
     if not implements and look_for_proxy_contract:
-        # Case 1: Check if it's a standard proxy (has implementation() function)
-        if contract_implements_function(
-            contract_address, "implementation", web3, look_for_proxy_contract=False
-        ):
-            # Get the implementation address and check the function there
-            implementation_address = ContractProxyOnGnosisChain(
-                address=contract_address
-            ).implementation()
+        imp_address = uni_implementation_address(contract_address, web3)
```
kongzii, line 912: I refactored the previous version, where we had:

- if this contract is a proxy of type A, does that proxy implements this function?
- if this contract is a proxy of type B, does that proxy implements this function?
- if this contract is a proxy of type C, does that proxy implements this function?

Into:

-  implements this contract a proxy of typy A, B or C?
- and if yes, does that proxy implemnts this function?

Because in TG, I needed a way to obtain the implementation address of a potential proxy.

kongzii:
(failing tests are unrelated)

PR #935, opened 2026-03-31
Replace subgrounds with httpx for subgraph queries

### LINKED ISSUE #740: Replace subgrounds with regular httpx calls (that enable caching)

cats2101:
## Summary

Closes #740

Replaces the `subgrounds-gnosis` library with direct `httpx` GraphQL calls for all subgraph interactions:

- Rewrites `BaseSubgraphHandler` to use `httpx.Client` with `tenacity` retries and automatic skip-based pagination (handles The Graph's 1000-item limit)
- Converts all `_get_fields_for_*` methods from FieldPath objects to static GraphQL field selection strings
- Converts `_build_sort_params` to return plain field name strings instead of FieldPath objects
- Converts list-based where clauses (using FieldPath operator overloading) to dict-based where clauses in `get_trades` and `get_answers`
- Removes `subgrounds-gnosis` from `pyproject.toml` and regenerates `poetry.lock`

All handler public APIs remain unchanged — callers don't need any modifications.

## Test plan

- [x] All 201 unit tests pass (18 pre-existing failures due to missing `GRAPH_API_KEY` and postgres — same on `main`)
- [ ] Integration tests against live subgraph endpoints


kongzii:
Hey, on first look this looks good! Thanks!

We will need to test it more throgoughly to see if all agents are working with this, but so far lgtm

cats2101:
Gentle ping — anything else needed here? Happy to address any concerns from further testing.

PR #941, opened 2026-04-01
Fixes for replicator

kongzii:
deploy please


prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py
```diff
@@ -59,9 +56,6 @@
     GNOContract(),
     WETHContract(),
     EUReContract(),
-    SAFEContract(),
-    COWContract(),
-    MetriSuperGroup(),
```
kongzii, line 64: Metri token isn't tradeable on CoW anymore and looks like no one cares, so we cna just ignore it.

I also removed SAFE and COW to keep only big tokens and be less stressed about swap errors 😄

TS9001 (approved):

PR #944, opened 2026-04-02
Multimarket

TS9001:


prediction_market_agent_tooling/markets/polymarket/polymarket.py
```diff
@@ -279,24 +314,22 @@ def get_markets(
...
         condition_models_dict = {c.id: c for c in condition_models}
 
         result_markets: list[PolymarketAgentMarket] = []
         for m in markets:
-            market = PolymarketAgentMarket.from_data_model(m, condition_models_dict)
-            if market is not None:
-                result_markets.append(market)
+            result_markets.extend(
+                PolymarketAgentMarket.from_data_model_all(m, condition_models_dict)
+            )
```
TS9001, line 360: I decided to flatten the market - not sure if it is better for agent to have hierarchical or this condition

prediction_market_agent_tooling/markets/polymarket/api.py
```diff
@@ -279,6 +279,35 @@ def get_gamma_event_by_slug(slug: str) -> PolymarketGammaResponseDataItem:
...
+    client: httpx.Client = HttpxCachedClient(ttl=timedelta(seconds=60)).get_client()
+    url = urljoin(POLYMARKET_GAMMA_API_BASE_URL, "markets")
+    r = client.get(url, params={"condition_id": condition_id.to_0x_hex()})
+    r.raise_for_status()
+    data = r.json()
+    if not data:
+        raise ValueError(
+            f"No market found for condition_id '{condition_id.to_0x_hex()}'"
+        )
+    market_data = data[0]
```
kongzii, line 302: PR description says multimarket, but here it's just first index, or is this something different?

prediction_market_agent_tooling/markets/polymarket/data_models.py
```diff
@@ -32,7 +32,7 @@ class PolymarketSideEnum(str, Enum):
     SELL = "SELL"
 
 
-# TODO: Currently unused. Wire into epic tasks #2/#4/#6 or remove if not needed.
+# TODO: Use when CLOB API is introduced for real-time trading.
```
kongzii, line 35: What is difference between real tiem trading and current trading?

prediction_market_agent_tooling/markets/polymarket/data_models.py
```diff
@@ -58,6 +58,15 @@ class PolymarketGammaMarket(BaseModel):
     archived: bool
     questionId: str | None = None
     clobTokenIds: str | None = None  # int-encoded hex
+    question: str | None = None
```
kongzii: Why would question be None?

prediction_market_agent_tooling/markets/polymarket/api.py
TS9001: Unfortunatelly this is I think necessary - as far as I understood polymarket has two distinc objects. Events and Markets. Event  is the top-level question/topic (e.g. "2024 US Presidential Election") . Market is a specific tradeable outcome within that event (e.g. "Will Trump win?", "Will Biden win?", "Will RFK win?"). Unfortunately we cannot get event by condition id (market) so we have to take condition first (we want only data for this market) and get data by slug (possibly doable directly by event but I did not tested that - should work same way).

So here it is ok to have just [0] data - necessary, on the other hand we probably could use event_id instead of slug - that could be little-bit more readable. What do you think?

prediction_market_agent_tooling/markets/polymarket/data_models.py
TS9001: It looks directly to the orderbook - realtime prices etc. We will have to do this in order to have best possible pricing. Probably something for the future...

prediction_market_agent_tooling/markets/polymarket/api.py
kongzii: If it works as expected, I'm fine 😄 Just wanted to double check as it seemed against the PR intention at first look

prediction_market_agent_tooling/markets/polymarket/data_models.py
TS9001: Good catch fixed

prediction_market_agent_tooling/markets/polymarket/api.py
TS9001: Actually changed it to event_id :D more readable

kongzii (approved):

PR #947, opened 2026-04-08
Add Polymarket copy trading module

TS9001:


prediction_market_agent_tooling/markets/polymarket/copy_trading.py
```diff
@@ -0,0 +1,451 @@
...
+
+class TraderSortBy(str, Enum):
+    VOLUME = "volume"
+    PNL = "pnl"
+    ROI = "roi"
+    WIN_RATE = "win_rate"
+
+
+class TraderProfile(BaseModel):
+    address: ChecksumAddress
```
kongzii:
```suggestion
    address: VerifiedChecksumAddress
```

Try to get use to `VerifiedChecksumAddress` instead of simple `ChecksumAddress` 🙏 

The difference is:

```
VerifiedChecksumAddress: TypeAlias = Annotated[
    ChecksumAddress, AfterValidator(Web3.to_checksum_address)
]
```

This means that `VerifiedChecksumAddress` will always be checksum format, guaranteed. Even if the API doesn't return it in checksum format.

(but I also still use `ChecksumAddress` as I forget about it constantly 😄 )

prediction_market_agent_tooling/markets/polymarket/copy_trading.py
```diff
@@ -0,0 +1,451 @@
...
+            skip_reason=skip_reason,
+        )
+
+    def replicate_trade(self, trade: PolymarketTradeResponse) -> ReplicatedTradeResult:
+        # Look up market
+        try:
+            market = PolymarketAgentMarket.get_binary_market(
+                trade.conditionId.to_0x_hex()
+            )
+        except Exception as e:
```
kongzii, line 160: There is couple of `except Exception` without any error log or anything. Is that on purpose?

kongzii (approved):

